PR #15, opened 2015-01-12
now json is fashion

ahu0605:
The class should support data json

### DIFF
--- a/java/org/apache/catalina/connector/Request.java
+++ b/java/org/apache/catalina/connector/Request.java
@@ -2987,7 +2987,7 @@ public class Request
             } else {
                 contentType = contentType.trim();
             }
-
+                    
             if ("multipart/form-data".equals(contentType)) {
                 parseParts(false);
                 success = true;
@@ -2998,7 +2998,9 @@ public class Request
                 success = true;
                 return;
             }
-
+             if ("application/json".equals(contentType)) {
+                //TODO for post data type  json
+            }
             int len = getContentLength();
 
             if (len > 0) {


markt-asf:
This pull request is declined. Even ignoring the poor quality of this pull request, supporting JSON formatted data is an application concern, not a container concern.

ahu0605:
but when i push content type - json，there is some bug , 4 ex :data is json "{\"id\":\"asd\"}",the value  will be a Map key, request.getPamaterMap () the key is {"id":"asd"}, value is null ,is not logic

martin-g:
When you POST JSON data you should use http://docs.oracle.com/javaee/6/api/javax/servlet/ServletRequest.html#getReader() to consume the body and parse it as you find appropriate.

hazendaz:
We run json using Jackson with jersey in tomcat.  No issues.  This code doesn't do anything.  So unclear how it solves anything.  All you did was add an empty if statement.

--- Original Message ---

martin-g:
I'm pretty sure Jersey does something.

hazendaz:
Sure...what I meant here is that this patch does nothing from tomcat perspective.  It's just an empty if statement.  It does nothing at all.

--- Original Message ---

ahu0605:
I am very sad this bug is not be modifed, PHP had surpported  content -type :JSON ,but java ...

ebourg:
As I understand this pull request is more a request for enhancement than an actual code contribution. I guess you suggest that Tomcat implements the [W3C HTML JSON form submission](http://www.w3.org/TR/html-json-forms/) draft, correct?

ahu0605:
yes,from Perspective of development. it is benifit  java ,it is benifit  Tomcat ,our company has choice resin,if the problem keep alive, i maybe choice  PHP or Python

PR #19, opened 2015-03-30
Added support for addWebapp(String, InputStream) 

ThreaT:
When using embedded tomcat, if jsp files are inside the jar file, it is impossible to use the respective jsp files as a webapp unless tomcat accepts InputStreams as webapps. The coupled usages still need refactoring.

### DIFF
--- a/java/org/apache/catalina/startup/Tomcat.java
+++ b/java/org/apache/catalina/startup/Tomcat.java
@@ -68,26 +68,29 @@ import org.apache.tomcat.util.descriptor.web.LoginConfig;
 // TODO: lazy init for the temp dir - only when a JSP is compiled or
 // get temp dir is called we need to create it. This will avoid the
 // need for the baseDir
+
 // TODO: allow contexts without a base dir - i.e.
 // only programmatic. This would disable the default servlet.
+
 /**
  * Minimal tomcat starter for embedding/unit tests.
  *
- * Tomcat supports multiple styles of configuration and startup - the most
- * common and stable is server.xml-based, implemented in
- * org.apache.catalina.startup.Bootstrap.
+ * Tomcat supports multiple styles of configuration and
+ * startup - the most common and stable is server.xml-based,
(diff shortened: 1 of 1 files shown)


kkolinko:
1) Bugzilla issue = ?
2) "it is impossible to use the respective jsp files as a webapp" - I do not believe that,  with Resources Jars support in Servlet 3.0 spec.
3) -1 as changes formatting in many places, unrelated to your change.
"InputStream docBase" is a confusing name,  and I do not see how you use the InputStream, at all. It is an unused parameter.

ThreaT:
1) I'm not sure
2) Why does it have to?
3) Sorry about formatting. Please see last commit of the three as I removed it. I kept "docBase" in order to accurately overload the original method to avoid usage confusion. It is an unused paramter at the moment but the method itself I am already using as a placeholder until the dev around it is complete.

markt-asf:
-1.
Servlet 3.0 and later webapps have resource JARs (assuming the JAR is packaged correctly). For single JAR apps, the Tomcat 8 resources implementation can be used to serve JSPs from any location within a JAR.
Supporting this in Tomcat 7 would require a new DirContext impl. Given how fragile that code is I'd be extremely reluctant to make any changes to it that were not strictly necessary.

ThreaT:
Just so we're on the same page, I tried:

```
context = tomcat.addWebapp("/", "jar:file:/home/user/test/MyJar.jar!/com/mvc/view");
```

Which I found did not work. Are you saying that this **does** work?

markt-asf:
No. That is not what I am saying at all.

I am saying the functionality you state you want to add (serving JSPs from within a JAR) can already be achieved by existing functionality for Tomcat 8 and trunk. Tomcat 7 would require a new DirContext implementation (which is likely to be very risky).

ThreaT:
Could you possibly demonstrate how this could be achieved with Tomcat 8?

PR #20, opened 2015-05-28
Recursively create document base directory

lvarvel:
The "expand" method will create the docBase directory, however sometimes this fails because the intermediate "webapps" directory has not been created already.  This will create the intermediate directory if it hasn't already been created externally.


markt-asf:
Starting Tomcat with an appBase that is not an existing directory is an error and is reported as such in the logs.

PR #21, opened 2015-06-07
Add JASPIC API (JSR 196) to build process

fjodorver:
I'd added jaspic api support in this commit.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -107,6 +107,8 @@
   <property name="jasper.jar" value="${tomcat.build}/lib/jasper.jar"/>
   <property name="jasper-el.jar" value="${tomcat.build}/lib/jasper-el.jar"/>
 
+  <property name="jaspic-api.jar" value="${tomcat.build}/lib/jaspic-api.jar"/>
+
   <!-- Standard Source JARs -->
   <property name="bootstrap-src.jar" value="${tomcat.src.jars}/bootstrap-src.jar"/>
   <property name="tomcat-juli-src.jar" value="${tomcat.src.jars}/tomcat-juli-src.jar"/>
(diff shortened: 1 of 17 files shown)


markt-asf:
OK. I've copied the Geronimo classes across and fixed the checkstyle and compilation errors. I've now seen some additional issues that need to be addressed.
1. Remove the @version lines from the Javadoc (and the whole comment if that is the only Javadoc present)
2. Line length (80 for comments, 100 for code)
3. Use imports (particularly noticed in AuthConfigFactory)

And something I forgot in my comment on the commit, the integration into build.xml should be a separate commit to.

fjodorver:
It's updated pull request.
1) Build.xml configuration in separate commit now
2) Fixed imports and formatting
3)  Method refactoring has been made separately. Geronimo uses OSGi stuff to create this factory class, I see you had already fixed that. 
Threre are two reasons why this refactoring required in my opinion: get rid of extra final variable and get rid of extra indent.
4) removed version from javadoc
5) fixed checkstyle errors
6) What about JASPIC 1.1 changes in API, I don't see any changes between this version. It looks like 1.1 changes is more like internal ones, and API classes are still the same. Correct me if I am wrong.

markt-asf:
1) I haven't reviewed the build.xml changes since I don't want to start building this JAR until the API is correct.
2) Patch was incomplete. Applied along with additional patch for missing elements.
3) I'm not convinced (yet) of the need for this refactoring. And it appears to have broken the caching.
4) Patch was incomplete. Applied along with additional patch for missing elements.
5) I'm not convinced. At least one commit contained tabs which checkstyle would have rejected.
6) Look harder. For example. AuthConfigFactory has a number of additional constants. I typically use javap against both Tomcat and the reference implementation and diff the two results. I'll even go as far as re-ordering the methods in Tomcat to remove changes from the diff. When I ran this against the Geronimo classes there were a number of differences between Geronimo and the reference implementation.

fjodorver:
2,4,5) I use ant validate for checkstyle validation. Is it enough?
3) It looks like we need to rewrite AuthConfigFactory anyway (for example, possible memory leaks, huge method and so on). I personally prefer to introduce small methods, which makes code reading much easier, because they work as self-commented code. Also, it simplifies code testing. As a bonus, in simple methods have shorter lines. For example, in current implementation need to introduce an ugly final variable, instead of just getting the correct value and make the first variable final. Second thing is guard clauses - usually I prefer to make such checks in the beginning of the method. It's quite good to get rid of necessary indentation and makes code lines shorter.
6. I've refined method order and some signatures (there are runtime exceptions declared, which is not necessary). However, constants in AuthConfigFactory are used for internal purposes, so I'd proposed security management as separate patch.

PR #23, opened 2015-11-03
Fixed UnsupportedOperationException when releasing an user-provided URLStreamHandlerFactory

ctalau:
CopyOnWriterArrayList iterator remove operation was invoked when
de-registering a user-provided URLStreamHandlerFactory.

I also added a test case for that.


violetagg:
Hi,
Thanks for the patch.
Fix is available in trunk and Tomcat 8 (for 8.0.29 onwards).
Regards,
Violeta

PR #25, opened 2015-12-17
Change response character encoding

ukari:
It seems that encoding was forgotten to be changed in function doFilter().

When I set AddDefaultCharsetFilter in web.xml, like the following:
```xml
<filter>
  <filter-name>addDefaultCharsetFilter</filter-name>
  <filter-class>org.apache.catalina.filters.AddDefaultCharsetFilter</filter-class>
  <init-param>
    <param-name>encoding</param-name>
    <param-value>utf-8</param-value>
  </init-param>
  <async-supported>true</async-supported>
</filter>
<filter-mapping>
  <filter-name>addDefaultCharsetFilter</filter-name>
  <url-pattern>/*</url-pattern>
</filter-mapping>
```
response encoding is still ISO-8859-1

### DIFF
--- a/java/org/apache/catalina/filters/AddDefaultCharsetFilter.java
+++ b/java/org/apache/catalina/filters/AddDefaultCharsetFilter.java
@@ -80,9 +80,10 @@ public class AddDefaultCharsetFilter extends FilterBase {
     @Override
     public void doFilter(ServletRequest request, ServletResponse response,
             FilterChain chain) throws IOException, ServletException {
-
+        
         // Wrap the response
         if (response instanceof HttpServletResponse) {
+            response.setCharacterEncoding(encoding);
             ResponseWrapper wrapped =
                 new ResponseWrapper((HttpServletResponse)response, encoding);
             chain.doFilter(request, wrapped);


violetagg:
Hello,

The encoding specified by the user is provided to the ResponseWrapper.
The ResponseWrapper is provided with the method `doFilter` instead of the original Response. Thus after this filter every component will work with this ResponseWrapper with the correct encoding.

Why do you need to set the encoding to the original Response?

Regards,
Violeta Georgieva

ukari:
Oh, you are right, sorry I make the mistake.
But why response's encoding not changed after I set AddDefaultCharsetFilter?(I mean a HttpServletResponse)
Is the warpper effective?

Thanks

violetagg:
Can you provide a sample example that demonstrates the issue?

wenjiezhang2013:
Here is the document for this filter, http://tomcat.apache.org/tomcat-7.0-doc/config/filter.html#Add_Default_Character_Set_Filter,
base on the code, this should work, you might want to debug your code to see what's going on there.

ukari:
Here is the example

2015-12-18 20:12 GMT+08:00 Wenjie Zhang <notifications@github.com>:

violetagg:
I meant an example application which we can use to reproduce the issue that you are describing.

ukari:
I put that example in email, so you can't see it in github.
https://drive.google.com/file/d/0BxfFe2h4UJX4OEdMZ2dZMTBvYzBvOEl5Ykx4WUxNVWJiRmpj/view?usp=sharing

violetagg:
Ok,

I understand now ...

You misunderstood the usage of that filter ...

It is specified in the documentation [1]:
"encoding - Name of the character set which should be set, if no other character set was set explicitly by a Servlet."

So for example if you specify the following `response.setContentType("text/html");` in your servlet `UsersController` then the specified encoding `UTF-8` will be used and not `ISO-8859-1`.

Regards,
Violeta

[1] http://tomcat.apache.org/tomcat-8.0-doc/config/filter.html#Add_Default_Character_Set_Filter

wenjiezhang2013:
@violetagg is right, it clearly says so from the code

 ```java
@Override
        public void setContentType(String ct) {

            if (ct != null && ct.startsWith("text/")) {
                if (ct.indexOf("charset=") < 0) {
                    super.setContentType(ct + ";charset=" + encoding);
                } else {
                    super.setContentType(ct);
                    encoding = getCharacterEncoding();
                }
            } else {
                super.setContentType(ct);
            }

        }
```

**Comment by ukari** [comment]
I get it, but it won't work for application/json due to this line 
```java
  if (ct != null && ct.startsWith("text/")) {
```

wenjiezhang2013:
If you read ```The HTTP specification is clear that if no character set is specified for media sub-types of the "text" media type, the ISO-8859-1 character set must be used. However, browsers may attempt to auto-detect the character set. This may be exploited by an attacker to perform an XSS attack. Internet Explorer and other browsers have an option to enable this behavior.``` from http://tomcat.apache.org/tomcat-8.0-doc/config/filter.html#Add_Default_Character_Set_Filter, you will be able to understand the problem

ukari:
Thank both of you very much!

PR #28, opened 2016-02-15
Generalize the precompressed content serving from DefaultServlet

gmokki:
Replaces all gzip specific handling with more generic precompressed content handling.
More specifically, if the new precompressed parameter of DefaultServlet is set to true also brotli compressed (.br) files will be served in addition to gzipped files.

The old gzip parameter functionality is preserved for backwards compatibility, but removed from documentation.

The defining of new precompression file formats as parameter to the DefaultServlet can be added at later time.


markt-asf:
Generally I like this patch. I do have a couple of comments:
- I think it makes sense to make the change to a generic approach immediately. One way of doing this would be for the precompressed init param to take a comma separated list like ".gz,gzip,.br,br". It would need suitable sanity checking.
- The preference for different encodings is set by the client, not the server. See RFC7231, section 5.3.4. Look at o.a.catalina.Request.parseLocales() for a way of handling q values.

gmokki:
Thank you for the review comments.

Modified pull request to accept and pare "gzip=.gz,br=.br" syntax in the precompressed init parameter.
Honors the full client preference settings from client as per RFC7231, including * and identity types.

PR #33, opened 2016-07-28
Clear up confusion about mbeans-descriptors.xml in docs; fix a type i…

rhusar:
…n core descriptor

Lets make sure its everywhere called "mbeans-descriptors" otherwise one missing "s" can cost developers quite a while debugging what is wrong.

### DIFF
--- a/java/org/apache/catalina/core/mbeans-descriptors.xml
+++ b/java/org/apache/catalina/core/mbeans-descriptors.xml
@@ -34,7 +34,7 @@
                 writeable="false"/>
 
      <attribute name="filterInitParameterMap"
-                description="Return the initiaization parameters associated with this filter"
+                description="Return the initialization parameters associated with this filter"
                 type="java.util.Map"
                 writeable="false" />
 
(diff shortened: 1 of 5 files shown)


violetagg:
Hi,

I have one comment about renaming the page.
If there are bookmarks to that page with this change we will break these bookmarks.

Otherwise the change is OK.

Wdyt?

Regards,
Violeta

violetagg:
Hello,

Typos are fixed with r1755675.

Regards,
Violeta

markt-asf:
Rename the page. If someone complains, we can look at adding a redirect although we'll need to handle both Tomcat (for releases) and httpd (for the website). I'd prefer to have a single solution for both but I don;t see one at this point.

rhusar:
My thinking is that first, we need to stop perpetuating the wrong name "mbeans-descriptor" which is the name of the page, and more importantly, the page gives such a bad advice that it  deserves a rename since nobody could have been using it.

violetagg:
Hi,

Your patch was applied with r1755717.

Regards,
Violeta

rhusar:
👍

PR #34, opened 2016-09-28
fixed retrieval org.apache.catalina.startup.EXIT_ON_INIT_FAILURE

alostale:
This property was incorrectly tried to be retrieved: it converted the
  proprety name directly into boolean instead of its value from System
  properties.

### DIFF
--- a/java/org/apache/catalina/startup/Catalina.java
+++ b/java/org/apache/catalina/startup/Catalina.java
@@ -605,7 +605,7 @@ public class Catalina {
         try {
             getServer().init();
         } catch (LifecycleException e) {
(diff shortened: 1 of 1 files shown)


SvetlinZarev:
Nope, Boolean.getBoolean("xxx") tries to convert the value of the system property "xxx" to a boolean or returns false if it's missing. The method name is indeed very misleading, but works as expected.

http://docs.oracle.com/javase/7/docs/api/java/lang/Boolean.html#getBoolean(java.lang.String)

alostale:
Sorry, I misread the docs :(

PR #39, opened 2017-01-13
Provide the correct port when using proxy

SvetlinZarev:
Fix "create-connection" failures in case of proxied connection.
The proxy needs the remote port information in order to know
to which port to open a tunnel.

### DIFF
--- a/java/org/apache/tomcat/websocket/WsWebSocketContainer.java
+++ b/java/org/apache/tomcat/websocket/WsWebSocketContainer.java
@@ -236,18 +236,19 @@ public class WsWebSocketContainer implements WebSocketContainer, BackgroundProce
             }
         }
 
+        //If the port is not explicitly specified, compute it based on the scheme
+        if(port == -1){
+            if("ws".equalsIgnoreCase(scheme)){
+                port = 80;
+            }else {
+                // Must be wss due to scheme validation above
+                port = 443;
+            }
+        }
+
         // If sa is null, no proxy is configured so need to create sa
         if (sa == null) {
-            if (port == -1) {
-                if ("ws".equalsIgnoreCase(scheme)) {
-                    sa = new InetSocketAddress(host, 80);
-                } else {
-                    // Must be wss due to scheme validation above
-                    sa = new InetSocketAddress(host, 443);
-                }
-            } else {
-                sa = new InetSocketAddress(host, port);
-            }
+            sa = new InetSocketAddress(host, port);
         } else {
             proxyConnect = createProxyRequest(host, port);
         }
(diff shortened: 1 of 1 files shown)


ChristopherSchultz:
This looks like a no-op change. Does the CONNECT verb not imply a default port number when none is present?

SvetlinZarev:
It's not a no-op. Imagine the following case:

WS-Host: wss://echo.websocket.org
proxy-host: proxy
proxy-port: 8080

The flow is as follows:

URI path = the WS-HOST
String host = path.getHost();
int port = path.getPort(); //-1 if the ":port" part of the authority is missing

if(hasProxy){
            proxyConnect = createProxyRequest(host, port);
}else{
           if(-1 == port) then do portDiscovery -> 80/443 depending on ws/wss scheme
          ...
}

So here in the proxy case the port will be -1, because WS-Host does not specify host. As a result the CONNECT request will not contain port in the request line -> hence invalid CONNECT request. 

In the non-proxy case the original code correctly performs port discovery based on the scheme( if a port is not explicitly  provided in WS-Host).

So what I did was to do port discovery if no port is specified, regardless of proxy configuration.

SvetlinZarev:
Issue Description
===

WsWebSocketContainer generates bad CONNECT request when a proxy is used. The host does not contain port information, hence the app cannot connect to wss:// endpoints. Also some stricter proxy servers may reject requests without port info, thus making WsWebSocketContainer.connectToServer() fail for both ws:// and wss:// endpoints.

Steps to reproduce
===

Prerequisites
---
1. Download Tomcat 8.5.9
2. Configure it to use port 8081
3. Clone git@github.com:adamfisk/LittleProxy.git
4. Clone MCVE: git@github.com:SvetlinZarev/websocket-tomcat-bug.git
5. Build (3), (4) -> mvn clean install (you may want to rename the mcve to ROOT)

How to reproduce it
---
1. Start LittleProxy (it defaults to port 8080)
2. Configure Tomcat use the proxy: export CATALINA_OPTS="-Dhttp.proxyHost=localhost -Dhttp.proxyPort=8080 -Dhttps.proxyHost=localhost -Dhttps.proxyPort=8080"
3. Start tomcat: catalina.sh run
4. Request the mcve: http://localhost:8081/ws

Expected behavior
---
You should see: 

Connected. Sending message...
Message: Hello World!
Disconnected: 

Actual behavior
---
An exception is thrown. In case of LittleProxy it is a timeout exception during channel.handshake().get(); In my other case it is a DeploymentException caused by a bad request, because the (stricter) proxy rejected the request because  it does not contain port info.

Additional information
---
You can play with the mcve by adding the following request parameters:
* scheme - i.e. ws://, wss://
* server - i.e. echo.websocket.org

Example: 4. Request the mcve with: http://localhost:8081/ws?scheme=wss://&server=echo.websocket.org:443
It will succeed. Now remove the port from teh server string. It will fail.

SvetlinZarev:
> Does the CONNECT verb not imply a default port number when none is present?

@ChristopherSchultz According to [rfc2817 section 5.2](https://www.ietf.org/rfc/rfc2817.txt):

>  A CONNECT method requests that a proxy establish a tunnel connection
>    on its behalf. The Request-URI portion of the Request-Line is always
>    an 'authority' as defined by URI Generic Syntax [2], which is to say
>    **the host name and port number** destination of the requested connection
>    separated by a colon

Also by the never finalized draft https://tools.ietf.org/html/draft-luotonen-web-proxy-tunneling-01#section-3.1:

> The client connects to the proxy server, and uses the CONNECT method
>    to specify the hostname and the port number to connect to.  The
>    hostname and port number are separated by a colon, and **both of them
>    must be specified**.

kkolinko:
Please file this issue into Bugzilla, as documented at
http://tomcat.apache.org/bugreport.html

So that it has proper visibility and can be linked to from a changelog file. (Do you read changelogs?)

SvetlinZarev:
Hi, I've already created a bugzilla ticket: https://bz.apache.org/bugzilla/show_bug.cgi?id=60617

> Do you read changelogs?

I do.

kkolinko:
OK, thank you.

PR #41, opened 2017-02-05
Check and update session record if one already exists.

JohnKiel:
Check and update session record if one already exists, instead of deleting and re-inserting new record.

### DIFF
--- a/java/org/apache/catalina/session/JDBCStore.java
+++ b/java/org/apache/catalina/session/JDBCStore.java
@@ -158,9 +158,11 @@ public class JDBCStore extends StoreBase {
     protected PreparedStatement preparedSizeSql = null;
 
     /**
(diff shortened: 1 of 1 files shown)


ChristopherSchultz:
You should use SELECT ... FOR UPDATE instead of SELECT / UPDATE. Much less code and transactionally-safe. Also, if you use SELECT ... FOR UPDATE, you can use ResultSet.moveToInsertRow to insert a record if necessary.

JohnKiel:
If you use SELECT FOR UPDATE, then you'll also need to start and commit a transaction, with the overhead that entails.  If the record is modified or even deleted between the time the SELECT query and UPDATE, it's of little to no consequence, and no worse than the current delete/insert method.  The worst result would be slightly stale session data being written, which is kind of an issue anyway if you have an application with that many threads all trying to save data for the same session at basically the same time.

PR #46, opened 2017-03-03
Bug_60808: ensure immutability of ParameterMap.

woonsan:


woonsan:
I'll propose a new PR using containment, not inheritance, as @markt-asf suggested.

PR #47, opened 2017-03-03
Make ParameterMap immutable using containment instead of inheritance.

woonsan:


markt-asf:
Patch applied. Thanks.

PR #49, opened 2017-03-31
Fix parser to fail if leading zeros in IPv4 part of IPv6 address

katya-stoycheva:
Leading zeros are not causing failure if in IPv4 addresses though.


markt-asf:
Thanks for the patch. Sorry it took a while to apply it. Due to our delays I had to adapt things a little.
Fixed in:
- trunk for 9.0.9 onwards
- 8.5.x for 8.5.32 onwards
- 8.0.x for 8.0.53 onwards
- 7.0.x for 7.0.89 onwards

PR #51, opened 2017-04-01
#41 Check and update session record if one already exists.

alexisgayte:
- Closed the Statement.
- Update followed by insert if row==0 to minimize request.


alexisgayte:
Hi Guys,
I could not run the test but I use this code for my tomcat8.
That will fix the issue #41.

Also remove and create the session create some issue if you want to link the PK in the database.

ChristopherSchultz:
Might this be better written as `SELECT ... FOR UPDATE` with either `updateRow` or `ResultSet.insert` if there are no results from the `SELECT`?

alexisgayte:
I can see the point to lock the table in case of load balancing.
But the previous version was deleting and inserting only.
Will it not put down the perf as 90% are update?

PR #54, opened 2017-04-24
Respect minimum pool size [#61032]

jelmerterwal:
https://bz.apache.org/bugzilla/show_bug.cgi?id=61032

### DIFF
--- a/modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/ConnectionPool.java
+++ b/modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/ConnectionPool.java
@@ -620,7 +620,8 @@ public class ConnectionPool {
         // we've asynchronously reduced the number of connections
         // we could have threads stuck in idle.poll(timeout) that will never be
         // notified
-        if (waitcount.get() > 0) {
+        // we also want to respect the minimum number of idle connections
+        if (idle.size() + busy.size() < getPoolProperties().getMinIdle() || waitcount.get() > 0) {
             idle.offer(create(true));
         }
     }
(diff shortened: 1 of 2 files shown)


jelmerterwal:
For an application I develop, I see a drop in connections. The screenshot show different values of the connection pool.

Legend for the colours;
- grey -> pool size
- greenish -> idle connections
- orange/red -> connections in use
- purple -> waiting for a connection

screen shot 2017-04-24 at 13 49 43

The application uses dropwizard (www.dropwizard.io) which internally uses Tomcat's JDBC Pool. The application has 4 instances. The graph shows the sum of these instances for the number of connections.

- min size of idle connection is set to 10 (so should be 10 x 4 = 40)
- the decrease happens due abandoning of connections

The documentation indicates a minimum pool of idle connections should be available. It scared me when I first saw this happening.

```
minIdle

(int) The minimum number of established connections that should be kept in the pool at all times. The connection pool can shrink below this number if validation queries fail. Default value is derived from initialSize:10 (also see testWhileIdle)
```

jelmerterwal:
Below an image of further progress during the day. Where graphs peek at 40 the application has been restarted.

screen shot 2017-04-24 at 18 48 00

jelmerterwal:
@markt-asf could you have a look at this PR?

modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/ConnectionPool.java
```diff
@@ -622,6 +622,11 @@ protected void release(PooledConnection con) {
         // notified
         if (waitcount.get() > 0) {
             idle.offer(create(true));
+
+        // we also want to respect the minimum number of idle connections
+        // (if the connection pool is closing do not create new connections!)
+        } else if (idle.size() + busy.size() < getPoolProperties().getMinIdle() && !isClosed()) {
+            idle.offer(create(true));
```
sakalaramesh: filling idle queue I think should be part of the PoolCleaner instead of being part of the release(), especially if poolSweeper is enabled. release() is invoked while application request is waiting most of the time. 

Also, minIdle is for number of idle connections, not the total of idle+busy connections. Shouldn't we be filling idle queue when idle queue size falls below minIdle independent of busy queue size?

PR #56, opened 2017-05-28
Convert Cluster Manager human-readable channelSendOptions names to integer [61127/8]

isapir:
This is the basis for tickets
https://bz.apache.org/bugzilla/show_bug.cgi?id=61127
https://bz.apache.org/bugzilla/show_bug.cgi?id=61128
With the added static method `int Channel.getOptionsFromString(String input)`

Calling that method with a String input would translate it to the the equivalent bitwise-ORd integer of the options in the comma-separated input, so that for example `Channel.getOptionsFromString("async,multicast")` would return `72`.

I am not sure yet where the value is read from the configuration file, but it would be simple to use the methods added here, e.g. in any Class that already has 

```
    public void setChannelSendOptions(int channelSendOptions) {
        this.channelSendOptions = channelSendOptions;
    }
```

We can add

```
    public void setChannelSendOptions(String channelSendOptions) {
        this.setChannelSendOptions( Channel.getOptionsFromString(channelSendOptions) );
    }
```

Or perhaps simply translate the input in the method(s) that calls `setChannelSendOptions(int channelSendOptions)` without modifying the classes themselves.

### DIFF
--- a/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java
+++ b/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java
@@ -400,6 +400,14 @@ public class SimpleTcpCluster extends LifecycleMBeanBase
     }
 
     /**
+     * returns the SendOptions as a comma separated list of names for use by JMX
+     * @return
+     */
+    public String getChannelSendOptionsName(){
+        return Channel.getSendOptionsAsString(channelSendOptions);
+    }
+
+    /**
      * Create new Manager without add to cluster (comes with start the manager)
      *
      * @param name
--- a/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java
+++ b/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java
@@ -401,7 +401,7 @@ public class SimpleTcpCluster extends LifecycleMBeanBase
 
     /**
      * returns the SendOptions as a comma separated list of names for use by JMX
-     * @return
+     * @return a comma separated list of the option names
      */
     public String getChannelSendOptionsName(){
         return Channel.getSendOptionsAsString(channelSendOptions);
--- a/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java
+++ b/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java
@@ -339,6 +339,13 @@ public class SimpleTcpCluster extends LifecycleMBeanBase
         this.channelSendOptions = channelSendOptions;
     }
 
+    public void setChannelSendOptions(String channelSendOptions) {
+
+        int value = Channel.parseChannelSendOptions(channelSendOptions);
+        if (value > 0)
+            this.setChannelSendOptions(value);
+    }
+
     /**
      * has members
      */
--- a/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java
+++ b/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java
@@ -341,7 +341,7 @@ public class SimpleTcpCluster extends LifecycleMBeanBase
 
     public void setChannelSendOptions(String channelSendOptions) {
 
-        int value = Channel.parseChannelSendOptions(channelSendOptions);
+        int value = Channel.parseSendOptions(channelSendOptions);
         if (value > 0) {
             this.setChannelSendOptions(value);
         }
--- a/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java
+++ b/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java
@@ -342,8 +342,9 @@ public class SimpleTcpCluster extends LifecycleMBeanBase
     public void setChannelSendOptions(String channelSendOptions) {
 
         int value = Channel.parseChannelSendOptions(channelSendOptions);
-        if (value > 0)
+        if (value > 0) {
             this.setChannelSendOptions(value);
+        }
     }
 
     /**
--- a/java/org/apache/catalina/tribes/Channel.java
+++ b/java/org/apache/catalina/tribes/Channel.java
@@ -369,4 +369,52 @@ public interface Channel {
      */
     public void setName(String name);
 
+    /**
+     * Translates the name of an option to its integer value
+     * @param opt The name of the option
+     * @return
+     */
+    public static int getOptionValue(String opt){
+
+        if ("asynchronous".equals(opt) || "async".equals(opt))
+            return SEND_OPTIONS_ASYNCHRONOUS;
+
+        if ("byte_message".equals(opt) || "byte".equals(opt))
+            return SEND_OPTIONS_BYTE_MESSAGE;
+
+        if ("multicast".equals(opt))
+            return SEND_OPTIONS_MULTICAST;
+
+        if ("secure".equals(opt))
+            return SEND_OPTIONS_SECURE;
+
+        if ("synchronized_ack".equals(opt) || "sync".equals(opt))
+            return SEND_OPTIONS_SYNCHRONIZED_ACK;
+
+        if ("udp".equals(opt))
+            return SEND_OPTIONS_UDP;
+
+        if ("use_ack".equals(opt))
+            return SEND_OPTIONS_USE_ACK;
+
+        return 0;
+    }
+
+    /**
+     * Translates a comma separated list of option names to their bitwise-ORd value
+     * @param input A comma separated list of options, e.g. "async,multicast"
+     * @return
+     */
+    public static int getOptionsFromString(String input){
+
+        int result = 0;
+
+        String[] options = input.split(",");
+
+        for (String opt : options)
+            result |= getOptionValue(opt);
+
+        return result;
+    }
+
 }
--- a/java/org/apache/catalina/tribes/Channel.java
+++ b/java/org/apache/catalina/tribes/Channel.java
@@ -435,7 +435,6 @@ public interface Channel {
         return result;
     }
 
-
     /**
      * Translates an integer value of SendOptions to its human-friendly comma separated value list for use in JMX and such.
      * @param input the int value of SendOptions
@@ -449,8 +448,8 @@ public interface Channel {
         StringJoiner names = new StringJoiner(", ");
         for (int bit=allOptionNames.length - 1; bit >= 0; bit--){
 
-            // if (2^bit & input) yields a positive value then the bit is set
-            if (((int)Math.pow(2, bit) & input) > 0){
+            // if the bit is set then add the name to the result
+            if (((1 << bit) & input) > 0){
                 names.add(allOptionNames[bit]);
             }
         }
--- a/java/org/apache/catalina/tribes/Channel.java
+++ b/java/org/apache/catalina/tribes/Channel.java
@@ -410,15 +410,15 @@ public interface Channel {
         try {
 
             return Integer.parseInt(input);
-        }
-        catch (NumberFormatException nfe){}
+        } catch (NumberFormatException nfe){}
 
         int result = 0;
 
         String[] options = input.split(",");
 
-        for (String opt : options)
+        for (String opt : options) {
             result |= getOptionValue(opt);
+        }
 
         return result;
     }
--- a/java/org/apache/catalina/tribes/Channel.java
+++ b/java/org/apache/catalina/tribes/Channel.java
@@ -378,7 +378,7 @@ public interface Channel {
      * @param opt The name of the option
      * @return
      */
-    public static int getOptionValue(String opt){
+    public static int getSendOptionValue(String opt){
 
         switch (opt){
 
@@ -415,7 +415,7 @@ public interface Channel {
      * @param input A comma separated list of options, e.g. "async,multicast"
      * @return
      */
-    public static int parseChannelSendOptions(String input){
+    public static int parseSendOptions(String input){
 
         try {
             return Integer.parseInt(input);
@@ -428,7 +428,7 @@ public interface Channel {
 
         int result = 0;
         for (String opt : options) {
-            result |= getOptionValue(opt);
+            result |= getSendOptionValue(opt);
         }
 
         return result;
--- a/java/org/apache/catalina/tribes/Channel.java
+++ b/java/org/apache/catalina/tribes/Channel.java
@@ -405,7 +405,13 @@ public interface Channel {
      * @param input A comma separated list of options, e.g. "async,multicast"
      * @return
      */
-    public static int getOptionsFromString(String input){
+    public static int parseChannelSendOptions(String input){
+
+        try {
+
+            return Integer.parseInt(input);
+        }
+        catch (NumberFormatException nfe){}
 
         int result = 0;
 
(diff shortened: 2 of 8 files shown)


java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java
```diff
@@ -339,6 +339,13 @@ public void setChannelSendOptions(int channelSendOptions) {
         this.channelSendOptions = channelSendOptions;
     }
 
+    public void setChannelSendOptions(String channelSendOptions) {
+
+        int value = Channel.parseChannelSendOptions(channelSendOptions);
+        if (value > 0)
```
violetagg: Hi,

From code format point of view: We use curly braces even when there is only one statement in "if", "for" blocks.

Regards,
Violeta

java/org/apache/catalina/tribes/Channel.java
```diff
@@ -369,4 +369,58 @@
...
+     * @return
+     */
+    public static int parseChannelSendOptions(String input){
+
+        try {
+
+            return Integer.parseInt(input);
+        }
+        catch (NumberFormatException nfe){}
+
```
violetagg: Hi,

From code format point of view: Keep the closing curly brace with "catch" on the same line.

Regards,
Violeta

java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java
isapir: Violeta, thank you for your feedback.

I'll add curly braces to the `for` block.

So each of those if's should have curly braces?

```if ("asynchronous".equals(opt) || "async".equals(opt))
            return SEND_OPTIONS_ASYNCHRONOUS;```

That's a lot of curly braces...  Perhaps a `switch` statement would be "cleaner" here then?

**Comment by isapir** [comment]
@violetagg Thank you for your feedback, I updated the `for`, `catch`, and one of the `if`'s.

While looking into updating the rest of the `if` statements I noticed instances of the code where curly braces were omitted for single-statement `if`s, e.g.

https://github.com/isapir/tomcat/blob/e51512fdc062ffd07a3dd323cba76ab057d23322/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java#L256

https://github.com/isapir/tomcat/blob/e51512fdc062ffd07a3dd323cba76ab057d23322/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java#L424

https://github.com/isapir/tomcat/blob/e51512fdc062ffd07a3dd323cba76ab057d23322/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java#L470

https://github.com/isapir/tomcat/blob/e51512fdc062ffd07a3dd323cba76ab057d23322/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java#L500

https://github.com/isapir/tomcat/blob/e51512fdc062ffd07a3dd323cba76ab057d23322/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java#L503

https://github.com/isapir/tomcat/blob/e51512fdc062ffd07a3dd323cba76ab057d23322/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java#L537

https://github.com/isapir/tomcat/blob/e51512fdc062ffd07a3dd323cba76ab057d23322/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java#L546

https://github.com/isapir/tomcat/blob/e51512fdc062ffd07a3dd323cba76ab057d23322/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java#L564

https://github.com/isapir/tomcat/blob/e51512fdc062ffd07a3dd323cba76ab057d23322/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java#L565

https://github.com/isapir/tomcat/blob/e51512fdc062ffd07a3dd323cba76ab057d23322/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java#L570

https://github.com/isapir/tomcat/blob/e51512fdc062ffd07a3dd323cba76ab057d23322/java/org/apache/catalina/ha/tcp/SimpleTcpCluster.java#L621

Anyway, it seemed like I shouldn't even continue in that file.  If you want to enforce these styles then perhaps it'd be better to run the source code through a code formatter.

Thanks,


Igal

**Comment by ChristopherSchultz** [comment]
A few comments:

1. I think the patch should include some TRACE-level logging, especially when catching the NumberFormatException.
2. If there is an unrecognized option string, it should at least result in a WARN message emitted to the log. I might even lobby for an exception to be thrown.
3. There are options that seem to be in conflict with each other, if only in terminology: e.g. "async" and "sync" can be specified together, which would read like a bug in the config.
4. Splitting on a comma should include optional whitespace e.g. "\\s,*\\s*" as the pattern.
5. The patch needs to include documentation for the string-based configuration options (webapps/docs/cluster-howto.html).
6. The javadoc in the patch needs to include the list of acceptable string option-names.

**Comment by violetagg** [comment]
> I noticed instances of the code where curly braces were omitted for single-statement ifs, 

Yep, we know that. If you take a look at the history you will see that typically when we prepare changes first make a change with formatting if we see that it is needed and then implement the change.
We use 
- the standard Eclipse formatting  
- spaces only 
- lines up to 100 chars for code and 80 chars for javadoc, comments etc. 
- some additional settings for the compiler and javadoc http://svn.apache.org/repos/asf/tomcat/trunk/res/ide-support/eclipse/java-compiler-errors-warnings.txt

**Comment by isapir** [comment]
@violetagg Yes, I read that on http://tomcat.apache.org/getinvolved.html#Coding_Conventions

I was pointing out these examples while trying to figure out if you still require that I wrap each single statement `if` with braces, which would make the code less readable IMO.

**Comment by isapir** [comment]
@ChristopherSchultz That's exactly the kind of feedback that I was looking for.  I will address the points to the best of my ability.

Keep in mind that after those changes, the commit will be much larger and take longer to review, which raises the concerns that I mentioned in the mailing list.

**Comment by violetagg** [comment]
> I was pointing out these examples while trying to figure out if you still require that I wrap each single statement if with braces, which would make the code less readable IMO.

+1 for `switch` as you proposed earlier

**Comment by ChristopherSchultz** [comment]
@isapir I appreciate your core patch as an initial effort. It looks like it will work. In order for it to be a good patch, it needs to be expanded, so a larger patch is appropriate.

Did you actually test it to see if it would give you the desired channel options?

A small unit test wouldn't hurt, as well (just testing setChannelSendOptions(String) -> getChannelSendOptions:int would be fine).

**Comment by isapir** [comment]
> Did you actually test it to see if it would give you the desired channel options?
I actually did

> A small unit test wouldn't hurt
Will do

**Comment by isapir** [comment]
@violetagg I refactored the method to use `switch` statement as advised. https://github.com/apache/tomcat/pull/56/files#diff-5c721c838c78fa7c31f9eb62c27863ceR383

@ChristopherSchultz 

>I think the patch should include some TRACE-level logging, especially when catching the NumberFormatException.

Added

>If there is an unrecognized option string, it should at least result in a WARN message emitted to the log. I might even lobby for an exception to be thrown.

Throwing IllegalArgumentException

>There are options that seem to be in conflict with each other, if only in terminology: e.g. "async" and "sync" can be specified together, which would read like a bug in the config.

If we add this then it should go into setChannelSendOptions() so that we also check if two conflicting integer options are passed then we warn/throw.  I can add that, but it should be a separate ticket/PR IMO.

>Splitting on a comma should include optional whitespace e.g. "\s,\s" as the pattern.

Changed regex to `\s*,\s*` 

>The patch needs to include documentation for the string-based configuration options (webapps/docs/cluster-howto.html).

Added

>The javadoc in the patch needs to include the list of acceptable string option-names.

Added

>A small unit test wouldn't hurt, as well (just testing setChannelSendOptions(String) -> getChannelSendOptions:int would be fine).

Added, **but I couldn't run the test case** in my IDE.  getting an error `Error:(33, 16) java: package trailers does not exist` for `test\org\apache\coyote\http2\TestStream.java` and other files and I'm not sure yet how to resolve the dependencies.  where is package `trailers` coming from?

**Comment by ChristopherSchultz** [comment]
How would you like your name to appear in the changelog? "Patch provided by Igal Sapir"?

**Comment by isapir** [comment]
That's fine.

**Comment by KeiichiFujino** [comment]
Hi,

some comments : 

- SimpleTcpCluster exposes the configuration via JMX.
If you configure channelSendOptions in string form, you should also expose it to JMX in string format.
You can see org/apache/catalina/ha/tcp/mbeans-descriptors.xml

- It is necessary to fix configuration docs (webapps/config/cluster.xml).

- The patch needs to include fix for mapSendOptions.
BackupManager use mapSendOptions instead of channelSendOptions.

**Comment by ChristopherSchultz** [comment]
@KeiichiFujino Is this patch acceptable as-is, with additional patches for your items listed above? I think it's okay to add this feature only to channelSendOptions as a first commit. Adding mapSendOptions and JMX support can come later. Okay?

**Comment by isapir** [comment]
>It is necessary to fix configuration docs (webapps/config/cluster.xml).

Added

>The patch needs to include fix for mapSendOptions.  BackupManager use mapSendOptions instead of channelSendOptions.

Added

**Comment by isapir** [comment]
>If you configure channelSendOptions in string form, you should also expose it to JMX in string format.
You can see org/apache/catalina/ha/tcp/mbeans-descriptors.xml

I updated the type from `int` to `String`.  I see that the current implementation does the same for `channelStartOptions` which is also an `int`, but set as a `String` in the mbeans-descriptor, so I expect it to work the same even though TBH I have not tested with JMX.

**Comment by isapir** [comment]
Can we merge this or do we need to wait for @KeiichiFujino 's input?  I believe that I addressed all of the issues that he pointed out.

**Comment by KeiichiFujino** [comment]
Sorry for the late reply.

I do not think it work as expected.
Even if updating the type of channelStartOptions from int to String in mbeans-descriptor, 
it will display int format values.
Because the channelSendOptions is always set to int format.
if you set channelSendOptions = "asynchronous", 8 will be displayed  via JMX.

For example, 
The channelSendOptions in the String format specified by the argument must be saved under a different name(e.g. channelSendOptionsName) and published to JMX.
Then add the following definition to mbeans-descriptor.

`
<attribute
  name="channelSendOptionsName"
  description="channelStartOptions name."
  writeable="false"    
  type="java.lang.String"/>
`

**Comment by ChristopherSchultz** [comment]
@KeiichiFujino Okay, so we'd have a synthetic mbean attribute that we can re-construct from the int value we store internally? Or, do you think we should save the string that was used in configuration?

**Comment by isapir** [comment]
>Even if updating the type of channelStartOptions from int to String in mbeans-descriptor,
it will display int format values.
Because the channelSendOptions is always set to int format.
if you set channelSendOptions = "asynchronous", 8 will be displayed via JMX.

Right, because we are not changing the way that the value is stored, we are only adding a way to make it easier to set it.  You will still be able to change the value (not sure if that's even permitted at runtime) via JMX with either `String` or `int` value, but it will be displayed in its `int` form.

This is the way that it is already implemented for `channelStartOptions`

        <attribute
      name="channelStartOptions"
      description="This sets channel start behaviour."
      type="java.lang.String"/>

Which is also an `int`, but specified as `String` in the mbeans-descriptor, and TBH, it will be inconsistent if other similar values will be displayed as `int` and this one will be displayed as `String`.

I can write a method to translate back from `int` to `String`, but IMO this is really unnecessary, as it is an advanced option which you don't normally change at runtime, and if you need to troubleshoot an issue then you know what value you expect to see.

Again, setting the value via JMX should work now with either `String` or `int` input.  It's only the way that it is displayed that remains an `int`.

**Comment by isapir** [comment]
@KeiichiFujino @ChristopherSchultz I added:

* a method to translate the int value to a string list:
https://github.com/apache/tomcat/pull/56/files#diff-5c721c838c78fa7c31f9eb62c27863ceR439

* `public String getChannelSendOptionsName()`
https://github.com/apache/tomcat/pull/56/files#diff-8e81c1ebb6efdca13082a2798a305251R403

* a read-only `channelSendOptionsName` in mbeans-descriptor
https://github.com/apache/tomcat/pull/56/files#diff-9dca3596f50abfe90c0ed9b54d4e2092R32

* a test-case that checks the string representation
https://github.com/apache/tomcat/pull/56/files#diff-a638e8120005efa2cccdf63066b7fde5R53

I believe that this addresses all of the issues with this PR.

**Comment by ChristopherSchultz** [comment]
@isapir FYI `1 << bit` is the same as `(int)Math.pow(2, bit)` and probably slightly more efficient. Better yet, it's less code to read. No particular reason to change the patch. I'm already happy with what we've got.

**Comment by isapir** [comment]
I'll change it.  Please wait.

**Comment by ChristopherSchultz** [comment]
I'm waiting on a final review from @KeiichiFujino before I commit.

In the meantime there are 2 `@return` javadoc annotations with no value. Could you complete those?

**Comment by ChristopherSchultz** [comment]
Thanks @Igal for your patience with this PR review. I know we've been a bit of a pain. But this work will make the contribution that much more useful.

**Comment by isapir** [comment]
@ChristopherSchultz I updated the patch according to your feedback: https://github.com/apache/tomcat/pull/56/commits/f0c3968570f70823d30cf18144e82591a7ee0cee

I was actually trying at first to get it to work with a bit shift but when it didn't work I resorted to  `Math.pow(2, bit)` -- bit shift is definitely better, though it's possible that `Math.pow(2, x)` utilizes shift in the JRE anyway.

In any event, great feedback, thanks!

**Comment by KeiichiFujino** [comment]
There are two comments.

you need to add BackupManager mbean definitions.
```
<attribute
      name="mapSendOptionsName"
      description="mapSendOptions name."
      writeable="false"
      type="java.lang.String"/>
```
in org/apache/catalina/ha/session/mbeans-descriptors.xml

There is a typo in 
```
<attribute
      name="channelSendOptionsName"
      description="channelStartOptions name."
      writeable="false"
      type="java.lang.String"/>
```
"channelStartOptions name."  -> "channelSendOptions name."

isapir:
@KeiichiFujino Thank you for your prompt review.

>you need to add BackupManager mbean definitions ... in org/apache/catalina/ha/session/mbeans-descriptors.xml

Thank you for providing the path.  I was looking for that and didn't find it easily.

>There is a typo in ... "channelStartOptions name." -> "channelSendOptions name."

Corrected.  I actually copied and pasted it from your comment above --
 https://github.com/apache/tomcat/pull/56#issuecomment-307055014 -- and simply added new line characters, but I should have caught that ;)

Can you please confirm that as far as you're concerned this can now be merged?  @ChristopherSchultz is awaiting your approval.

Thank you :)

isapir:
@KeiichiFujino p.s. This is the last commit that addresses the issues above: https://github.com/apache/tomcat/pull/56/commits/192d2eb13e9ec9448e183dcc3b166f7d3577c250

KeiichiFujino:
I confirmed. there is no problem.
Thanks.

isapir:
@ChristopherSchultz can we merge this please?

ChristopherSchultz:
Committed to tomcat-trunk in ASF svn r1799462. Back-ports to follow.

isapir:
Thank you 👍

ChristopherSchultz:
I'm not sure why the commit message hasn't closed this issue. Any thoughts on why that might be?

violetagg:
There aren't any replications since yesterday :(

isapir:
Should I close it manually?

ChristopherSchultz:
Not quite yet. Let's see if we can get it closed automatically through the "proper" channels.

PR #58, opened 2017-06-05
add nioendpoint note

powerYao:


markt-asf:
Given the name of the method call, the proposed comment adds no value. The patch also uses tabs rather than spaces.

PR #59, opened 2017-06-05
add some useful note

powerYao:


markt-asf:
Most added comments add no value. The s/poller/pollers/ change is the only one worth considering and that is a trivial change.
The patch uses a mix of tabs and spaces for indents.

powerYao:
I get it,thanks

PR #60, opened 2017-06-06
Introduce new logging configuration per handler - maxDays.

violetagg:
This configuration gives the possibility to specify the maximum days
that the log files will be kept on the file system.
This can be applied on a handler level.
If the specified value is <= 0 then the log files will be kept forever,
otherwise they will be kept the specified maximum days and after that they will be deleted.
The default value is 90 days.


rmannibucau:
@violetagg hello Violeta, would it be interesting to merge it with tomee handler (https://github.com/apache/tomee/blob/master/tomee/tomee-juli/src/main/java/org/apache/tomee/jul/handler/rotating/LocalFileHandler.java) ?

violetagg:
The change was applied with commit https://github.com/apache/tomcat/commit/cb884bfdcb537edadd9185fbd8af126c72af4ba5

PR #63, opened 2017-06-10
added portOffset attribute to server.xml per BZ-61171

isapir:
https://bz.apache.org/bugzilla/show_bug.cgi?id=61171

### DIFF
--- a/java/org/apache/catalina/Server.java
+++ b/java/org/apache/catalina/Server.java
@@ -71,6 +71,20 @@ public interface Server extends Lifecycle {
     public javax.naming.Context getGlobalNamingContext();
 
 
+
+    /**
+     * @return the server's port offset value
+     */
+    public int getPortOffset();
+
+
+    /**
+     * @param portOffset the new port offset value
+     */
+    public void setPortOffset(int portOffset);
+
+
+
     /**
      * @return the port number we listen to for shutdown commands.
      */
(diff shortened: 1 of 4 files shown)


powerYao:
i feel it's not very necessary for tomcat.What if the port is already bound?

markt-asf:
I think more discussion is required on how to approach this. Specifically, when should the offset be applied? What is the impact on getters and setters? For example connector.setPort (getPort ()) should be a NO-OP. What is reported via JMX? There are all sorts of side effects that need to be thought through here else we'll end up creating some very non-intuative behaviour.

isapir:
>I think more discussion is required on how to approach this. Specifically, when should the offset be applied? 

@markt-asf I agree, and I was trying to find different options for this in order to initiate a discussion, but the way that Tomcat reads the config files and initializes objects via introspection really limits us here unless there is a major overhaul.

The process is somewhat like this:

1) `Connector`s are initialized (no reference to `Service` or `Server`)
2) `Service` is linked to the `Connector`s (still no reference to `Server`)
3) `Server` is added to Service

Point (3)  is point where I took the `portOffset` from Server and applied it to the `Connector`s.

>What is the impact on getters and setters? For example connector.setPort (getPort ()) should be a NO-OP. 

At point (1) above the Connector's setPort() is called by the IntrospectionUtils with the values from `server.xml`.  That can not be a NO-OP at that point as that would practically disable the Connector, but we can check if `Service` and `Server` are null or not (which they are at initialization) and perhaps do something different accordingly?

Is it even possible/allowed to change the port after initialization?

markt-asf:
I've added some comments to the bugzilla issue on how this might be addressed.

isapir:
Great, I will implement as advised on BZ.

PR #65, opened 2017-06-24
Fix bug #61214 : NoSuchMethodException in JMX Proxy (Context MBeans)

hasalex:
Remove servlets attribute in the ServletContext mbean description


markt-asf:
My favourite kind of patch. It fixes a problem by deleting stuff. :)

PR #66, opened 2017-06-24
Fix bug #61215 : NoSuchMethodException in JMX Proxy (RemoteAddrValve)

hasalex:


hasalex:
OK, just discovered the is="false" attribute in mbean-descriptors.xml.

PR #68, opened 2017-06-26
Fix #61224 - Make the GlobalRequestProcessor mbeans read-only

hasalex:


markt-asf:
Tomcat fails to compile after applying this patch.

Looking in to this, I don't see any reason for BaseModelMBean to throw exceptions in its constructor. As far as I can tell, the constructor (and the matching constructors on all the sub-classes) can be removed.

I'm going to look into this before coming back to this PR. No need to update the PR. The changes required are minor enough that I'll handle it on merge.

hasalex:
Thanks for fixing (and merging) the PR
It worked on my machine ;) I'll have to learn more on the build process.

PR #71, opened 2017-08-01
Fix #61366 - Set the TCCL to that of the webapp context in JDBCStore.open().

Invader444:
https://bz.apache.org/bugzilla/show_bug.cgi?id=61366


jhorowitz-firedrum:
Modified to include the recommendation from Mark Thomas: https://bz.apache.org/bugzilla/show_bug.cgi?id=61366#c3

PR #73, opened 2017-10-02
Bug 57767 - Websocket client proprietary configuration

joferna:
The original patch for redirection support (https://bz.apache.org/bugzilla/show_bug.cgi?id=57767)  is not compliant with the websocket spec since the redirectSet variable is not thread-safe. This pull request aims to fix the concurrency issue by creating a new class "WsWebSocketClient" with its own set of variables. It also includes the proposed changes for authentication support.


markt-asf:
Chris's original concern with the BZ 57767 patch (lack of Javadoc) still needs to be addressed.

rmaucher:
Well, it looks ok to me overall, so I'll add a bit of javadoc and merge it. Any issue with backporting it ?

markt-asf:
No objections to back-porting here.

PR #74, opened 2017-10-02
added javadoc comment

isapir:
Took me a few minutes to figure out what the method is doing (and how) so added a JavaDoc comment to clarify.

### DIFF
--- a/java/org/apache/catalina/filters/CorsFilter.java
+++ b/java/org/apache/catalina/filters/CorsFilter.java
@@ -75,7 +75,7 @@ import org.apache.tomcat.util.res.StringManager;
  * @see <a href="http://www.w3.org/TR/cors/">CORS specification</a>
  *
  */
-public class CorsFilter extends GenericFilter {
+public class CorsFilter extends FilterBase {
 
     private static final long serialVersionUID = 1L;
     private static final Log log = LogFactory.getLog(CorsFilter.class);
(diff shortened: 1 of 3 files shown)


isapir:
@markt-asf - What do you think about changing `FilterBase` so that it extends `javax.servlet.GenericFilter`?  That way we can use methods like `getInitParameter()` and have a single base class and "cleaner" code.

For example, ATM some filters extend `org.apache.catalina.filters.FilterBase` (e.g. ExpiresFilter, FailedRequestFilter) and others extend `javax.servlet.GenericFilter` (e.g. CorsFilter, RemoteIpFilter).

I can provide a patch if you agree.

markt-asf:
Generally, I'm in favour of refactoring that reduces the overall volume of code.

isapir:
I believe that this will both reduce the overall volume of code and make the code more maintainable moving forward.

I usually try to submit small patches so that they're easier to review and accept though, so the reduction of code might only be visible in the second step.  I will work on it a bit and see how it goes.

isapir:
I plan to refactor the rest of the Filters as I did for `CorsFilter`, but wanted you to see first what I meant.

isapir:
FYI:  It's easier to see the changeset via this link (adding `?w=1` to a GitHub PR URL ignores same-line whitespace changes):
https://github.com/apache/tomcat/pull/74/files?w=1

isapir:
TBH upon further inspection the benefit of this patch is limited since most filters do extend FilterBase.

While I do think that it provides cleaner code that can be helpful for future filters as well, I understand if you choose to reject it.

Best,

Igal

markt-asf:
There is definitely some value here although we might not choose to skip some parts of the patch. Let me pull in the obvious stuff and then we can see what is left.

isapir:
Sounds good.

If you have more specific guidelines in mind then I'd be happy to implement/refactor accordingly.  Please advise if so.

Thanks.

markt-asf:
I think I've pulled the obvious wins into Tomcat. Take a look and if you think there is merit in further changes please feel free to open another pull request.

isapir:
I saw the SVN commits -- thanks!

PR #75, opened 2017-10-02
added javadoc comments and a method that takes default value

isapir:
The `getInitParameter(String name, String defaultValue)` adds useful functionality to this base class so that any optional parameter can be set in one line of code instead of checking whether it is null or not, e.g.

    String value = deafultValue;
    if (getInitParameter(param) != null)
      value = getInitParameter(param);

or

    if (getInitParameter(param) != null)
      value = getInitParameter(param);
    else
      value = defaultValue;

becomes

    value = getInitParameter(param, defaultValue);

### DIFF
--- a/java/javax/servlet/GenericFilter.java
+++ b/java/javax/servlet/GenericFilter.java
@@ -32,6 +32,24 @@ public abstract class GenericFilter implements Filter, FilterConfig, Serializabl
     }
 
 
+    /**
+     * This method returns the parameter's value if it exists, or defaultValue if not.
+     *
+     * @param name - The parameter's name
+     * @param defaultValue - The default value to return if the parameter does not exist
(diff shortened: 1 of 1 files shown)


markt-asf:
-1
The public API of specification defined classes may not be changed.
This method would be worth adding to o.a.catalina.filters.FilterBase if any of Tomcat's internal filter implementations could make use of it (i.e. if it resulted in removing more code than it added)

isapir:
OK, the reason that I added this method is that I saw `init()` and I did not realize that it was part of the spec since it is not marked with `@Override` and I've never seen it before (looks like it's new in Servlet 4.0)

Should I re-submit the JavaDoc enhancements and add `@Override` to `init()`?

isapir:
Actually, I see that I can not add `@Override` since it's not part of the `javax.servlet.Filter` interface.

PR #76, opened 2017-10-05
added SessionInitializerFilter

isapir:
This Filter initializes the HttpSession object for the Request, which is required for certain operations with WebSocket requests where it is too late to initialize the HttpSession, but the HttpSession object is required to retrieve information during the Handshake process.

One common use case is to retrieve the `ServletContext` during a WebSocket Handshake by calling `handshakeRequest.getHttpSession().getServletContext()`.  According to the WebSocket specification a call to getHttpSession() should not initialize one if one does not exist.  If the HttpSession is not initalized then this call throws an NPE.

Hopefully the WebSocket EG will offer a better way in the future, but until such time the HttpSession needs to be initialized by a Filter like this one,or a Listener which will run for all URIs and is therefore less optimal.

### DIFF
--- a/java/org/apache/catalina/filters/SessionInitializerFilter.java
+++ b/java/org/apache/catalina/filters/SessionInitializerFilter.java
@@ -16,8 +16,8 @@
  */
 package org.apache.catalina.filters;
 
+import javax.servlet.Filter;
 import javax.servlet.FilterChain;
-import javax.servlet.GenericFilter;
 import javax.servlet.ServletException;
 import javax.servlet.ServletRequest;
 import javax.servlet.ServletResponse;
@@ -35,7 +35,7 @@ import java.io.IOException;
  * does not provide a way to do so.
  * </p>
  */
-public class SessionInitializerFilter extends GenericFilter {
+public class SessionInitializerFilter implements Filter {
 
     /**
      * Calls {@link HttpServletRequest}'s getSession() to initialize the HttpSession
--- a/java/org/apache/catalina/filters/SessionInitializerFilter.java
+++ b/java/org/apache/catalina/filters/SessionInitializerFilter.java
@@ -1,0 +1,62 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.catalina.filters;
+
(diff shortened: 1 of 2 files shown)


isapir:
Actually, I just noticed that `javax.servlet.Filter` added default no-op `init()` and `destroy()` (yay!), so this Filter can be modified to implement `javax.servlet.Filter` instead of extending `javax.servlet.GenericFilter` without further changes since it doesn't require FilterConfig or anything else.

isapir:
I updated accordingly to save a few CPU cycles and a few bytes of memory for each Request that hits the Filter.  

Cumulative change can be seen at https://github.com/apache/tomcat/pull/76/files?w=1

isapir:
p.s. I noticed that you sometimes backport code to 8.5 and even further back.  If you accept this patch, and if you want to backport, I can issue a different patch since this one will not work on 8.x without adding `init()` and `destroy()`.

Please let me know if you want me to issue a patch for the older versions.

markt-asf:
I did wonder if this would be better as configuration on WsFilter but on reflection a separate Filter looks to be simpler for users to configure.

The patch needs documentation (webapps/docs/config/filter.xml)

No need to patches for the back-ports. We'll use svn merge and add NO-OP methods as necessary.

isapir:
Sure, I will add the documentation.

Thanks :)

isapir:
@markt-asf I added the documentation as per your feedback

isapir:
@markt-asf I added the documentation that you requested

isapir:
@markt-asf Can you please pull this patch in?  It'd be great to have it available in Tomcat 9.  Thanks.

markt-asf:
Looking at this is on my TODO list to look at before the next release

ChristopherSchultz:
@markt-asf re: `Filter` versus configuration

This could be something that we add now just as a `Filter` but in the future add a configuration option that just automatically attaches the `Filter` -- just like how using `<auth-method>` in `WEB-INF/web.xml` auto-attaches the appropriate `Valve` to the valve chain.

kkolinko:
1. Create an entry in Bugzilla and reference this PR, so that this is correctly mentioned in change log.
http://tomcat.apache.org/bugreport.html

2. I do not mind this utility filter being added to Tomcat. It is small and there might be uses for it.
It is the same as the default of "<%@page session="true"%>" in JSPs.

But I think that its use for WebSocket needs some additional explanation / discussion, and maybe a different fix.

> "too late to initialize the HttpSession object"

The HttpSession is a tool that is used to share some information between requests. It is only useful if there is a subsequent request from the same client that is associated with the same session.

If the session is identified via a cookie (there are also different configuration options), it means that the client should receive a Set-Cookie header with a HTTP response and process it (in "101 Switching Protocols" HTTP response).

I have not tested whether the Set-Cookie header in a "101 Switching Protocols" response is recognized by actual clients.

> handshakeRequest.getHttpSession().getServletContext()

The above code is odd. You are not sharing some information, but are accessing some static configuration object. There should be an easier way to obtain a ServletContext. It is not what a session is needed for.

isapir:
@kkolinko I will create a ticket.

Unfortunately the issue with WebSocket is due to the JSR-356 specification.  I have spent a lot of time researching this issue and unfortunately that was the only solution that I found.

It's weird and it's a hack, but unless the EG updates the spec that's all we have.

See discussions at 
https://stackoverflow.com/questions/21888425/accessing-servletcontext-and-httpsession-in-onmessage-of-a-jsr-356-serverendpo
https://stackoverflow.com/questions/17936440/accessing-httpsession-from-httpservletrequest-in-a-web-socket-serverendpoint

PR #77, opened 2017-10-07
Removed findbugs bad practice warnings by making classes final

isapir:
CN org.apache.catalina.tribes.io.ChannelData.clone() does not call super.clone()
CN org.apache.catalina.tribes.membership.Membership.clone() does not call super.clone()
CN org.apache.catalina.tribes.tipis.AbstractReplicatedMap$MapMessage.clone() does not call super.clone()
CN org.apache.catalina.util.URLEncoder.clone() does not call super.clone()

### DIFF
--- a/java/org/apache/catalina/tribes/io/ChannelData.java
+++ b/java/org/apache/catalina/tribes/io/ChannelData.java
@@ -32,7 +32,7 @@ import org.apache.catalina.tribes.util.UUIDGenerator;
  * interceptors, the message data can be manipulated as each interceptor seems appropriate.
  * @author Peter Rossbach
  */
-public class ChannelData implements ChannelMessage {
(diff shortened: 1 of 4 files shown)


markt-asf:
Making those classes final will cause problems if users have extended any of them. While that seems unlikely, experience suggests it has probably been done somewhere.  Is there any reason super.clone() can't be used instead for these?

isapir:
I actually started with using super.clone() but upon further research opted for marking the classes as `final`.  All the "best practices" advocates say to avoid using the faulty clone() method as described in Joshua Bloch's book _Effective Java_, and also described in articles online like http://www.artima.com/intv/bloch13.html

I will refactor these to use `super.clone()`.

markt-asf:
These have been resolved since this PR was opened.

PR #79, opened 2017-10-19
remove placeholders from introduction doc

pmcelhaney:


markt-asf:
Those aren't placeholders. It is intentional content. I believe they are meant to be a lighthearted way of indicating that 'Context' is the only term that needs introducing.

csutherl:
This is a least the fourth time I've seen someone asking about this. Should we make a note or use a sarcasm font or something? :) It's funny, but seems to be confusing new folks.

markt-asf:
We could pull the entries and add "That's it!" after the list.

pmcelhaney:
IMHO the humor doesn’t work here. It comes across as condescending. It’s *definitely* not helpful.

ChristopherSchultz:
A few years ago, Rich Bowen gave a talk @ ApacheCon about documentation and communities. He used Python as an example of a community where the Monty Python jokes are so pervasive, it appears that they have become the focus of the documentation instead of the actual subject (the code, etc.).

He suggested that humor, while intended to be fun, lighthearted, etc. often comes across as mean-spirited or condescending, and/or introduces confusion especially to those who do not understand (a) the language (e.g. English), (b) that they are in fact jokes or (c) the content of the jokes (e.g. unfamiliar with Monty Python).

Basically, this kind of thing doesn't have a place in high-quality documentation.

This example is relatively benign, but there are probably other places where the humor is at least distracting if not worse.

ChristopherSchultz:
References:
https://apacheconna2015.sched.com/event/2P6d/rtfm-write-a-better-fm-rich-bowen-apache-software-foundation
http://feathercast.apache.org/podcasts/ApacheConNA2015/Monday/Community/Monday_Community_04%20-%20Rich%20Bowen%20-%20RTFM%20-%20Write%20a%20Better%20FM.mp3
http://drbacchus.com/rtfm-write-a-better-fm-recording/

PR #83, opened 2017-11-03
Bug 61668 - Possible NullPointerException in AbstractHttp11Protocol.java

mithunsasidharan:
Bug 61668 - Possible NullPointerException in org.apache.coyote.http11.AbstractHttp11Protocol

https://bz.apache.org/bugzilla/show_bug.cgi?id=61668


markt-asf:
The duplicated code is a sign that this could be handled elsewhere with less code (i.e. in StringUtils)

mithunsasidharan:
@markt-asf : Can I just call `getSSLProtocol()` from within `getSslEnabledProtocols()` ?

```
    public String getSslEnabledProtocols() {
        registerDefaultSSLHostConfig();
        getSSLProtocol()
    }
```

markt-asf:
Thanks for the proposed patch. I opted for a more general solution in StringUtils since fixing that problem there should - in theory - prevent it appearing anywhere else StringUtils is used.

PR #84, opened 2017-11-06
Add tomcat in the cloud abstract implementation

maxime-beck:
## Tomcat in the Cloud
_Tomcat in the Cloud_ is a project that seeks to add support for clustering into a cloud environment for a variety of cloud service providers such as OpenShift, Kubernetes, Azure and more.

This commit is the first of many related to this project. It adds the abstract implementation of both **_MemberProvider_** and **_StreamProvider_**.

### MemberProvider
Discovers all running instances of Tomcat in the cloud environment.

### StreamProvider
Establish connection throughout the cloud network to provide multi-casting in the cloud environment.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -374,6 +374,7 @@
     <include name="org/apache/tomcat/util/res/**" />
     <include name="org/apache/tomcat/util/security/**" />
     <include name="org/apache/tomcat/util/threads/**" />
+    <include name="org/apache/tomcat/util/providers/**" />
     <include name="org/apache/tomcat/util/*" />
     <exclude name="org/apache/tomcat/util/bcel" />
     <exclude name="org/apache/tomcat/util/descriptor" />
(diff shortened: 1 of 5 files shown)


markt-asf:
This needs a wider discussion on the dev list first.

However I do have some initial high level comments:
- The choice of package is unexpected
- Headers need to be standard ASF headers
- Author details need to be removed
- At least one file appears to have been written by someone other than the person submitting the pull request. That is a huge red flag as it is not clear that the person submitted the pull request has the necessary rights to do so.

jfclere:
See also in https://github.com/jgroups-extras/jgroups-kubernetes/blob/master/src/main/java/org/jgroups/protocols/kubernetes/stream/StreamProvider.java it seems we are reusing their internal interface, what doesn't look quite good, AKA it looks like an implementation detail of the Kubernetes logic.

maxime-beck:
Apparently Microsoft recently came up with a new container management service **[1]** based on Kubernetes for Azure. This means that all container management providers that we're aiming to support now potentially work with Kubernetes. It makes us rethink our data model and especially our necessity for abstraction.

Because this now needs a higher level of investigation and analysis, I'll close this pull request for now.
___
**[1]** https://azure.microsoft.com/en-us/blog/introducing-azure-container-service-aks-managed-kubernetes-and-azure-container-registry-geo-replication/

rmaucher:
Abstraction is ok, more importantly limiting the number of impls will make it much much easier to maintain.

PR #85, opened 2017-11-18
Fix NPE in AuthConfigFactoryImpl.detachListener()

lkirchev:
Test case and fix for https://bz.apache.org/bugzilla/show_bug.cgi?id=61777


markt-asf:
Note there is a minor typo in the test method name. I'll fix that when I apply the patch.

PR #87, opened 2017-11-18
Fix NPE on removeRegistration in case getConfigProvider is called wit…

lkirchev:
Test case and fix for https://bz.apache.org/bugzilla/show_bug.cgi?id=61779


markt-asf:
Applied in r1815802

PR #92, opened 2017-11-18
Support null for provider class name when registering a provider

lkirchev:
Test case and fix for https://bz.apache.org/bugzilla/show_bug.cgi?id=61784


lkirchev:
Will replace it with a new PR.

PR #95, opened 2017-11-30
Fix https://bz.apache.org/bugzilla/show_bug.cgi?id=61223

tluce:
Added the description of each mbean attribute to the documentation.


markt-asf:
This is a partial fix. It only addresses the attributes of the mbean element.

I'm wondering if, rather than duplicate the comments in the DTD file, if it would be simpler to copy the DTD to the docs and link to the DTD.

markt-asf:
Thanks for the patch. There wasn't anything wrong with it but the more I thought about it, the more I thought that referencing the DTD directly was a better solution.

PR #97, opened 2018-02-07
Bz48672

m-czernek:
Providing documentation patch for the https://bz.apache.org/bugzilla/show_bug.cgi?id=48672 bugzilla (and removing the TODO from the HTML interface).

Feel free to suggest any changes to the documentation, or further enhance it afterwards.


markt-asf:
Thanks looks great. I'll add it shortly. I'll apply to all versions and then tweak the various version specific parts so they reflect the current version.

markt-asf:
All updated.

PR #99, opened 2018-02-13
Don't gzip brotli compressed content

burka:
We try to serve brotli compressed static files while setting `Http11Protocol.COMPRESSION` to `on` (because we have some other files that benefit a lot from that.) 

That doesn't work because Tomcat compresses anything but gzip compressed content. 

This patch will allow brotli compressed content (`Content-Encoding: br`) to pass untouched :-)


burka:
Thanks @markt-asf :+1:

PR #101, opened 2018-02-27
Fix compressibleMimeType handling for compression="on"

px-stkn:
This change fixes a regression introduced by 913cb6d / r1816549
"Refactor: Move compression code to new class to allow re-use with HTTP/2",
which (among other things) consolidates isCompressible() and useCompression()
into a single functions.

Unfortunately, the call to getCompressibleMimeTypes(), got lost in the process,
leaving the compressibleMimeTypes array uninitialized, disabling the mimetype check
in useCompression(), effectively turning it into compression="force".

Signed-off-by: Stefan Knoblich <s.knoblich@praxino.de>

### DIFF
--- a/java/org/apache/coyote/CompressionConfig.java
+++ b/java/org/apache/coyote/CompressionConfig.java
@@ -203,7 +203,7 @@ public class CompressionConfig {
             }
 
             // Check for compatible MIME-TYPE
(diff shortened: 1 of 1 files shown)


markt-asf:
Thanks for the report and the patch. Fixed for trunk and 8.5.x.

PR #103, opened 2018-03-21
Use initSQL instead of Validator on connection initialization

jihor:
As of now, if Validator (not to be confused with validationQuery) is defined for the connection pool, initSQL will not be used on connection initialization, which I think is a bug, not a feature.

This commit changes this behavior so if initSQL is present, it will be used on connection initialization (and only then) instead of Validator.


ChristopherSchultz:
Honestly, I would expect `initSQL` to be executed regardless of any other configuration (e.g. Validator present or not).

jihor:
@gmcdonald I see that the PR is closed. Any plans to fix the issue? The following code block: https://github.com/apache/tomcat/blob/3e5ce3108e2684bc25013d9a84a7966a6dcd6e14/modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/PooledConnection.java#L489-L492

will lead to returning a connection without executing initSQL (this will happen on VALIDATE_INIT). As I understand, this means initSQL will never run on a connection if a validator is present.

PR #105, opened 2018-03-30
Iamtjw patch 1

IAMTJW:
set ervery war deployed timeout  when  a war  has endless loop and other war cannot deploy success，set timeout that some war can deploy success


markt-asf:
This needs some discussion on the dev@ list.

IAMTJW:
thanks，in my project has this problem and I dont know why the war cannt deployed , to solve this problem I download tomcat sources code and deploy that war and know the reason the other war cannt deloyed.

PR #107, opened 2018-04-24
underlying connection closed unexpectedly, so we must destroy this proxy

dltc:
if underlying connection closed,  we must destroy this proxy。


ChristopherSchultz:
Is this "just" a mild performance optimization? It looks like more code where no more code is needed.

dltc:
mysql-drive close the connection while IOException occur. if testOnReturn is false the connection will have been returned to the pool and may have  been borrowed by another thread

PR #108, opened 2018-04-24
Improve undeployment in parallel deployment scenario

koraktor:
Main reason for this pull request is to improve automatic undeployment of applications when using common version numbers, e.g. `application##1.2.3`.

Currently, `HostConfig#checkUndeploy()` uses plain sorting of the context names without any special handling of the version identifier. When having a version `1.0.9` deployed, deploying `1.0.10` will lead to immediate undeployment of `1.0.10` because the ordering is wrong.

This change introduces `ContextName#compareTo()` which knows how to handle commons version numbers.

### DIFF
--- a/java/org/apache/catalina/startup/HostConfig.java
+++ b/java/org/apache/catalina/startup/HostConfig.java
@@ -39,17 +39,14 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
-import java.util.SortedSet;
 import java.util.TreeSet;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Future;
-import java.util.function.Supplier;
 import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
-import java.util.stream.Collector;
 
 import javax.management.ObjectName;
 
@@ -1663,10 +1660,10 @@ public class HostConfig implements LifecycleListener {
         }
 
         // Need ordered set of names
-        SortedSet<ContextName> sortedAppNames = deployed.keySet().stream().
-                map(name -> new ContextName(name, false)).
-                collect(Collector.of((Supplier<SortedSet<ContextName>>) TreeSet::new, SortedSet::add,
-                        (left, right) -> { left.addAll(right); return left; }));
+        TreeSet<ContextName> sortedAppNames = new TreeSet<>();
+        for (String name : deployed.keySet()) {
+            sortedAppNames.add(new ContextName(name, false));
+        }
         Iterator<ContextName> iter = sortedAppNames.iterator();
 
         ContextName previous = iter.next();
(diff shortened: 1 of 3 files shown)


markt-asf:
I have applied the optimisation commit (thanks) but not the ordering changes.

The patch changes the order for undeploy without changing the order for version mapping. This will lead to unexpected behaviour.

The current simple `String` based ordering was selected as a trade off between performance and usability. The type of ordering proposed in this patch was considered too expensive to incur on every request when multiple versions are deployed. To consider such a change, we'd need to see evidence of the performance impact (including GC) on the mapping process of switching to this ordering mechanism. Be aware that mapping code is highly optimised for `String`.

Reviewing the proposed patch I see several things that would need to be fixed in additional to the more architectural issue described above:
 - imports from sun.* are not allowed
 - the code is Java 8 specific - any fix needs to be available to back-ported to 7.0.x will has to run on Java 6
 - the patch does not compile
 - `Pattern` instances should be pre-compiled

Finally, the patch only considers purely numeric versions. Many projects include a test string (e.g. RELEASE) in the version number. It would be nice to handle these as well.

koraktor:
@markt-asf @ChristopherSchultz 

Thanks for your feedback so far and for the partial merge. To be honest I didn’t think about any performance implications or side effects because I didn’t notice that this could be used in other places, especially during each request.

I’m sorry for the broken second patch and pushed an updated version which fixes the unused imports in `HostConfig` and broken syntax in `ContextName`.

I’ll have another look at the regex precompilation and Java 7 compatibility later. After that I want to look at improving performance and backwards-compatibility. Maybe someone can point me at the right code locations where this might be important?

markt-asf:
I think there is a lot of merit in Chris's idea. The concept of some sort of internal pre-computed version string that addresses ordering concerns by zero padding numerical components is one that would be easy to implement and have minimal impact on the existing system.

koraktor:
I added another bunch of patches. One of them implementing a different approach to a cached version representation based on @ChristopherSchultz’s suggestion. Instead of creating zero-padded strings I had the idea of using the version components as UTF-16 code points and building a string from those.

I did not yet test all edge cases. So this might not be fail-safe.
I’m still using regular expressions as the overhead of implementing a tokenizer seemed a bit much given the one-time nature of the new approach.

I did not include code for non-numeric version components (especially suffixes, e.g. `1.2.3-SNAPSHOT` or `1.2.3.pre5`.

koraktor:
I added support for version suffixes and rebased this on `trunk`. Maybe you want to have another look.

koraktor:
@markt-asf I’m sorry to nag you about this, but I didn’t get any feedback since my improvements on this patch series.

ChristopherSchultz:
I'm starting to lean toward requiring this new feature to require a configuration option to enable it, and have it default to `false`. My justification is that it represents a breaking behavioral change that is *just slightly different* than previous behavior as to be unnoticeable until it starts behaving completely unexpectedly.

Can you please add a configuration option for `<Host>` to enable or disable this? Simply swap-out the `compareTo` implementation depending upon the value of that setting.

markt-asf:
I agree this should configurable on the Host.
It should be possible to remove the Comparator entirely if the version code is used directly in the Mapper.
I'd like this to be more robust. I'm thinking pad any dot-separated segment to a given length with a suitable character (possible space).

PR #110, opened 2018-05-10
Debug log for pool's db properties

wildloop:
I found that all non-string values are ignored by [java.util.Properties.java](https://github.com/JetBrains/jdk8u_jdk/blob/master/src/share/classes/java/util/Properties.java#L970) and sometime we don't know what exactly properties are passed to the driver. 
The way to log these properties out would be helpful...

For example yaml:
- ` db-properties.oracle.net.CONNECT_TIMEOUT: "30000"` - works
- ` db-properties.oracle.net.CONNECT_TIMEOUT: 30000` - doesn't work :(

### DIFF
--- a/modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/PooledConnection.java
+++ b/modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/PooledConnection.java
@@ -20,6 +20,7 @@ package org.apache.tomcat.jdbc.pool;
 import java.sql.DriverManager;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.util.Enumeration;
 import java.util.HashMap;
 import java.util.Properties;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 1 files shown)


ChristopherSchultz:
What problem does this solve? What property file do you have that contains non-strings?

wildloop:
It's just a debug-log to help find problems with incorrect config... in mys case it was: `db-properties.oracle.net.CONNECT_TIMEOUT: 30000`

ChristopherSchultz:
So you were providing a `Properties` object that was not loaded from a file, and had non-string values in it? Your patch would not identify that problem. Non-string values should probably be allowed within tomcat-pool. Care to do a PR for that as well?

wildloop:
Yes, cloud foundry connector initializes datasource basing on enviroment variables.
My patch shows all properties, so it shows that something is missing.

>Non-string values should probably be allowed within tomcat-pool. 

I think so, problem is here (`(oval instanceof String) ? (String)oval : null`): https://github.com/JetBrains/jdk8u_jdk/blob/master/src/share/classes/java/util/Properties.java#L970

PR #112, opened 2018-05-20
Fix BZ Bug 62391

isapir:
This PR fixes bug https://bz.apache.org/bugzilla/show_bug.cgi?id=62391

It is a very simple fix so might be easier to apply directly rather than accepting this patch.

If approved, I can probably apply it directly myself in SVN, which will give me some practice with that.


isapir:
Closing on Github since I can now patch SVN directly if there is no objection to this change

PR #113, opened 2018-05-20
Modify Windows startup command from start to run

isapir:
Applies enhancement as proposed in BZ-62392 - https://bz.apache.org/bugzilla/show_bug.cgi?id=62392

It is a very simple fix so might be easier to apply directly rather than accepting this patch.

If approved, I can probably apply it directly myself in SVN, which will give me some practice with that.


isapir:
Closing per comments in the BZ ticket

PR #117, opened 2018-08-02
Enhance the CATALINA_BASE documentation

m-czernek:
As per the mailing list discussion, we'd like to enhance the catalina.home and catalina.base documentation so that it is present in the main html documentation, and not only in RUNNING.txt and various script files.

Feel free to request corrections or further changes. What do you think?

### DIFF
--- a/webapps/docs/introduction.xml
+++ b/webapps/docs/introduction.xml
@@ -63,15 +63,6 @@ do let us know.</p>
 
 <section name="Directories and Files">
 
-<p>Throughout the docs, you'll notice there are numerous references to
-<strong>$CATALINA_HOME</strong>. This represents the root of your Tomcat
-installation. When we say, "This information can be found in your
-$CATALINA_HOME/README.txt file" we mean to look at the README.txt file at the
-root of your Tomcat install. Optionally, Tomcat may be configured for multiple
-instances by defining <strong>$CATALINA_BASE</strong> for each instance. If
-multiple instances are not configured, <strong>$CATALINA_BASE</strong> is the
-same as <strong>$CATALINA_HOME</strong>.</p>
-
 <p>These are some of the key tomcat directories:</p>
 
 <ul>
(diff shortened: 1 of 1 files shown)


webapps/docs/introduction.xml
```diff
@@ -89,6 +80,122 @@ same as <strong>$CATALINA_HOME</strong>.</p>
...
+      <ul>
+        <li>conf/server.xml</li>
+        <li>conf/web.xml</li>
+      </ul>
+      That includes the <code>conf</code> directory. Otherwise, Tomcat may
+      fail to start.
+    </p>
+    <p>
+      Additionally, it may also contain the following:
+      <ul>
```
markt-asf: I think some slightly stronger wording is required here. Not having some of these directories is going to cause problems.

webapps/docs/introduction.xml
```diff
@@ -89,6 +80,122 @@ same as <strong>$CATALINA_HOME</strong>.</p>
...
+      However, in case you require your own logging implementation, you can 
+      replace the <code>tomcat-juli.jar</code> file in a CATALINA_BASE location
+      for the specific Tomcat instance.
+    </p>
+    <p>
+      We also recommend you not to copy over any configuration file that is identical
+      for multiple Tomcat instances, with the exception of the mandatory files.
+      For example, copy over the <code>conf/logging.properties</code> only if 
+      you require the particular Tomcat instance to have specific logging settings.
+    </p>
```
markt-asf: That won't work. Tomcat looks in CATALINA_BASE for the logging configuration. You could change that in catalina.[sh|bat] but that starts to make the setup more complicated.

Generally, ALL configuration files are only read form BASE and DO NOT fall back to HOME. The only exception in bin/setenv.sh and this onyl falls back to HOME for historical backwards compatibility reasons. I'd suggest updating this to say copy all of conf/* across.

webapps/docs/introduction.xml
m-czernek: Could you please be a bit more specific? It's interesting because before putting anything in the docs, I always try it out (just to turn it on and see if there are any errors). Strictly speaking, conf/server.xml is the only file necessary for no errors during startup. Which other dirs/files would you recommend marking as required?

webapps/docs/introduction.xml
m-czernek: Incorporated.

webapps/docs/introduction.xml
markt-asf: Strictly speaking, I agree with you. My concern is that a novice user will read this and think they only need those files whereas most users will want at least logs, temp, webapps, work, most of conf/* and probably lib too.

Just considering the work directory - if that is missing and Tomcat can't create one lots of stuff breaks, the most obvious of which is JSP compilation.

What I think is required here is a change of emphasis. Maybe start with what is typically required (most / all of it) and then discuss what the bare minimum looks like.

I'd also add notes to each dir about the relationship between HOME and BASE. ie for bin/setenv.sh Tomcat looks in BASE then HOME. For lib, Tomcat looks in BASE then HOME (configured in conf/catalina.propertis), for all other dirs Tomcat only looks in BASE (unless configured to use an absolute path that points elsewhere which could include HOME).

m-czernek: Thank you for your thorough review and my apologies for the delay in implementing it. I think what you said does make sense. I have implemented your feedback, just please  check whether what I wrote is indeed correct. 

I provided the "order of lookup" note for directories where it made sense. Not sure there is any lookup going on in logs, for example (Imho that's rather scope than lookup, while for lib, lookup makes sense to me).

What do you think?

markt-asf:
Looks great. Applied. Many thanks.
(note: there was one s/HOME/BASE/ required in the webapps section)

PR #120, opened 2018-09-04
fix -Djava.security.policy setting

EstherLacan:
fix -Djava.security.policy setting


rainerjung:
I think the double "==" here is itentional. For the meaning of the double equals sign see e.g.

https://docs.oracle.com/javase/7/docs/technotes/guides/security/PolicyFiles.html

(look for ==). Keeping it open for others to chime in.

markt-asf:
The double = is deliberate. The OP needs to RTFM.

PR #121, opened 2018-09-12
restart.bat

akag98:
Bat file that calls shutdown.bat and then startup.bat to restart the server.

### DIFF
--- a/bin/restart.bat
+++ b/bin/restart.bat
@@ -1,0 +1,2 @@
+call shutdown.bat
+call startup.bat


kkolinko:
1. This does not conform to code style. The "echo off" command and license header are missing.
2. This approach is naive and does not work. The "shutdown.bat" script does not stop Tomcat! It just commands Tomcat to stop. The shutting down may take some time.
3. A rather rare use case.
If one is developing, I think an IDE should know how restart Tomcat (e.g. Eclipse IDE knows how to do it and how to stop Tomcat process forcefully if it does not shut down in time). 

Thus I am -1 on merging this.

4. IMHO, the proper place to file enhancement requests is Bugzilla, so that they can be referenced in changelog.

akag98:
Thanks @kkolinko for your review...

PR #122, opened 2018-09-16
Added a default value for ApplicationSessionCookieConfig#name

bitstorm:
By spec. javax.servlet.SessionCookieConfig#getName should return 'JSESSIONID' as default value (see https://docs.oracle.com/javaee/7/api/javax/servlet/SessionCookieConfig.html). 
However, ApplicationSessionCookieConfig has no default value for this field and returns 'null' if not set.
This PR set  'JSESSIONID'  as default value.

### DIFF
--- a/java/org/apache/catalina/core/ApplicationSessionCookieConfig.java
+++ b/java/org/apache/catalina/core/ApplicationSessionCookieConfig.java
@@ -38,7 +38,7 @@ public class ApplicationSessionCookieConfig implements SessionCookieConfig {
     private int maxAge = -1;
     private String comment;
     private String domain;
-    private String name;
+    private String name = "JSESSIONID";
     private String path;
     private StandardContext context;
 


kkolinko:
The javadoc that you linked
https://docs.oracle.com/javaee/7/api/javax/servlet/SessionCookieConfig.html
says "Returns: [...] or null if setName(java.lang.String) was never called"

Thus it documents null as the default value.

BTW, the cookie name can be overwritten by sessionCookieName attribute on Context,
http://tomcat.apache.org/tomcat-9.0-doc/config/context.html#Common_Attributes

markt-asf:
Closing the PR as the requested change is not correct.

bitstorm:
@kkolinko 

Thank you for the clarification!

martin-g:
The new question is "Should Tomcat call `#setName("JSESSIONID")` during start of an application (just after creation of `SessionCookieConfig`) ?" With the current behavior every application/framework should make checks for `null` and assume `JSESSIONID`.

markt-asf:
There is nothing in the spec that says Tomcat is required to do that.

martin-g:
There is also nothing that says not to do it. The question is which behavior is better.
IMO returning the actual value that is in use is much better than `null`.

solomax:
+1 no-one like NPE :)

martin-g:
s/edge/age/

solomax:
@markt-asf ping :)

PR #124, opened 2018-10-03
Allow empty 'Host' HTTP header

michaelomichael:
Sending an empty string for the value of the 'Host' header no longer
results in a 400 (Bad Request) response.  This new behaviour reflects
RFC 7230 and is particularly helpful for load balancer ping-style
requests, where there is often limited control over the HTTP request
headers.

https://bz.apache.org/bugzilla/show_bug.cgi?id=62739


markt-asf:
Nice catch with handling the port. The patch I had come up with missed that.
I refactored populateHost() as it bugged me to set the server name one way (that might throw an IOE and break the request) only to immediately overwrite the result with an empty String.

PR #130, opened 2018-11-01
Add two spaces (nbsp) in front of the Undeploy button

ljader:
Since there is no JavaScript in the Manager HTML,
I decided to visually separate the Undeploy button from the rest of the buttons.

This is to "prevent" **missclicks** on the Reload button,
which ends with the Undeploy being clicked.

(there is no Bugzilla ticket for this - I've created this to get your opinion on this.)


markt-asf:
I can see what you are trying to do but I'm not sure the change creates enough of a visual separation to make much of a difference. I don't think it will do much to reduce the risk of an accidental click.
On the other hand, it would have taken me less time to just apply the change than I have spent writing this comment.
I'm going to apply the change as it won't make the issue worse, it might help and it is a trivial change.

PR #132, opened 2018-11-11
Fix cygpath error by adding cond check with JAVA_ENDORSED_DIRS

zemian:
The new `bin/tool-wrapper.sh` script gives the error `cygpath: can't convert empty path` when used in Cygwin shell. This is due to `JAVA_ENDORSED_DIRS` was converted even when it's empty.

This PR fixes the error by checking if `JAVA_ENDORSED_DIRS` is not empty first.


markt-asf:
Fixed in:
- trunk for 9.0.14 onwards
- 8.5.x for 8.5.36 onwards
- 7.0.x for 7.0.93 onwards

Thanks for the report and for the patch.

PR #137, opened 2019-01-18
Add HTML lang="en"

skotfred:
This is primarily an accessibility issue for screen reader software.

https://webaim.org/techniques/screenreader/

### DIFF
--- a/java/javax/servlet/http/package.html
+++ b/java/javax/servlet/http/package.html
@@ -1,5 +1,5 @@
 <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
-<HTML>
+<HTML lang="en">
 <HEAD>
 <!--
  Licensed to the Apache Software Foundation (ASF) under one or more
--- a/java/javax/servlet/http/package.html
+++ b/java/javax/servlet/http/package.html
@@ -1,5 +1,5 @@
 <!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 3.2 Final//EN">
-<HTML lang="en">
+<HTML>
 <HEAD>
 <!--
  Licensed to the Apache Software Foundation (ASF) under one or more
--- a/java/javax/servlet/jsp/el/package.html
+++ b/java/javax/servlet/jsp/el/package.html
@@ -1,5 +1,5 @@
 <!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML//EN">
-<html lang="en">
+<html>
 <head>
 <!--
  Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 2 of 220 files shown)


markt-asf:
I'm all in favour of improving accessibility.
The changes to package.html and similar look good.
The changes to anything under test/ need to be reverted. Those are test artefacts.
The changes to to anything under webapps/ needs to be considered in the context of the work to improve localisation where there is at least one open enhancement to add the necessary framework to provide that content in languages with the correct language being selected based on the user's locale. It may be that the changes proposed here are OK. It just needs some thought.

kkolinko:
> The changes to package.html and similar look good.

The doctype declaration in package.html says HTML 3.2 and I that version does not have the lang attribute on HTML element.  HTML 4.01 DTD has the lang attribute, so files using that specification are OK.

Anyway, those package.html files are just fragments consumed by javadoc tool. I suspect that changing the attributes of the HTML element does not have any effect on generated Javadoc documentation, but it needs testing.

> The changes to to anything under webapps/ needs to be considered in the context of the work to improve localisation

I agree.
404/403/401.jsp pages in manager and host-manager may benefit from localization, and in that case the lang attribute has to be calculated dynamically to reflect the actual language being used. Though currently those pages are hard-coded with some English text.

markt-asf:
> Anyway, those package.html files are just fragments consumed by javadoc tool. I suspect that changing the attributes of the HTML element does not have any effect on generated Javadoc documentation, but it needs testing.

I took a quick look. You are correct. The package.html files become package-summary.html files and they all use HTML 4.0.1 and have lang="en"

On that basis package.html changes are unnecessary and should also be removed from this PR.

I also looked at the other HTML files and they don't seem to serve any purpose. I'll look into them some more but I'll probably end up removing them.

ChristopherSchultz:
> 404/403/401.jsp pages in manager and host-manager may benefit from localization, and in that case the lang attribute has to be calculated dynamically to reflect the actual language being used. Though currently those pages are hard-coded with some English text.

When introducing i18n for things like this, it's important to know which language was actually chosen when requesting, for example, some unsupported language. The `StringManager` only uses one locale, but it never checks to see which locale it actually got. Client code can't request a string from a particular locale... it can only ask the StringManager which locale it's actually serving.

Java's `ResourceBundle` is actually insufficient for implementing real i18n unless every single string is always translated in each resource bundle into each language for which a bundle file exists. When you have non-100% coverage, it's not possible to use just `ResourceBundle` to do the job.

Instead, you need a hierarchical system based upon `ResourceBundle` and a few other things. It gets messy.

If we just want to use whatever `StringManager` loads on startup, or instantiate a new `StringManager` for each requested-locale and hope that we have 100% string-coverage, we can use that and just put `<html lang="<%= sm.getLocale().getLanguage() %>"` into each file.

If we want to provide a better example of implementing i18n in Tomcat, we'll need something more elaborate.

I'm +1 on adding `lang="en"` for now since the files are hard-coded in English.

markt-asf:
Can you re-submit this against master please. We migrated from svn to git as the primary repo and did a little re-organisation.

PR #139, opened 2019-02-19
service.bat handles changed Service at installation

bkiselka:
Enable service.bat to deal with changed "Windows Service Name" defined during Tomcat Windows Installation.

During the Apache Tomcat Setup You can define on the "Configuration Options" page a "Windows Service Name". If you change the default "Tomcat8" this results not only in a changed installation folder and a changed Windows Service Name, but also to renamed files bin/Tomcat8.exe and bin/Tomcat8w.exe.

The proposed changes deal with renamed Windows service wrapper executables by using the provided parameter name for the service name. Note that the parameter is no longer optional then!

Another solution would be to change the Windows installer and do not rename the bin/Tomcat@VERSION_MAJOR@.exe and                     bin/Tomcat@VERSION_MAJOR@w.exe during installation.

### DIFF
--- a/bin/service.bat
+++ b/bin/service.bat
@@ -24,22 +24,50 @@ rem remove                 Remove the service from the System.
 rem
 rem name        (optional) If the second argument is present it is considered
 rem                        to be new service name
+rem                        Must be defined, if bin/Tomcat@VERSION_MAJOR@.exe and 
+rem                        bin/Tomcat@VERSION_MAJOR@w.exe are renamed (e.g. at installation)!
 rem ---------------------------------------------------------------------------
 
 setlocal
 
 set "SELF=%~dp0%service.bat"
+
+rem Set default Service name
+set SERVICE_NAME=Tomcat@VERSION_MAJOR@
+set DISPLAYNAME=Apache Tomcat @VERSION_MAJOR_MINOR@ %SERVICE_NAME%
+
+if "x%1x" == "xx" goto displayUsage
+set SERVICE_CMD=%1
+shift
+if "x%1x" == "xx" goto checkServiceCmd
+:checkUser
+if "x%1x" == "x/userx" goto runAsUser
+if "x%1x" == "x--userx" goto runAsUser
+set SERVICE_NAME=%1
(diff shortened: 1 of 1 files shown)


bkiselka:
Could You please state, why you closed the Pull Request?

The Tomcat Windows Installer renames the files bin/Tomcat8.exe and bin/Tomcat8w.exe, which cannot be found by the bin/service.bat afterwards.
To me this is a bug!

Both the installer https://github.com/apache/tomcat/commits/master/res/tomcat.nsi and the https://github.com/apache/tomcat/blob/master/bin/service.bat did not change.

Or shall I open a bug on https://bz.apache.org/bugzilla first?

markt-asf:
Re-open PR that was closed as a side effect of the svn -> git migration.

markt-asf:
For reference, this was the trigger for the filename to change:
https://bz.apache.org/bugzilla/show_bug.cgi?id=50949

The advantage of changing the file name is that Commons Daemon derives the current service name from it so it doesn't have to be explicitly specified.

markt-asf:
The patch needs some more work. Currently it works if you install a different service name using the installer (because `tomcat9[w].exe` are renamed) but it fails if you install from the zip and try `service.bat install foo`
I suggest you update the patch to first look for `tomcat9[w].exe` and then, if that is not found, `servicename[w].exe`

markt-asf:
I started to implement the above suggestion and hit another issue. If you install via the installer with a non-standard name you can;t rename the service because the .exe files have been renamed.

Having looked at various options I think the best option is as follows:
- leave the installer as is
- require the service name to be provided unless the default name is used
- rename .exe files when installing non-default service name from batch file (to align with installer behaviour)
- rename .exe files to defaults when uninstalling non-default service name from batch file (to allow subsequent installations to use any valid service name)

These last two points are the main change. Unless there are objections, I'll proceed on this basis later this week.

bkiselka:
Sorry that I had no time to tackle your proposed changes from 9 Apr for the installation from zip file via batch file.
All your proposed changes from 30 July sound reasonable and clear to me (no objections!) and are necessary to solve the issue.

markt-asf:
Closing at this has been implemented but with a modified approach that handled the renaming of the executables.

PR #140, opened 2019-02-27
jdbc-pool: Improve maxAge handling

toby1984:
### Motivation

At our company we're dealing with very large PostgreSQL clusters (100TB on-disk size,up to 300k tables per instance) and have recently been bitten by PostgreSQL backends holding onto non-trivial (>1GB per backend) amounts of memory that do not get released after the end of a transaction.

While trying to use the pool's "maxAge" property to automatically recreate connections after 1 hours we noticed the following:

#### 1.) No age check on idle connections

The current implementation only performs age checks when connections are borrowed or returned.
Since the ConnectionPool implementation does not guarantee that idle connections are handed out in a round-robin fashion, connections may sit in the idle queue for a long time and thus exceed maxAge by an arbitrary amount, holding onto memory on the server side. This is especially an issue for large connection pools that briefly saw high utilization and then only very little use for a long time afterwards while also having idle connections cleanup disabled.

#### 2.) Connections are not actually getting recreated right away but discarded and only recreated on demand later on

This was surprising behaviour to us as we expected the maxAge setting to be purely a work-around for memory retention issues on the server side but not changing the number of established connections in the pool. 
While this will usually not be a problem, in our case it matters because we unfortunately have to compete for database connections with other processes in our product (yeah, bad design) so we're forced to hold onto the connections that we managed to obtain as much has possible.

This pull request contains the following changes:

1. Updated Javadoc / documentation
2. ConnectionPool.java
  - new method reconnectIfExpired() that checks a connection's max age and will
    try to reconnect if maxAge is exceeded
  - initializePoolCleaner() now makes sure the pool cleaner execution interval
    is no larger than maxAge+10%
  - shouldClose() no longer checks maxAge
  - returnConnection() now invokes new method ConnectionPool#reconnectIfExpired(PooledConnection)
  - testAllIdle() got a 'checkMaxAgeOnly' parameter in case the pool cleaner
    thread is only running because maxAge > 0
  - PoolCleaner#run() will now invoke testAllIdle(boolean) when
    testWhenIdle() == false and maxAge > 0
3. Added a unit test to check that maxAge now also gets enforced for idle connections and the connection is re-created immediately
4. Added a unit test to check that returning a connection after it exceeded maxAge will immediately re-create it

### DIFF
--- a/modules/jdbc-pool/doc/jdbc-pool.xml
+++ b/modules/jdbc-pool/doc/jdbc-pool.xml
@@ -335,7 +335,7 @@
     <attribute name="timeBetweenEvictionRunsMillis" required="false">
       <p>(int) The number of milliseconds to sleep between runs of the idle connection validation/cleaner thread.
          This value should not be set under 1 second. It dictates how often we check for idle, abandoned connections, and how often
-         we validate idle connections.
+	 we validate idle connections. This value will be overridden by <code>maxAge</code> if the latter is non-zero and lower.
          The default value is <code>5000</code> (5 seconds). <br/>
       </p>
     </attribute>
@@ -463,17 +463,22 @@
     </attribute>
 
     <attribute name="maxAge" required="false">
-      <p>(long) Time in milliseconds to keep this connection. This attribute
-         works both when returning connection and when borrowing connection.
+      <p>(long) Time in milliseconds to keep a connection before recreating it.
          When a connection is borrowed from the pool, the pool will check to see
          if the <code>now - time-when-connected > maxAge</code> has been reached
          , and if so, it reconnects before borrow it. When a connection is
          returned to the pool, the pool will check to see if the
          <code>now - time-when-connected > maxAge</code> has been reached, and
-         if so, it closes the connection rather than returning it to the pool.
+         if so, it tries to reconnect.
+         When a connection is idle and <code>timeBetweenEvictionRunsMillis</code> is
+         greater than zero, the pool will periodically check to see if the
+	     <code>now - time-when-connected > maxAge</code> has been reached, and
+	     if so, it tries to reconnect.
+	     Setting <code>maxAge</code> to a value lower than <code>timeBetweenEvictionRunsMillis</code>
+	     will override it (so idle connection validation/cleaning will run more frequently).
          The default value is <code>0</code>, which implies that connections
          will be left open and no age check will be done upon borrowing from the
-         pool and returning the connection to the pool.</p>
+         pool, returning the connection to the pool or when checking idle connections.</p>
     </attribute>
 
     <attribute name="useEquals" required="false">
--- a/modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/ConnectionPool.java
+++ b/modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/ConnectionPool.java
@@ -524,6 +524,12 @@ public class ConnectionPool {
             log.warn("maxIdle is smaller than minIdle, setting maxIdle to: "+properties.getMinIdle());
             properties.setMaxIdle(properties.getMinIdle());
         }
+        if (properties.getMaxAge()>0 && properties.isPoolSweeperEnabled() &&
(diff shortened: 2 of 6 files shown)


rmaucher:
1 seems ok. 2 is probably not since there are attributes like "minIdle" and "initialSize".

toby1984:
But a connection expiring because it is idle is not the same as it expiring because it reached a certain age. The one configuration considers the time since the connection last got used while the other considers the time since creation of the connection. I don't find mixing these two concepts intuitive.

toby1984:
Could you please elaborate what you meant by your reference to the "maxIdle" and "initialSize" options ? How do they relate to the "maxAge" one in your opinion ?

rmaucher:
I think maxAge is ok, but then the pool should indeed have at least minIdle. Basically, if you have 0 active, then you have at least minIdle, that sounds intuitive to me. I'm not doing the pool btw, just commenting based off the thread pool configurations.

toby1984:
Sorry for not getting back any earlier. I did not touch the existing code responsible for discarding idle connections or the creation of new connections, so if your statement "if you have 0 active, then you have at least minIdle" used to be true in the past, it still holds now. 
The only change to the previous behaviour that *might* be perceived by a user is that busy connections returned to the pool that also exceeded maxAge are not getting discarded anymore but instead only the underlying JDBC connection gets recreated. This is only visible to a user if the pool is not being used again right away, otherwise a new connection would've been created and the change would be imperceivable.

KeiichiFujino:
If testOnConnect is true or initSQL is set, we need validate the connection( with VALIDATE_INIT) when reconnecting.

toby1984:
Good catch, I will fix this.

toby1984:
I fixed the missing calls to PooledConnection#validate(int). 

While reading the code I was kind of unsure whether I should call PooledConnection#setTimestamp(long) when reconnecting an idle connection. Looking at the code it seems that the timestamp should only be updated when user-initiated (like borrow or return) operations are executed on the PooledConnection, not when (possibly internal) operations are executed on the underlying JDBC Connection - right ? The JavaDoc on setTimestamp() is kind of fuzzy and does not explain what kind of 'operations' are meant.

KeiichiFujino:
> I fixed the missing calls to PooledConnection#validate(int).

Thanks.



> Looking at the code it seems that the timestamp should only be updated when user-initiated (like borrow or return) operations are executed on the PooledConnection, not when (possibly internal) operations are executed on the underlying JDBC Connection - right ?

That's right.
There is no need to call PooledConnection#setTimestamp(long) when reconnecting an idle connection.

KeiichiFujino:
Personally, the primary factor in starting PoolCleaner should be timeBetweenEvictionRunsMillis.
Same as removeAbandoned / testWhileIdle / minEvictableIdleTimeMillis.
If we want to reconnect to an idle connection,
We should set both maxAge and timeBetweenEvictionRunsMillis.


(JMX)ConnectionPool provides the feature to restart PoolCleaner when attributes related to PoolCleaner are changed.
If maxAge is changed, you should restart PoolCleaner.

see:org.apache.tomcat.jdbc.pool.jmx.ConnectionPool

toby1984:
Thanks for taking the time to look at my changes! I totally missed the code dealing with dynamic reconfiguration through JMX.

I changed my code to keep the old behaviour (pool cleaner is only started when  timeBetweenEvictionRunsMillis > 0) and added documentation to maxAge pointing out that the age of idle connections will only be checked if both maxAge and timeBetweenEvictionRunsMillis are set.
I also adjusted ConnectionPool#checkPoolConfiguration() to set timeBetweenEvictionRunsMillis=maxAge in case that timeBetweenEvictionRunsMillis>maxAge (because otherwise the maxAge cannot be guaranteed). setMaxAge() now calls both checkPoolConfiguration() and a new poolCleanerAttributeUpdated() method where I moved all the duplicated "start/stop pool cleaner if necessary"  code to.

KeiichiFujino:
Thanks for the fix.
I have no objections to the current code.
I'll merge this into master branch.

PR #142, opened 2019-03-06
Defer creation of encodingToCharsetCache

philwebb:
Defer creation of `encodingToCharsetCache` until absolutely necessary
and update `B2CConverter.getCharset` to first attempt a direct lookup
using `Charset.forName`.

This change helps improve startup time and reduces the number of
classes loaded when `B2CConverter` is only called with standard
charsets.

See https://bz.apache.org/bugzilla/show_bug.cgi?id=63235

### DIFF
--- a/java/org/apache/tomcat/util/buf/B2CConverter.java
+++ b/java/org/apache/tomcat/util/buf/B2CConverter.java
@@ -25,9 +25,12 @@ import java.nio.charset.CharsetDecoder;
 import java.nio.charset.CoderResult;
 import java.nio.charset.CodingErrorAction;
 import java.nio.charset.StandardCharsets;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import org.apache.tomcat.util.res.StringManager;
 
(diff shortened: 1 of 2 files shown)


rmaucher:
If there is a lookup with Charset.forName, then would the cache be used at all ? I don't understand this sort of Windows style "startup time" optimization, where you hide stuff behind making running actual application code slower. This is often a bad design, so I don't see what this change brings.
-0

philwebb:
I've pushed an update following the discussion on https://bz.apache.org/bugzilla/show_bug.cgi?id=63235

@rmaucher 

> If there is a lookup with Charset.forName, then would the cache be used at all

Correct. The updated code relies on the [internal cache](https://github.com/openjdk-mirror/jdk7u-jdk/blob/f4d80957e89a19a29bb9f9807d2a28351ed7f7df/src/share/classes/java/nio/charset/Charset.java#L465-L498) that the `Charset` class uses.

> I don't understand this sort of Windows style "startup time" optimization, where you hide stuff behind making running actual application code slower

I don't know what you mean by "Windows style startup time optimization". Why would the actual actual application code now be slower?

philwebb:
I've pushed another update now that I know about https://bz.apache.org/bugzilla/show_bug.cgi?id=51400

rmaucher:
Besides the 31ms startup time, it's still worse in every way compared to the current code.

philwebb:
> Besides the 31ms startup time, it's still worse in every way compared to the current code.

I'm really not sure what I'm supposed to do with a comment like that. As someone who's just trying to help contribute a fix for an identified problem, it really doesn't make me feel very welcome :(

I'm happy to have another go at fixing this if you can provide something more constructive than "it's worse in every way". Having the current code call `availableCharsets()` which is specifically documented with "may cause time-consuming disk or network I/O operations to occur" in a static initializer with no way to opt-out doesn't seem ideal either. AFAICT this is something that neither Jetty or Undertow do, so I don't believe it's a problem that _can't_ be solved in another way.

I understand that Tomcat has many users that start it only once, and aren't really bothered by these kind of optimizations. To them 30ms and not loading an extra 50-60 classes isn't really a bit deal. However, I strongly believe that there is another class of user that want to run Embedded Tomcat and will care deeply about this kind of thing.

If the type of change I'm suggesting is never going to be acceptable, perhaps I can change my approach. Would it be better if I tried to make the charset lookup code plugabble? That way, the existing code would not need to change, but users who do care about this issue could replace the lookup code with something else.

rmaucher:
You submitted three items, and indeed I am not convinced by this one. Looking back to the old BZs, there was also there the conclusion that the current code is the best solution for the general Tomcat use case.

However, there is the "but I only want UTF-8" argument, in which case the solution is to hardcode UTF-8 and default to Charset.forName for the rest. I think system property configuration is supposed to be removed, but it would be a good solution for this pluggability (it's global, no one really actually needs it, etc).

markt-asf:
Nooooo. Not an system property ;).

Seriously, I think there is a way that we can improve start-up time without impacting the vast majority of users. Phil's patch is heading in the general direction I was thinking. As ever, there are likely to be trade-offs involved. I hope to be able to have some hard numbers on which to base that discussion soon(ish).

markt-asf:
See #146 for an alternative approach

philwebb:
Closing in favor of #146. Thanks @markt-asf!

PR #143, opened 2019-03-06
Apply deduplication to certain loaded and created Strings

philwebb:
Apply deduplication to certain loaded and created Strings by using
a static Map. This applies the same technique as described in
https://shipilev.net/jvm/anatomy-quarks/10-string-intern/ to
reduce the number of String instances that ultimately contain
the same contents.

See Fix https://bz.apache.org/bugzilla/show_bug.cgi?id=63236

### DIFF
--- a/java/org/apache/tomcat/util/digester/SetPropertiesRule.java
+++ b/java/org/apache/tomcat/util/digester/SetPropertiesRule.java
@@ -18,16 +18,13 @@
 
 package org.apache.tomcat.util.digester;
 
-
 import org.apache.tomcat.util.IntrospectionUtils;
 import org.xml.sax.Attributes;
-
 
 /**
  * <p>Rule implementation that sets properties on the object at the top of the
  * stack, based on attributes with corresponding names.</p>
  */
-
 public class SetPropertiesRule extends Rule {
 
     /**
@@ -62,7 +59,7 @@ public class SetPropertiesRule extends Rule {
             if ("".equals(name)) {
                 name = attributes.getQName(i);
             }
-            String value = attributes.getValue(i);
+            String value = attributes.getValue(i).intern();
 
             if (digester.log.isDebugEnabled()) {
                 digester.log.debug("[SetPropertiesRule]{" + digester.match +
--- a/java/org/apache/tomcat/util/modeler/modules/MbeansDescriptorsIntrospectionSource.java
+++ b/java/org/apache/tomcat/util/modeler/modules/MbeansDescriptorsIntrospectionSource.java
@@ -348,8 +348,8 @@ public class MbeansDescriptorsIntrospectionSource extends ModelerSource
                 for(int i=0; i<parms.length; i++ ) {
                     ParameterInfo pi=new ParameterInfo();
                     pi.setType(parms[i].getName());
-                    pi.setName( "param" + i);
-                    pi.setDescription("Introspected parameter param" + i);
+                    pi.setName(("param" + i).intern());
+                    pi.setDescription(("Introspected parameter param" + i).intern());
                     op.addParameter(pi);
                 }
                 mbean.addOperation(op);


rmaucher:
Can you give a metric of actual memory gains made using this change ? Adding complexity for very little benefit doesn't seem like a good move overall.

philwebb:
Sure. I discovered these by using the "Inspections" feature from [YourKit](https://www.yourkit.com/) on an [embedded Tomcat application](https://github.com/spring-projects/spring-boot/tree/87d696d6971c8bbd28be4fa885f9cf5719309c99/spring-boot-samples/spring-boot-sample-tomcat).

Without the change YourKit reports 2192 duplicate Strings, with the change it's 1126 (so a saving of 1066). It's hard to get the exact memory saving but scanning the UI for the top hitters we have:

| String | Duplicates | Waste
| --- | --- | -- |
"java.lang.String" | 2192 | 149112 |
"ACTION" | 542 | 38952 |
"boolean" | 191 | 10640 |
"void" | 204 | 9744 |
"Introspected parameter param0" | 80 | 8216 |
"int" | 155 | 7392 |
"[Ljava.lang.String;" | 42 | 3280 |

So it looks like at least 227336 bytes can be saved.

One other thing to consider is that I think these Strings are retained for the life of the application. They can't be GC'd because references remain until shutdown.

ChristopherSchultz:
I like the idea,  but I think. the implementation could be better. For instance:

1. There is no need to have two collections; just use one and cache everything.
2. The cache should be "weak" so that any value no longer used can be GC'd

rmaucher:
Ok, I'm not sure it is a very significant saving overall (probably not), but it's still more than I expected.

Yes, +1for Chris, it needs to be something like TomcatString.getString I suppose.

philwebb:
The other option would be to use `String.intern` for these limited cases. Either way, I'm happy to rework the PR if you have a specific direction you'd like me to take it?

ChristopherSchultz:
Good point: String.intern is already implemented, and it's already "weak". Less code, same result = better.

philwebb:
I've pushed an update that uses `String.intern` for those two places. With this change 2192 duplicates are reduced to 300 and I'm not seeing any difference in startup time.

rmannibucau:
Is generating a java class at build time instead of reading the xml possible? Then a preregistration in a registry would enable to bypass a lot of xml parsing and reflection which should give a real boost.

markt-asf:
I've been running YourKit with allocation tracking enabled and tracking each of the duplicated strings to the root cause. I started with the biggest duplicates and stopped somewhere around 700 bytes / per duplicated String. I ended up with a slightly different patch that covers the same duplicates and a few more. Start-up time appears to remain unaffected. Total saving ~245k.

I'm going to close this PR but I'll credit you in the commit message.

philwebb:
@rmannibucau I've opened another issue for that https://bz.apache.org/bugzilla/show_bug.cgi?id=63237. If I get a second I'll try some experiments.

philwebb:
@markt-asf Thanks Mark!

java/org/apache/tomcat/util/digester/SetPropertiesRule.java
```diff
@@ -62,7 +59,7 @@ public void begin(String namespace, String theName, Attributes attributes)
             if ("".equals(name)) {
                 name = attributes.getQName(i);
             }
-            String value = attributes.getValue(i);
+            String value = attributes.getValue(i).intern();
```
paplorinc: FYI: https://shipilev.net/jvm/anatomy-quarks/10-string-intern

> In almost every project we were taking care of, removing String.intern() from the hotpaths, or optionally replacing it with a handrolled deduplicator, was the very profitable performance optimization. Do not use String.intern() without thinking very hard about it, okay?

bokken: So Aleksey's article was used to determine that de-dup could help and then implemented it in the exact way that Aleksey proved was terrible?

markt-asf: For the relatively small number of strings concerned, we did not observe any performance issues and intern() was considered to be a reasonable approach. Happy to review contrary data, should any be provided.

bokken: What was the observed benefit from using String.intern()?
My take away from Aleksey's article was that String.intern should effectively never be used.
If the only observed benefit was removal of long lived "duplicate" strings, then the [G1 string dedup](http://openjdk.java.net/jeps/192) already provides the majority of that benefit with none of the drawbacks of intern usage.

markt-asf: See previous comments.

PR #144, opened 2019-03-06
  Variable adds final modifier

y987425112:
If these variables do not add final modifiers, there will be ambiguity and security risks.

### DIFF
--- a/java/org/apache/catalina/core/StandardContext.java
+++ b/java/org/apache/catalina/core/StandardContext.java
@@ -221,7 +221,7 @@ public class StandardContext extends ContainerBase
      * SCIs and other code may use the pluggability APIs to add listener
      * instances directly to this list before the application starts.
      */
-    private List<Object> applicationEventListenersList = new CopyOnWriteArrayList<>();
+    private final List<Object> applicationEventListenersList = new CopyOnWriteArrayList<>();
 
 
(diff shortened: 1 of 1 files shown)


markt-asf:
Potential security vulnerabilities should be reported privately to security@tomcat.apache.org. Not in a  PR or any other public forum.

That said, there are no valid security risks here so - fortunately - no harm is done.

The additional of final here (and the many, many other places tools such as UCDetector will identify automatically) is more a matter of style than anything else.

Making invalid claims of 'security risks' is not helpful. It undermines the credibility of the PR and makes it more likely it will be rejected.

We generally do not make changes purely for stylistic reasons. There is a code quality case that could be made for this change but it isn't a particularly strong one.

I am -1 on the PR as currently submitted due to the incorrect statement regarding security risks in the commit comment.

rmaucher:
That's a very useful reminder for me, if a PR is pulled, it *also* needs to have a decent description as well, otherwise misleading information gets into the commit info [and for sure it is the sort that would bring us trouble]. So -1 as well for this reason.

y987425112:
I am very sorry, this is the first time I submit the code, because I am Chinese, my English is a little poor, there are errors in the description. Thank you very much for your guidance. I will continue to work hard.

PR #146, opened 2019-03-08
Fix https://bz.apache.org/bugzilla/show_bug.cgi?id=63235

markt-asf:
Implement an alternative Charset caching strategy that does not load all
Charsets at Tomcat start. This reduces start time by ~30ms and does not,
based on the performance tests included in this commit, have a negative
impact on runtime look-ups.
It does require that the names of all supported Charsets are known to
Tomcat at compile time. The code has been tested with a range of JVMs
and a unit test is provided for testing new JVMs.

### DIFF
--- a/java/org/apache/tomcat/util/buf/B2CConverter.java
+++ b/java/org/apache/tomcat/util/buf/B2CConverter.java
@@ -25,9 +25,7 @@ import java.nio.charset.CharsetDecoder;
 import java.nio.charset.CoderResult;
 import java.nio.charset.CodingErrorAction;
 import java.nio.charset.StandardCharsets;
-import java.util.HashMap;
 import java.util.Locale;
-import java.util.Map;
 
 import org.apache.tomcat.util.res.StringManager;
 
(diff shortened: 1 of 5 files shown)


rmaucher:
Ok I guess. Wasn't the plan going to include hardcoding UTF-8 ?
The previous code still looks better to me (HashMap vs concurrent HashMap) but as you seem to care about resolving this now this is ok with me.

markt-asf:
Thanks for taking a look at the code.

I was concerned about the HashMap vs ConcurrentHashMap as well but in the tests included with the patch the difference was very small. Averaged over several runs it looked like ConcurrentHashMap was slower but by less than 1ns per lookup on average. That seemed small enough not to worry about to me.

In terms of special handling of UTF-8 my plan was only to pre-load the cache for UTF-8 and ISO-8859-1. I thought about adding other charsets but given ISO-8859-1 is the spec default and nearly everyone uses UTF-8 those seemed like the best place to start. Anything else I could think of just added more complexity and reduced performance.

I'm currently leaning towards applying this after the next 9.0.x tag so there is still plenty of time for folks to review and suggest improvements.

wilkinsona:
In addition to the improved startup time, this change has an even bigger positive impact on Tomcat's memory footprint. With the latest 9.0 snapshot, I have an app using embedded Tomcat that will start with `-Xmx13M` but not with `-Xmx12M`. The addition of the change proposed here, allows the same app to start with `-Xmx9M`. That ~4MB saving in heap footprint is very welcome indeed.

PR #148, opened 2019-03-14
Remove duplicated comments

samslow:
Signed-off-by: samslow <shsvkdldj@likelion.org>

I don't know why this comments is here.

If there is a reason why the comment is here and I do not know, Abort this PR and Teach me if you can


markt-asf:
No reason. It looks like a leftover from the last major redesign.

PR #149, opened 2019-03-18
Adding ReDoS warning/documentation to RewriteValve

salgattas:
After reporting a potential DoS in "Rewrite Rules" to the Tomcat security team, it was decided that there was no bug in Tomcat itself, but rather in how a user sets up their Tomcat server. Thus, I was instructed by the security team to create a PR for updated documentation to better educate users on appropriate usage of Rewrite Rules. This commit added javadoc comments for the RewriteValve class, as instructed.

Furthermore, I'd like to update the documentation on this page as well, however I cannot find a mechanism to do so: https://tomcat.apache.org/tomcat-9.0-doc/rewrite.html

### DIFF
--- a/java/org/apache/catalina/valves/rewrite/RewriteValve.java
+++ b/java/org/apache/catalina/valves/rewrite/RewriteValve.java
@@ -52,6 +52,18 @@ import org.apache.tomcat.util.file.ConfigFileLoader;
 import org.apache.tomcat.util.file.ConfigurationSource;
 import org.apache.tomcat.util.http.RequestUtil;
 
+/**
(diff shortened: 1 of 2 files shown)


markt-asf:
That page is generated from this file:
https://github.com/apache/tomcat/blob/master/webapps/docs/rewrite.xml

salgattas:
Great, thanks. I just included a statement for the documentation in there as well.

markt-asf:
Thanks. I applied the patch along with a couple of formatting tweaks, a cross-reference from the security how to page and a change log entry (giving you credit for the change).

PR #150, opened 2019-03-20
Add compiler source and target support for all Java versions over 10

tcollignon:
Hi,

I maintain the maven plugin to compile JSP : https://github.com/leonardehrenfried/jspc-maven-plugin

We have receive an issue when someone trying to compile JSP with Java 11 (to obtain target jsp class in Java 11 bytecode) see https://github.com/leonardehrenfried/jspc-maven-plugin/issues/39

I have look at the JDTCompiler source code and it appear that it handle only Java 10.

So I propose this modifications : 

- If compiler source or target is set, and value is not known by ECJ constant, we even set this directly and not fallback to Java 1.8.
Java move faster now, and probably ECJ will not move as faster.
- If compiler source or target is not set, we fallback to Java 1.8 as previous

WDYT ?

Thanks you all

### DIFF
--- a/java/org/apache/jasper/compiler/JDTCompiler.java
+++ b/java/org/apache/jasper/compiler/JDTCompiler.java
@@ -324,6 +324,9 @@ public class JDTCompiler extends org.apache.jasper.compiler.Compiler {
             } else if(opt.equals("11")) {
                 settings.put(CompilerOptions.OPTION_Source,
                              CompilerOptions.VERSION_11);
+            } else if(opt.equals("12")) {
+            // Constant not available in latest ECJ version
+                settings.put(CompilerOptions.OPTION_Source, "12");
             } else {
                 log.warn(Localizer.getMessage("jsp.warning.unknown.sourceVM", opt));
                 settings.put(CompilerOptions.OPTION_Source,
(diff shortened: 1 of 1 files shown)


markt-asf:
We want to keep the allowed values to the valid ones. 7.0.x has the right sort of code in place for Java 10 support. We should replicate that. We should also check for ECJ updates - keeping in mind the minimum Java versions of reach Tomcat release.

tcollignon:
Thx for your anwser
So I will wait for ECJ right ? 
Or maybe I can try to add options to Java 11 in ECJ ?

tcollignon:
Ok sorry I didn't understand the first comment

So I will update the PR like you said

markt-asf:
I actually just done this for Java 11. Could you do the same for Java 12?

tcollignon:
Ok I have add 11 and 12, I wait for your commit, and I will update this with just 12 if necessary

thx

tcollignon:
Hi

I have update the code, and squash commit

Tell me if it's ok for you

thx

markt-asf:
Hmm. If you pass "12" or "13" directly to ECJ it silently switches to the latest version it knows - currently 11. Whereas Tomcat would log a warning and switch to the minimum Java version required by the spec (Java 8 for Tomcat 9). Applying this patch would mean "12" was equivalent to "11" until ECJ supports Java 12. The silent switching makes me uneasy. I'm going to see if there is a way to get at the version actually being used. From that we could generate a suitable warning.

markt-asf:
Thanks. I took your patch and extended to cover Java 13 as well and added a check to log a warning if an unsupported version was used and ECJ fell back to the latest.

tcollignon:
Nice, thx you a lot @markt-asf

PR #152, opened 2019-04-03
Add async IO API to NIO

rmaucher:
It has very small performance improvements (except for "sendfile" situations) and non blocking frame read for HTTP/2. The performance looks rather similar to NIO2 as well, it is hard to find a very meaningful difference in my basic tests.
This should allow actual use of the API by users (as long as the APR connector is not used), and also using it in Tomcat for HTTP/3 once NIO UDP capability is added to the endpoint.
New "useAsyncIO" flag on the connector (defaults to true; could be false for now though).


rmaucher:
Note: I will only merge this after 9.0.18.

csutherl:
Since you opened a PR, can you squash it down to one commit since it's all for the same feature?

PR #153, opened 2019-04-04
Add async API for NIO

rmaucher:
Includes NIO vectored IO calls.


rmaucher:
Since I got no objections, I merged the code.

PR #155, opened 2019-04-08
Don't place CrawlerSessionManagerValve into session, place data-holder only. For 8.5.x

mlem:
It helps to serialize this sessions with libraries like kryo.

Background:
we've found out, that our kryo library serializes the whole valve.
Our stack indicates, that jdk internals are getting serialized into the
sessions. It's unnecessary and can be solves more easily with this fix.


markt-asf:
Thanks for the PR. I cherry-picked the 9.0.x fix (with the minor adjustments) and then made the 8.5.x specific changes.

PR #160, opened 2019-04-24
Add javadoc to describe IntrospectionUtils.isInstance

ChristopherSchultz:


isapir:
@ChristopherSchultz Should I merge it or are you just testing things out?  

Not sure why you opened a PR rather than push it directly given that this is not a change that requires much review.

ChristopherSchultz:
Just exercising everything to make sure it works how I expect it to work.

For example, Eclipse crapped its pants when I tried to import my fork of tomcat-master. It seems like the GitHub 2FA basically makes Eclipse stop working. I'm sure it has something to do with cached credentials in one of the dozen or so keychains that are in play, here.

isapir:
Is your remote configured for https:// (rather than git://)?.  Run `git remote -v` to view the remote URLs.  I was trying at first to use the git:// scheme and it didn't work.

PR #161, opened 2019-04-25
Early check end of input condition to avoid ArrayIndexOutOfBoundsExceptions

uhees:
Exceptions are expensive and should only be thrown in an exceptional state. End of input is no exception but should be expected. With the previous code a perfectly valid hostname will produce and process two ArrayIndexOutOfBoundsExceptions at its end, one in HttpParser.isAlpha() and one in HttpParser.isNumeric(), before detecting the end of input.


markt-asf:
Thanks for the PR. I applied (rather than merged it) along with a update to the changelog that included crediting you as author.

PR #163, opened 2019-05-09
Remove multiple pollers from NIO

rmaucher:
NIO2 manages without it, so it should work fine overall. This allows
removing a lot of complexity and disconnect between channel <-> wrapper
<-> poller, as both now use a known shared poller.


rmaucher:
Commited as e813ae0d9329ebf4b95c02043c39c676edb47d3c since nobody complained about it.

gaocher:
Just one question: I think multiple poller threads would have more performance on multi-core CPU. Is there any mistake in my mind? Could you help me explain why do we need to do this change (remove multiple pollers)?

markt-asf:
Yes, you are mistaken. Multiple poller threads adds complexity with no measurable performance gain.

PR #165, opened 2019-05-13
 Implementation of SameSite cookie attribute

farnulfo:
# Introduction
Hi folks, this pull request is an attemp to implement the SameSite cookie attribute according to https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-03 .

The main objective is to reduce the CSRF vulnerability.
Some articles : 
https://web.dev/samesite-cookies-explained
https://www.owasp.org/index.php/SameSite

As this my first real pull request I'm listening to you :-)

# Implementation:
This implementation is on two level:
- In Cookie / CookieGenerator classes
  - Add variable and getter/setter for this attribute
  - Generate the Set-Cookie header with this attribute if needed
- In Apache Tomcat Context and subclasses
  - Enable the use of SameSite attribute for the session cookie (JSESSIONID) which can’t be easily done

Implementation details:
Classes patched based on what was needed to add HttpOnly (thanks to Mark Thomas comment http://tomcat.10.x6.nabble.com/Support-SameSite-cookie-attribute-in-Tomcat-td5075308.html that helped to identify which classes to edit).

Design :
Name of the cookie attribute : SameSiteEnforcement (with getSameSite/setSameSite methods)
No boolean/method like isSameSite:
If not null, the “SameSite” attribute with only allow the following valid value : None, Lax, Strict
I don’t see the need for a isSameSite because we need to get the value which is not a boolean (not the case with httpOnly attribute).
Choice open for discussion. FYI Undertow has isSame and (get|set)SameSiteMode : https://github.com/undertow-io/undertow/pull/499

To set the SameSiteEnforcement for session cookie, set it on the Context Container like httpOnly before it was true by default or before it was part of Java EE standard.

Not sure about ApplicationSessionCookieConfig.java modification : do the ApplicationSessionCookieConfig may be have a samesitesite ?

# Next ?
If you’re ok with the idea, I can go further :
Add/Polish Javadocs
Polish context.xml documentation
Add tests 
...

Thansk !

### DIFF
--- a/java/javax/servlet/SessionCookieConfig.java
+++ b/java/javax/servlet/SessionCookieConfig.java
@@ -83,6 +83,10 @@ public interface SessionCookieConfig {
     public void setHttpOnly(boolean httpOnly);
 
     public boolean isHttpOnly();
+    
+    public void setSameSiteEnforcement(String sameSiteEnforcement);
+    
+    public String getSameSiteEnforcement();
 
     /**
      * Sets the secure flag for the session cookie.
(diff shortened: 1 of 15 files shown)


farnulfo:
I'm really sorry I didn't see the PR https://github.com/apache/tomcat/pull/162 .
I closed mine.

PR #167, opened 2019-05-21
Use test -t command to check if STDOUT is opened on a terminal

rjozwik:
Current check for TTY availability in catalina.sh doesn't work if non-English locales are installed and activated. This fix uses `test` command instead of `tty` command.


markt-asf:
Back-ported to 8.5.x and 7.0.x as well.

PR #169, opened 2019-06-16
Use STDIN for terminal availability check

rjozwik:
My previous patch for terminal availability check (using STDOUT), breaks when catalina.sh output is used in a pipe. This patch switches test to be based on STDIN. It better resembles behaviour of `tty` command previously used here and hopefully doesn't have any other side effects.


markt-asf:
Thanks for catching this and providing the fix. I applied to manually so I could include a changelog entry.

PR #170, opened 2019-06-18
Connections validated without explicit validation query leave a transaction open

davoustp:
Component: jdbc-pool

When connection validation is enabled and no validation query is defined, the PooledConnection class validates the connection using `Connection.isValid(int)`.

Problem is that this JDBC driver method may (and usually does) executes a query against the backend database, which actually opens a transaction.

Quote from Javadoc:
> The query submitted by the driver to validate the connection shall be executed in the context of the current transaction.

As a result, the connection then has a pending transaction (it opened one if none existed), which then leads to an error when the client code using jdbc-pool attempts to perform operations such as changing the transaction isolation level or the auto-commit flag.

This problem has already been taken care of when a validation query is provided - by silently committing or rollbacking the transaction only when auto-commit is set to false.
The same approach is used here to fix this issue.

In addition to this, I did not find any unit test around the connection validation feature, so I did write these prior to writing this fix (included in the PR), allowing to check that the entire connection validation code works as expected.

### DIFF
--- a/modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/PooledConnection.java
+++ b/modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/PooledConnection.java
@@ -514,11 +514,13 @@ public class PooledConnection implements PooledConnectionMBean {
             try {
                 if (connection.isValid(validationQueryTimeout)) {
                     this.lastValidated = now;
+                    silentlyCommitTransactionIfNeeded();
                     return true;
                 } else {
                     if (getPoolProperties().getLogValidationErrors()) {
                         log.error("isValid() returned false.");
                     }
+                    silentlyRollbackTransactionIfNeeded();
                     return false;
                 }
             } catch (SQLException e) {
@@ -527,6 +529,7 @@ public class PooledConnection implements PooledConnectionMBean {
                 } else if (log.isDebugEnabled()) {
                     log.debug("isValid() failed.", e);
                 }
+                silentlyRollbackTransactionIfNeeded();
                 return false;
             }
         }
@@ -543,6 +546,7 @@ public class PooledConnection implements PooledConnectionMBean {
             stmt.execute(query);
             stmt.close();
             this.lastValidated = now;
+            silentlyCommitTransactionIfNeeded();
             return true;
         } catch (Exception ex) {
             if (getPoolProperties().getLogValidationErrors()) {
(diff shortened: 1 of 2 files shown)


ChristopherSchultz:
I'm curious why you do not call `silentlyRollbackTransactionIfNeeded` in the `finally` clause every  time, instead of call it individually in several places *and also* in the `finally` block?

davoustp:
You mean that `silentlyRollbackTransactionIfNeeded()` at line 532?
There is no `finally` block in this `try` statement.
The `try` block needs either a commit or rollback, depending on the returned value from `isValid(int)`, and we need to rollback whenever an `Exception` is raised (in the `catch` block).

There is one `finally` in the other (existing) code block at line 560: this was already in place, so left it here.

I believe that this is actually an issue: the `rollback` is issued even if the `try` block did not raise any exception and committed its changes successfully - meaning that, unless the driver knows that this is a no-op (which I doubt), it will issue an unnecessary query (therefore a round-trip) to the DB.

ChristopherSchultz:
Ah, okay. I was only looking at the diff and it wasn't clear that code was in two different locations.

As for the "extra" `ROLLBACK`, you could always maintain a flag to avoid a second call. I'm sure the driver shouldn't be trying to "know" whether or not a `COMMIT` or `ROLLBACK` will be a no-op, and every programmer should expect that calling `rollback()` or `commit()` will result in the corresponding command *always* being sent to the server. So if you want to be completely tidy, you could maintain a little state for that.

davoustp:
Agreed, the driver is expected to be as thin and stateless as possible.
Indeed, using a flag to track if the transaction has been cleared or not is and performing the rollback in the latter case in a `finally` block is a much better solution as it would simplify the code and, even more importantly, it will also cover the corner cases which are not taken care of with by my initial commit (unchecked exceptions, or `Error`s).
I'll do this for both code paths (with and without a validation query) and commit the changes to the PR so we can have another look to check if there are some remaining issues.

davoustp:
Done. Please let me know what you think.

davoustp:
Hi, any idea if and when this fix will be merged into a Tomcat release?
Thx

markt-asf:
Thanks for the patch and especially for the test case. I made a few minor changes (formatting, debug logging, license header) and have applied the fix to master (9.0.x) and back-ported it to 8.5.x and 7.0.x.

davoustp:
Thx Mark!

PR #171, opened 2019-06-20
Added new interface for pluggable paramater configuration

wjase:
Signed-off-by: Jason Wraxall <wjase@yahoo.com.au>

Hi there - we upgraded tomcat-embed-core to 9.0.21 and found our code broke when JSSEUtilBase.getParameters visibility was tightened from protected to private.
Seeing no easy way to implement OCSP certificate checking with SOFT_FAIL, I implemented an interface to allow pluggable parameter configuration. Feel free to ask if you have any more qns or suggestions for the code. Cheers

### DIFF
--- a/java/org/apache/tomcat/util/net/SSLUtilBase.java
+++ b/java/org/apache/tomcat/util/net/SSLUtilBase.java
@@ -46,6 +46,7 @@ import java.util.List;
 import java.util.Locale;
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


rmaucher:
Tomcat maintains API compatibility in major branches (unless there's an extremely good reason to do otherwise). As a result, these methods should be reverted to protected.

PR #172, opened 2019-06-21
Process request range as specified in RFC 7233

zhanhb:


FSchumacher:
This seems to be a bigger change. It would be really awesome, if you could add junit tests that show the old misbehaviour and the new more conformant behaviour.
I haven't looked too hard at the changes at the moment, though.

markt-asf:
+1. I was thinking list out issues, separate commit (and PR?) for each and some unit tests would make this easier to review.
I suspect some parts could be applied easily whereas others will require more discussion.

zhanhb:
Thanks for read the changes carefully. I'll add test cases soon. IMO, most changes can be treated as bug fix or clean the source rather than behavior changed.

zhanhb:
RFC 7233 `If-Range should be compared with strong validator.` This can't be applied. I test on chrome 75 and find Head If-Range is sent. Maybe this will be like http 302.

zhanhb:
Before I write unit tests, maybe I need to explain what I my PR do.
[Line L559](https://github.com/apache/tomcat/pull/172/files#diff-355391a6aff99c525e8cd9165efd395dL559) I don't think is a good practise to referrer package visible class in another source file.
[Line R602-R603](https://github.com/apache/tomcat/pull/172/files#diff-355391a6aff99c525e8cd9165efd395dR602-R603) Add comment, which might be done in tomcat 10 or future release.
[Line L1016-R1018](https://github.com/apache/tomcat/pull/172/files#diff-355391a6aff99c525e8cd9165efd395dL1016-R1018) Clean the source. Might be incompatible if a user define a class extends DefaultServlet and parse the range himself with parameter or another header field. But maybe we should not consider such case. We should make the method `parseRange` meaningful and easy to maintain.
[Line R1448-R1451](https://github.com/apache/tomcat/pull/172/files#diff-355391a6aff99c525e8cd9165efd395dR1448-R1451) Report a full range for such as HEAD request. All other request methods should not contains a `Range` header.
[Line L1464-R1473](https://github.com/apache/tomcat/pull/172/files#diff-355391a6aff99c525e8cd9165efd395dL1464-R1473) As specified in [rfc7233#3.2](https://tools.ietf.org/html/rfc7233#section-3.2), apply strong validator on etags. (My opinion now is that this should not be applied.)
[Line L1471-R1481](https://github.com/apache/tomcat/pull/172/files#diff-355391a6aff99c525e8cd9165efd395dL1471-R1481) Our etag depends the modified time and sizeof the file. This won't change too much.
[Line L1440-R1441](https://github.com/apache/tomcat/pull/172/files#diff-355391a6aff99c525e8cd9165efd395dL1490-R1441), [L1490](https://github.com/apache/tomcat/pull/172/files#diff-355391a6aff99c525e8cd9165efd395dL1490), [R1497](https://github.com/apache/tomcat/pull/172/files#diff-355391a6aff99c525e8cd9165efd395dR1490) Make method parseRange meaningfull and easier to maintain.
[Line L1491-L1492](https://github.com/apache/tomcat/pull/172/files#diff-355391a6aff99c525e8cd9165efd395dL1491-L1492), [R1498-R1501](https://github.com/apache/tomcat/pull/172/files#diff-355391a6aff99c525e8cd9165efd395dR1498-R1501) [rfc7233#3.1](https://tools.ietf.org/html/rfc7233#section-3.1) says that we must ignore a Range header field that contains a range unit it does not understand. Not change too much for a comman used user agent won't send such header "Range: none".
[Line L1508-R1519](https://github.com/apache/tomcat/pull/172/files#diff-355391a6aff99c525e8cd9165efd395dL1508-R1519), [L1517-R1528](https://github.com/apache/tomcat/pull/172/files#diff-355391a6aff99c525e8cd9165efd395dL1517-R1528), [L1530-R1544](https://github.com/apache/tomcat/pull/172/files#diff-355391a6aff99c525e8cd9165efd395dL1530-R1544), [L1550-R1564](https://github.com/apache/tomcat/pull/172/files#diff-355391a6aff99c525e8cd9165efd395dL1530-R1564), [L1570-R1579](https://github.com/apache/tomcat/pull/172/files#diff-355391a6aff99c525e8cd9165efd395dL1570-R1579) no change, just clean the source.
[Line R1533-R1541](https://github.com/apache/tomcat/pull/172/files#diff-355391a6aff99c525e8cd9165efd395dR1533-R1541), [L1543-R1564](https://github.com/apache/tomcat/pull/172/files#diff-355391a6aff99c525e8cd9165efd395d#L1543-R1564), [L2664-R2673](https://github.com/apache/tomcat/pull/172/files#diff-355391a6aff99c525e8cd9165efd395d#L2664-R2673) For a request who is requesting a single range, we response a modified range range than 416.
For a request who is requesting multiple ranges, we will remove some unsatisfiable ranges and response remain ranges. Behaviour might be incompative we will reponse a single range style for multiple range requests.
For a put request, original implementation will be OK if such request is sent "Content-Range: 1-5/4".
The meaning for method `Range.validate` is `isSatisfiable` for GET method, `isValid` for PUT method.

zhanhb:
I search the source in [chromium/chromium?q=ParseRangeHeader](https://github.com/chromium/chromium/search?q=ParseRangeHeader&unscoped_q=ParseRangeHeader), and found they won't send multiple range request. And now I want to know what user agent will send multi-range request. curl?

kkolinko:
You have to create an issue in Bugzilla and describe your problem there. (So that it could be properly referenced form the Change log, as well as if there is more than one pull request for the issue).

zhanhb:
OK, I'll create more PRs.

PR #175, opened 2019-06-23
Apply the suggestion in rfc7233

zhanhb:


markt-asf:
Thanks. I have applied 2 of the 3 changes to 9.0.x (range units and etag) with some additional refactoring / clean-up as well as additional comments.
I thin the remaining change needs to be made optional as many clients have implemented partial PUT even if the intention of the RFC 2616 authors was not to allow it. This is next on my TODO list followed by back-porting all the Rnage improvements inspired by your PRs.

zhanhb:
For a put request, usually it won't have a header 'Range', The header specified in rfc 2616 is 'Content-Range'.
What will happen to a put request if header 'Range' is specified?
Put the content in specified URI and then return partial content of the entity. We can see existed source in tomcat, either 201 or 204 is returned, both have no response body content.
IMO, this change can also be applied too.

markt-asf:
Thanks for the the review. I had mis-remembered how partial PUTs were implemented. I'll take another pass at the patch.

markt-asf:
I think everything is complete now. Feel free to re-open if you spot something I missed.

zhanhb:
Good job, there is no issue with the range request now.

PR #176, opened 2019-06-29
CoyoteAdapter: fix out-of-bounds read in checkNormalize

alpire:
On malformed requests, checkNormalize would throw an ArrayIndexOutOfBoundsException leading to a 500 response. This change fixes checkNormalize to return false instead of throwing exception on those inputs, and adds a few tests to check the new functionality.

For the record, the exception is below:

```
java.lang.ArrayIndexOutOfBoundsException: -1
        at org.apache.catalina.connector.CoyoteAdapter.checkNormalize(CoyoteAdapter.java:1275) ~[tomcat-embed-core-9.0.19.jar!/:9.0.19]
        at org.apache.catalina.connector.CoyoteAdapter.postParseRequest(CoyoteAdapter.java:647) ~[tomcat-embed-core-9.0.19.jar!/:9.0.19]
        at org.apache.catalina.connector.CoyoteAdapter.service(CoyoteAdapter.java:337) ~[tomcat-embed-core-9.0.19.jar!/:9.0.19]
        at org.apache.coyote.http11.Http11Processor.service(Http11Processor.java:408) ~[tomcat-embed-core-9.0.19.jar!/:9.0.19]
        at org.apache.coyote.AbstractProcessorLight.process(AbstractProcessorLight.java:66) [tomcat-embed-core-9.0.19.jar!/:9.0.19]
        at org.apache.coyote.AbstractProtocol$ConnectionHandler.process(AbstractProtocol.java:836) [tomcat-embed-core-9.0.19.jar!/:9.0.19]
        at org.apache.tomcat.util.net.NioEndpoint$SocketProcessor.doRun(NioEndpoint.java:1747) [tomcat-embed-core-9.0.19.jar!/:9.0.19]
        at org.apache.tomcat.util.net.SocketProcessorBase.run(SocketProcessorBase.java:49) [tomcat-embed-core-9.0.19.jar!/:9.0.19]
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [na:1.8.0_212]
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [na:1.8.0_212]
        at org.apache.tomcat.util.threads.TaskThread$WrappingRunnable.run(TaskThread.java:61) [tomcat-embed-core-9.0.19.jar!/:9.0.19]
        at java.lang.Thread.run(Thread.java:748) [na:1.8.0_212]
```

### DIFF
--- a/java/org/apache/catalina/connector/CoyoteAdapter.java
+++ b/java/org/apache/catalina/connector/CoyoteAdapter.java
@@ -1252,11 +1252,6 @@ public class CoyoteAdapter implements Adapter {
 
         int pos = 0;
 
-        // An empty URL is not acceptable
-        if (start == end) {
-            return false;
-        }
-
         // Check for '\' and 0
         for (pos = start; pos < end; pos++) {
             if (c[pos] == '\\') {
--- a/java/org/apache/catalina/connector/CoyoteAdapter.java
+++ b/java/org/apache/catalina/connector/CoyoteAdapter.java
@@ -1252,6 +1252,11 @@ public class CoyoteAdapter implements Adapter {
 
         int pos = 0;
 
+        // An empty URL is not acceptable
+        if (start == end) {
+            return false;
+        }
+
         // Check for '\' and 0
         for (pos = start; pos < end; pos++) {
             if (c[pos] == '\\') {
@@ -1269,6 +1274,11 @@ public class CoyoteAdapter implements Adapter {
                     return false;
                 }
             }
+        }
+
+        // The URL must start with '/'
+        if (c[start] != '/') {
+            return false;
         }
 
         // Check for ending with "/." or "/.."
--- a/test/org/apache/catalina/connector/TestCoyoteAdapter.java
+++ b/test/org/apache/catalina/connector/TestCoyoteAdapter.java
@@ -326,6 +326,7 @@ public class TestCoyoteAdapter extends TomcatBaseTest {
     @Test
     public void testNormalize01() {
         doTestNormalize("/foo/../bar", "/bar");
+        doTestNormalize("..", null);
     }
 
     private void doTestNormalize(String input, String expected) {
@@ -342,6 +343,29 @@ public class TestCoyoteAdapter extends TomcatBaseTest {
             Assert.assertTrue(result);
             Assert.assertEquals(expected, mb.toString());
         }
+    }
+
+    @Test
+    public void testCheckNormalize() {
+        doTestCheckNormalize("/url", true);
+
+        doTestCheckNormalize("", false);
+        doTestCheckNormalize("..", false);
+        doTestCheckNormalize("/.", false);
+        doTestCheckNormalize("/..", false);
+        doTestCheckNormalize("/./", false);
+        doTestCheckNormalize("//", false);
+        doTestCheckNormalize("/../", false);
+        doTestCheckNormalize("\\", false);
+        doTestCheckNormalize("\0", false);
+    }
+
+    private void doTestCheckNormalize(String input, boolean expected) {
+        MessageBytes mb = MessageBytes.newInstance();
+        mb.setChars(input.toCharArray(), 0, input.length());
+
+        boolean result = CoyoteAdapter.checkNormalize(mb);
+        Assert.assertEquals(expected, result);
     }
 
 
--- a/test/org/apache/catalina/connector/TestCoyoteAdapter.java
+++ b/test/org/apache/catalina/connector/TestCoyoteAdapter.java
@@ -349,7 +349,6 @@ public class TestCoyoteAdapter extends TomcatBaseTest {
     public void testCheckNormalize() {
         doTestCheckNormalize("/url", true);
 
-        doTestCheckNormalize("", false);
         doTestCheckNormalize("..", false);
         doTestCheckNormalize("/.", false);
         doTestCheckNormalize("/..", false);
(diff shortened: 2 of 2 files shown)


java/org/apache/catalina/connector/CoyoteAdapter.java
```diff
@@ -1252,6 +1252,11 @@ public static boolean checkNormalize(MessageBytes uriMB) {
 
         int pos = 0;
 
+        // An empty URL is not acceptable
+        if (start == end) {
+            return false;
+        }
+
         // Check for '\' and 0
```
markt-asf: I believe the above return would represent unreachable code. Given this is on the code path every request must pass through, we do not want to add any unnecessary tests however trivial. If you can find an input to `CoyoteAdapter.postParseRequest()` that would result in this code being executed then that would likely be sufficient justification to add this test.

java/org/apache/catalina/connector/CoyoteAdapter.java
```diff
@@ -1271,6 +1276,11 @@ public static boolean checkNormalize(MessageBytes uriMB) {
             }
         }
 
+        // The URL must start with '/'
+        if (c[start] != '/') {
+            return false;
+        }
+
         // Check for ending with "/." or "/.."
```
markt-asf: The same goes for this test.

test/org/apache/catalina/connector/TestCoyoteAdapter.java
```diff
@@ -326,6 +326,7 @@ public boolean isResponseBodyOK() {
     @Test
     public void testNormalize01() {
         doTestNormalize("/foo/../bar", "/bar");
+        doTestNormalize("..", null);
     }
```
markt-asf: We aim for one test per method. This would be a candidate for `testNormalize02()`

test/org/apache/catalina/connector/TestCoyoteAdapter.java
```diff
@@ -344,6 +345,29 @@ private void doTestNormalize(String input, String expected) {
...
+        doTestCheckNormalize("", false);
+        doTestCheckNormalize("..", false);
+        doTestCheckNormalize("/.", false);
+        doTestCheckNormalize("/..", false);
+        doTestCheckNormalize("/./", false);
+        doTestCheckNormalize("//", false);
+        doTestCheckNormalize("/../", false);
+        doTestCheckNormalize("\\", false);
+        doTestCheckNormalize("\0", false);
+    }
```
markt-asf: One test per method here as well. `testCheckNormalize01()`, `testCheckNormalize02()` etc.

While I am not against direct testing of any value that could be passed to `checkNormalize()`, I think the test would be more useful if the appropriate input was passed to `normalize()`, `convertURI()` and then `checkNormalize()`. I appreciate `checkNormalize()` is a public method and could - in theory - receive any input but in this instance the testing needs to focus on what is possible given Tomcat's usage of the method.

java/org/apache/catalina/connector/CoyoteAdapter.java
alpire: I do not have an input that would trigger a 500 if we removed that check. I understand the performance consideration, so I'll remove it.

java/org/apache/catalina/connector/CoyoteAdapter.java
alpire: I do have an input that triggers a 500 response if that check is removed:
```
00000000: 202e 2e20 3020 0a                         .. 0 .
```

That input trigger the exception that's in the PR description.

test/org/apache/catalina/connector/TestCoyoteAdapter.java
alpire: I split the tests into different functions and removed a few of them since we want to focus the `checkNormalize()` tests to what is possible given its usage.

I did not update the tests to go through `normalize` -> `convertURI` -> `checkNormalize`, but I'd be happy to do this if you think that'd be more appropriate here.

java/org/apache/catalina/connector/CoyoteAdapter.java
markt-asf: I don't see how that can trigger a 500. The request line:
`GET  %20%2e%2e%20%30%20%0a HTTP/1.1`
is going to result in a 400 response in all current Tomcat versions.
Please provide a request line example that triggers the exception in the original report.

alpire: Sorry for the confusion. What I posted was the entire query, i.e. you can reproduce the issue with:
```
$ echo -ne "\x20\x2e\x2e\x20\x30\x20\x0a" | nc localhost 8081
HTTP/1.1 500
Date: Tue, 02 Jul 2019 13:55:20 GMT
Connection: close
```

Looking more into it, you can also be reproduced with a normal looking HTTP query:
```
$ echo -ne "GET .. HTTP/1.1\r\n\r\n" | nc localhost 8081
```

markt-asf: Thanks. That is what I needed. That should be caught and rejected earlier. I wonder what is going wrong.

alpire: I was surprised to see that `checkNormalize` was still called when `normalize` returns false in postParseRequest. Maybe that'd be a better fix?

markt-asf: Agreed. I reached the same conclusion. I'll get that fixed shortly. Thanks for highlighting this.

markt-asf:
Many thanks for raising this issue. I think we have tracked down the root cause and fixed it. Note that 7.0.x was not affected. I am, therefore, closing this PR. If we have missed something please feel free to re-open it.

PR #177, opened 2019-07-07
Code polish: replace classic for-loop with enhanced

govi20:
ApplicationContext has classic-for loops, these for loops can be replaced with enhanced for-loop to improve code readability.


ChristopherSchultz:
The patch is smaller is you don't replace `listener` with `servletContextAttributeListener` everywhere. Any reason to re-name this local variable reference?

govi20:
Yes. `for-loop's` temp and `ServletContextAttributeListener's` old name are clashing. I don't think using `tempListener` in a `for-loop` declaration is a good idea.

markt-asf:
`for (Object obj : listeners) {...` results in a cleaner patch. I've applied that version to master and will back-port shortly. I've also applied some other clean-up and will back-port that too.

govi20:
Cool Thanks!

PR #179, opened 2019-07-20
PropertySource: Add an environment variable based source

thomasmey:
When tomcat runs in an Openshift based container a Secret containing passwords
can be map as environment variables (with an additional prefix).
An webapp containing an embedded context.xml which defines JDBC datasources and
placeholder variables can be used with this new PropertySource to easily inject
configuration from a Secret or ConfigMap.


markt-asf:
Fixed in:
- master for 9.0.25 onwards
- 8.5.x for 8.5.46 onwards
- 7.0.x for 7.0.97 onwards

PR #183, opened 2019-08-02
preload driver in connection pool

rmannibucau:
share it for all connections to not have mixed connections (multiple driver instances when fallbacking on TCCL)

Side note: the setters of DataSource(Proxy) should likely recreate the pool when called, not sure it is not already the case but looks fishy so a follow up issue can be needed

### DIFF
--- a/modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/PooledConnection.java
+++ b/modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/PooledConnection.java
@@ -17,6 +17,7 @@
 package org.apache.tomcat.jdbc.pool;
 
 
+import java.sql.Driver;
 import java.sql.DriverManager;
 import java.sql.SQLException;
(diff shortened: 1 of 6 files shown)


modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/PooledConnection.java
```diff
@@ -335,7 +308,7 @@ protected void connectUsingDriver() throws SQLException {
             }
         }
         if (connection==null) {
-            throw new SQLException("Driver:"+driver+" returned null for URL:"+driverURL);
+            throw new SQLException("Driver:"+parent.getDriver()+" returned null for URL:"+driverURL);
```
pirateskipper, line 311: ` throw new SQLException("Driver:" + parent.getDriver() + " returned null for URL:" + driverURL);`

rmannibucau:
workaround being to have multiple pools and due to the age of the PR I guess it can be closed

PR #189, opened 2019-08-14
Additional translations of Chinese

leeyazhou:
Additional translations of Chinese for tomcat manager

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -859,6 +859,16 @@
       <zipfileset file="${tomcat.manifests}/default.license"
         fullpath="META-INF/LICENSE" />
     </jar>
+    <jar jarfile="${tomcat.build}/lib/tomcat-i18n-zh.jar"
+      manifest="${tomcat.manifests}/default.manifest">
+      <fileset dir="${tomcat.classes}">
+        <include name="**/LocalStrings_zh.properties" />
+      </fileset>
+      <zipfileset file="${tomcat.manifests}/default.notice"
+        fullpath="META-INF/NOTICE" />
+      <zipfileset file="${tomcat.manifests}/default.license"
+        fullpath="META-INF/LICENSE" />
+    </jar>
     <jar jarfile="${tomcat.build}/lib/tomcat-i18n-fr.jar"
       manifest="${tomcat.manifests}/default.manifest">
       <fileset dir="${tomcat.classes}">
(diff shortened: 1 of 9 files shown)


markt-asf:
I'm all in favour of expanding the i18n support in Tomcat and I agree adding Chinese translations to 8.5.x is a good next step.

If you aren't already aware, Tomcat uses [POEditor](https://poeditor.com/projects/view?id=221603) to manage translations for 9.0.x and then back-ports changes to earlier versions.

I think there it should be possible to add significantly more Chinese translations than proposed in this PR and to do so in a semi-automated fashion that supports additional languages.

There are probably several different ways of doing this. My first idea is to use the utilities in org.apache.tomcat.buildutil.translate as a starting point, and write some code to support the following:

- For every key in 9.0.x, if the key exists in 8.5.x replace the 8.5.x English text with the 9.0.x version
- Manually review the diff and revert any changes that don't look right
- For every key in 9.0.x, if the key exists in 8.5.x and the English text is the same in both versions, copy the text for all other languages supported by 8.5.x from 9.0.x to 8.5.x.
- Repeat for 8.5.x and 7.0.x

Suggests to improve the process and/or patches to implement it welcome.

markt-asf:
I've completed the first two bullets for 9.0.x->8.5.x. There was only one change required to fix a key that had been added out of order.

leeyazhou:
> I'm all in favour of expanding the i18n support in Tomcat and I agree adding Chinese translations to 8.5.x is a good next step.
> 
> If you aren't already aware, Tomcat uses [POEditor](https://poeditor.com/projects/view?id=221603) to manage translations for 9.0.x and then back-ports changes to earlier versions.
> 
> I think there it should be possible to add significantly more Chinese translations than proposed in this PR and to do so in a semi-automated fashion that supports additional languages.
> 
> There are probably several different ways of doing this. My first idea is to use the utilities in org.apache.tomcat.buildutil.translate as a starting point, and write some code to support the following:
> 
> * For every key in 9.0.x, if the key exists in 8.5.x replace the 8.5.x English text with the 9.0.x version
> * Manually review the diff and revert any changes that don't look right
> * For every key in 9.0.x, if the key exists in 8.5.x and the English text is the same in both versions, copy the text for all other languages supported by 8.5.x from 9.0.x to 8.5.x.
> * Repeat for 8.5.x and 7.0.x
> 
> Suggests to improve the process and/or patches to implement it welcome.

Thank u. I'll finish translations on  [POEditor](https://poeditor.com/projects/view?id=221603) first.

markt-asf:
Just to let you know, I've written some utilities to back-port the translations and have run them for 9.0.x -> 8.5.x. I've also picked up the translations you've added so far at poeditor.
I'll work on 8.5.x -> 7.0.x next. Watch the Tomcat dev list for progress updates.

PR #191, opened 2019-08-17
Fix typo for prefect

KangZhiDong:
fix typo for prefect


markt-asf:
Thanks for the PR. I have applied the change (with a modified comment) to 9.0.x and 8.5.x.
Note that it isn't clear to me what you mean by "...for prefect" in the comment. I've dropped that part. "Fix typo" appears to describe the commit sufficiently.

PR #192, opened 2019-08-18
Fix typo

KangZhiDong:
Fix typo

### DIFF
--- a/java/org/apache/catalina/core/StandardWrapper.java
+++ b/java/org/apache/catalina/core/StandardWrapper.java
@@ -1253,7 +1253,7 @@ public class StandardWrapper extends ContainerBase
             return;
         unloading = true;
 
-        // Loaf a while if the current instance is allocated
+        // Load a while if the current instance is allocated
         // (possibly more than once if non-STM)
         if (countAllocated.get() > 0) {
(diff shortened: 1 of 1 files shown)


java/org/apache/catalina/core/StandardWrapper.java
```diff
@@ -1253,7 +1253,7 @@ public synchronized void unload() throws ServletException {
             return;
         unloading = true;
 
-        // Loaf a while if the current instance is allocated
+        // Load a while if the current instance is allocated
```
gmshake: Not typo here :)

qixiaobo: Means have a break ?

markt-asf:
It does. The use of loaf is meant to indicate that the thread waits for a while, idle, not doing anything.

qixiaobo:
maybe spin~

jfclere:
I love the definition "to spend time doing nothing, usually when you should be working": macmillandictionary.com

qixiaobo:
> I love the definition "to spend time doing nothing, usually when you should be working": macmillandictionary.com

intelligent

PR #196, opened 2019-08-26
fix JSSE_OPTS quoting

peteruhnak:
Otherwise if JSSE_OPTS is set with the same value, the quotation will cancel itself in the if condition.

```
if not ""-Djdk.tls.ephemeralDHKeySize=2048""
```

resulting in the following error

```
=2048"" was unexpected at this time.
```


FSchumacher:
Thanks for the patch. It has been applied to trunk and I will apply it to the other trees later.

PR #197, opened 2019-08-28
Remove unuse catch exception

KangZhiDong:
Remove unuse catch exception


KangZhiDong:
duplicate exception handling @markt-asf

PR #207, opened 2019-09-14
Fix typos

KangZhiDong:
remove redundant or reformat javadoc for clean(some can be ingored). fix typos, such as: a non-zero, a long, a utf-8, an array ,an html, an xa connection, an XAResource, an X-Powered-By, an OutputStream, an rpc, an XByteBuffer, an ArrayList, an abstract, an internal...


java/javax/servlet/http/Part.java
```diff
@@ -48,9 +48,6 @@
     public String getContentType();
 
     /**
-     * Obtain the name of the field in the multipart form corresponding to this
-     * part.
-     *
      * @return The name of the field in the multipart form corresponding to this
```
markt-asf: -1 to this and all similar changes. The first sentence in the Javadoc is used as the method description in the table of methods. By removing this text, the description is lost from that table. While I am never a fan of duplication, in this case it is better to have the duplication than not.

KangZhiDong:
i will make other changes in spare time @markt-asf

PR #209, opened 2019-09-28
BZ 63778: Jre7Compat cannot find getPseudoColumns with zero parameters

lgxbslgx:
Solve the MethodNotFoundException.
Add log when the exception occurs.

### DIFF
--- a/java/org/apache/tomcat/util/compat/Jre7Compat.java
+++ b/java/org/apache/tomcat/util/compat/Jre7Compat.java
@@ -16,6 +16,10 @@
  */
 package org.apache.tomcat.util.compat;
 
+import org.apache.juli.logging.Log;
+import org.apache.juli.logging.LogFactory;
+import org.apache.tomcat.util.res.StringManager;
+
 import java.io.OutputStream;
 import java.lang.reflect.Constructor;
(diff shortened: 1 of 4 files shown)


java/org/apache/tomcat/util/compat/Jre7Compat.java
```diff
@@ -32,6 +36,9 @@
 
 class Jre7Compat extends JreCompat {
 
+    private static final Log log = LogFactory.getLog(Jre7Compat.class);
+    private static final StringManager sm = StringManager.getManager(Jre7Compat.class);
```
govi20: The variable should be named as stringManager instead of sm?

markt-asf: No. It should be `sm` for consistency with the rest of the Tomcat code base.

markt-asf:
Thanks for the PR. As written, it generates unnecessary WARN log messages. DEBUG level messages are fine to report - for example - Java 8 wasn't found but WARN is not. I have added some log messages but with a slightly different approach that logs an ERROR if a problem like BZ 63778 is detected but otherwise uses DEBUG logging.

PR #210, opened 2019-10-01
BZ 63792: The tomcat log contains garbled code because of i18n

lgxbslgx:
Prevent garbled code when using ResourceBundle.


markt-asf:
This doesn't look right. While the properties files are stored in UTF-8 in source control, the build process should convert them to ISO-8859-1 with appropriate escapes so they work correctly with any Java version.

markt-asf:
I've confirmed that the 9.0.x build handles this correctly and calls the Ant Task native2ascii to convert the property files at build time.

lgxbslgx:
I have seen the native2ascii just now. Sorry for wasting your time.

PR #211, opened 2019-10-14
Fix semantic and punctuation in chinese

quaff:
I guess some translations was done by machine.


KangZhiDong:
> I guess some translations was done by machine.

    [you can modify and complete chinese translations at poeditor that manages all translations  . ](https://poeditor.com/projects/po_edit?id_language=31&id=221603)

markt-asf:
Thanks for the PR.
Because we manage translations via POEditor, and have imports/exports set up to keep POEditor and the source repository in sync, it is much easier for us if translation fixes are provided by POEditor.
I have transferred these changes to POEditor and will be importing them into Tomcat via that route.

markt-asf:
Applied via POEditor

PR #219, opened 2019-10-24
Fix https://bz.apache.org/bugzilla/show_bug.cgi?id=63865

johnkdev:
Add unset option to same-site cookies and pass through none value if set by user.


markt-asf:
Wow. I think you might just have submitted the perfect pull request. Bug fixed - yes. Docs updated - yes. Tests updated - yes. And special bonus points for including the change log. Thank you. I'll get this committed shortly.

PR #220, opened 2019-10-25
BZ 63879 - Remove unnecessary exception from NioSocketWrapper doClose

steves14:
https://bz.apache.org/bugzilla/show_bug.cgi?id=63879


rmaucher:
I preferred simply removing the exception, usually knowing the caller is not useful. Thanks for the report.

PR #221, opened 2019-11-01
RewriteMap with quoted parameters and ServiceLoader support

FSchumacher:
This PR combines two things:
* Allows quoted parameters for `RewriteValve` config and adds a new default method to `RewriteMap` to allow for more than one parameter
* Uses a `ServiceLoader` to make implementations of `RewriteMap` available in the httpd rewrite module style a-la `int:toupper`.

It is missing a lot of documentation and a few test cases.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -826,6 +826,7 @@
     <jarIt jarfile="${catalina.jar}"
       filesDir="${tomcat.classes}"
       filesId="files.catalina"
+      meta-inf="${tomcat.manifests}/catalina.jar"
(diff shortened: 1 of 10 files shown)


rmaucher:
This looks a bit overengineered, personally I am surprised to see the service loader being used for such a small feature [I added that map support because the original rewrite had it, and nobody ever cared]. We don't use the service loader functionality in Tomcat.

rmaucher:
I integrated everything (and a bit more) except the service loader part.

PR #222, opened 2019-11-06
EmbeddedServletOptions constructor has been refactored

dedovich-s:
Each value initializing moved to appropriate private method

### DIFF
--- a/java/org/apache/jasper/EmbeddedServletOptions.java
+++ b/java/org/apache/jasper/EmbeddedServletOptions.java
@@ -472,193 +472,216 @@ public final class EmbeddedServletOptions implements Options {
      * @param context The Servlet context
      */
     public EmbeddedServletOptions(ServletConfig config, ServletContext context) {
+        Enumeration<String> enumeration = config.getInitParameterNames();
 
-        Enumeration<String> enumeration=config.getInitParameterNames();
-        while( enumeration.hasMoreElements() ) {
-            String k=enumeration.nextElement();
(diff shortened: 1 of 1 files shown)


markt-asf:
This appears to be a purely cosmetic change.
Such changes can make back-porting more difficult unless the change is back-ported to all currently supported versions.
The change makes this class inconsistent with similar code throughout the Tomcat code base.
Changes such as this carry the risk of introducing regressions.
Unless a good reason can be provided for applying this, I am leaning towards closing this PR.

72er-de:
I think the constructor has a lot of lines of code and the idea to extract setter could be help to read the code. So I have ordered back the extracted methods so that we can compare with apache/master and perhaps I found a theoretical issue here: 
```
        // TODO: Check this behaviour, if scratchDir == null
        // then EmbeddedServletOptions is not correctly initiated!
        // Could only happen if System.getProperty("java.io.tmpdir")
        // is not set. -> unusual exception?
        if (scratchDir == null) {
            log.fatal(Localizer.getMessage("jsp.error.no.scratch.dir"));
            return;
        }
```
[EmbeddedServletOptions.zip](https://github.com/apache/tomcat/files/3871042/EmbeddedServletOptions.zip)

dedovich-s:
yes, my original intention was to reduce the number of lines in constructor and move appropriate code to init methods

markt-asf:
As per my previous comment, I am closing this PR without applying it.

PR #223, opened 2019-11-14
Log compression

habuzahra:
This modification has been done based on the below bugzilla

https://bz.apache.org/bugzilla/show_bug.cgi?id=62611

Below are the modifications that has been done during the task:

1-	org.apache.juli.FileHandler.compress();
2-	org.apache.juli.FileHandler.deleteRotatedFile();
3-	org.apache.catalina.valves compress();
4-	org.apache.catalina.valves.deleteRotatedFile();
5-    External configurations, where you can enable or disable the compression function based on new configurations keys to located in logging.properties
1catalina.org.apache.juli.AsyncFileHandler.compress = true
2localhost.org.apache.juli.AsyncFileHandler.compress = true
3manager.org.apache.juli.AsyncFileHandler.compress = true
4host-manager.org.apache.juli.AsyncFileHandler.compress = true


java/org/apache/juli/FileHandler.java
```diff
@@ -216,6 +218,12 @@ public FileHandler(String directory, String prefix, String suffix, Integer maxDa
     private Boolean rotatable;
 
 
+	/**
+     * Determines whether the log file should be compressed
+     */
+    private Boolean compress;
```
markt-asf: This needs to be documented in the class level Javadoc

java/org/apache/juli/FileHandler.java
```diff
@@ -393,6 +430,9 @@ private void configure() {
         if (suffix == null) {
             suffix = getProperty(className + ".suffix", ".log");
         }
+		if (compress == null) {
+			compress = Boolean.valueOf(getProperty(className + ".compress", "true"));
+		}
```
markt-asf: It looks like some tabs have crept in here. Tomcat uses 4 spaces for idents. I recommend enabling Checkstyle for you build as it will catch this sort of thing,

java/org/apache/tomcat/util/file/CompressFileUtils.java
```diff
@@ -0,0 +1,50 @@
+package org.apache.tomcat.util.file;
```
markt-asf: Needs an ALv2 header. Again Checkstyle will catch this.

java/org/apache/catalina/valves/AccessLogValve.java
```diff
@@ -37,6 +37,7 @@
 import org.apache.juli.logging.LogFactory;
 import org.apache.tomcat.util.ExceptionUtils;
 import org.apache.tomcat.util.buf.B2CConverter;
+import org.apache.tomcat.util.file.CompressFileUtils;
```
markt-asf: Why not use the JULI version since the dependency is already there rather than duplicating the code.

PR #230, opened 2019-12-11
Fix missing implicit TLD map entries.

ml86:
If TldResourcePath was already registred during parsing of jsp-config
section of web.xml, the implicit TLD map entries (derived from TLD files
URI) have not been added to uriTldResourcePathMap.

This was not in line with JSP2.2 specification section 7.3.4.


build.properties.default
```diff
@@ -306,3 +306,4 @@ bnd.loc=${base-maven.loc}/biz/aQute/bnd/biz.aQute.bnd/${bnd.version}/biz.aQute.b
 bndlib.home=${base.path}/bndlib-${bnd.version}
 bndlib.jar=${bndlib.home}/biz.aQute.bndlib-${bnd.version}.jar
 bndlib.loc=${base-maven.loc}/biz/aQute/bnd/biz.aQute.bndlib/${bnd.version}/biz.aQute.bndlib-${bnd.version}.jar
+test.entry=org.apache.jasper.TestJspC
```
markt-asf: This shouldn't be here.

ml86: Sorry i missed removing that part. Fixed.

markt-asf:
I've merged this along with a changelog entry. Many thanks for the PR.

PR #233, opened 2020-01-10
if TRACE is not allowed skip error page for such requests

panchenko:
I have a Spring Boot application, and Spring Boot by default configures an erorr page.
The TTACE method is not alloed (default configuration).

When a TRACE request happens it tries to return 405, which is then sent to the error page which is not expecting to handle it and finally request is served by `HttpServlet.doTrace()`, response status is 405, but headers are returned back.

I think it should be fixed in Tomcat - if TRACE is not allowed, then application should never receive such requests.

Please see the suggested change and tests.

### DIFF
--- a/java/org/apache/catalina/core/StandardHostValve.java
+++ b/java/org/apache/catalina/core/StandardHostValve.java
@@ -222,6 +222,13 @@ final class StandardHostValve extends ValveBase {
             return;
         }
 
+        /* Do not look for error pages if TRACE request and it's not allowed */
+        if (statusCode == HttpServletResponse.SC_METHOD_NOT_ALLOWED
+                && "TRACE".equals(request.getMethod())
+                && !request.getConnector().getAllowTrace()) {
+            return;
+        }
+
         ErrorPage errorPage = context.findErrorPage(statusCode);
         if (errorPage == null) {
             // Look for a default error page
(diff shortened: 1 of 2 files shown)


markt-asf:
The Servlet specification requires that applications are given the opportunity to handle errors. There are no exceptions to this. Further, the Servlet specification requires that the original request and response are forwarded to the error page.
It is the responsibility of any application error handler to ensure that it correctly handles a forwarded request with a method other than GET.
This is an application version of CVE-2017-5664 and needs to be handled in the application.

panchenko:
Surely I understand what you mean, but the Servlet specification says nothing about `Connector.getAllowTrace()`, which is checked in `CoyoteAdapter`. 

```java
        // Filter trace method
        if (!connector.getAllowTrace()
                && req.method().equalsIgnoreCase("TRACE")) {
...
            response.sendError(405, "TRACE method is not allowed");
            // Safe to skip the remainder of this method.
            return true;
        }
```

I would say this happens **before** entering into the area covered by the Servlet specification.
If TRACE is not allowed, then the most obvious behaviour would be not leaking it into an error page.
IMHO, If application is interested in handling TRACE requests - it shoud configure Connector accordingly.

Some puople consider the enabled TYACE method as a security vulrerability. Currently in Tomcat it becomes unexpectedly exposed via an error page path.

I agree that check I added to skip the error page lookup does not directly reference this condition from CoyoteAdapter and that causes confusion. I guess it can be improved.

PR #235, opened 2020-01-16
tomcat-jdbc check if returned connection is closed

panchenko:
I had a situation when driver closed connection because of I/O error, bur connection has been added to the avaialble ones and then returned to another caller.
The `testOn*` properties were not enabled in that case.

It looks like the pool can easily check if connection has been closed and don't accept it back.
WDYT?

### DIFF
--- a/modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/ConnectionPool.java
+++ b/modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/ConnectionPool.java
@@ -911,6 +911,12 @@ public class ConnectionPool {
     protected boolean shouldClose(PooledConnection con, int action) {
         if (con.getConnectionVersion() < getPoolVersion()) return true;
         if (con.isDiscarded()) return true;
+        try {
+            if (con.isClosed()) return true;
+        } catch (SQLException e) {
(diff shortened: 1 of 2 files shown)


panchenko:
The test checks that connection obtained from the pooll second time can be used for query execution.
Without this pool code change it returns the closed connection and query exection fails.

zyu-godaddy:
Also reported: https://bz.apache.org/bugzilla/show_bug.cgi?id=66502

This bug is the only bug that has been bugging us; otherwise the pool works flawlessly.

I wonder if you people want to revisit this issue; or I can offer to work on it.

markt-asf:
If you can address Chris's concerns with the test then I don't see why this PR (or one based on it) could not be merged.

zyu-godaddy:
~Cool, I will.~

zyu-godaddy:
While this change makes sense to me, and it should be inexpensive for the jdbc driver I am using (mysql-connector-java), I don't feel comfortable to add this check unconditionally without knowing the full impact. So I am not going to pursue this approach.

Instead, I will try `setDiscarded(true)` in an interceptor conditional on failed query and isClosed.

panchenko:
@markt-asf @ChristopherSchultz Addressed the comments about test, please take a look

markt-asf:
Merged manually and added change log entry.

PR #240, opened 2020-02-13

ThStock:


java/org/apache/catalina/ha/session/DeltaSession.java
```diff
@@ -106,7 +106,15 @@ public DeltaSession(Manager manager) {
         super(manager);
         boolean recordAllActions = manager instanceof ClusterManagerBase &&
                 ((ClusterManagerBase)manager).isRecordAllActions();
-        deltaRequest = new DeltaRequest(getIdInternal(), recordAllActions);
+        deltaRequest = newDeltaRequest(getIdInternal(), recordAllActions);
+    }
+
+    protected DeltaRequest newDeltaRequest() {
+        return new DeltaRequest();
```
ChristopherSchultz: I might do:

`return newDeltaRequest(null, false);`

here instead of calling the constructor. This allows subclasses to simply override a single method. Or maybe just use the single method and pass-in `null/false` instead of having an overloaded no-arg method.

ThStock: What do you think, is the latest change acceptable?

PR #241, opened 2020-02-14
Increment ReleaseIdle Counter when testAllIdle releases them

chaitan64arun:
In Tomcat JDBC module, I reckon PoolCleaner cleans the idle connections in two ways

- checkIdle
- testAllIdle

`releasedIdleCount` is incremented when released from `checkIdle` but not from `testAllIdle`. Is it part of the design? If so, why is it not counted?

IMO, the counter should be incremented in `testAllIdle` too. This PR includes this change.


aooohan:
Ok, you are right. It does seem to have been overlooked.

aooohan (approved):

PR #243, opened 2020-02-16
Improved CloudMembershipService documentation and configuration

bohmber:


java/org/apache/catalina/tribes/membership/cloud/DNSMembershipProvider.java
```diff
@@ -30,6 +30,61 @@
...
+ *         </Cluster>
+ *         ...
+ *  }
+ *  </pre>
+ *
+ * {@code dns-membership-service.yml }
+ *
+ * <pre>
+ * {@code
+ * apiVersion: v1
```
martin-g: This is Kubernetes config. Why it is in DNSMembershipProvider ?

martin-g: At line 150 it uses Kubernetes logging message too.

rmaucher: One uses rhe Kubernetes API and parses the JSON, the other one is "lightweight" and does the same with a DNS hack request. Personally, I have found the DNS technique wasn't so reliable, so I don't like it. But this is "recommended" as the proper way (access to the REST API) needs security tuning.
At line 150, it uses Kubernetes logging message because the two do the same thing and run into the same error. If you would want a separate key "because", feel free to "fix" it.

bohmber:
DnsMembershipProvider is need if you are in the cloud and you don't have kubernetes api access and multicast is not available. Therefor I added an example how to configure DNSMembershipProvider in kubernetes. In other env you have 

Here is an example from Jean-Frederic Clere
https://github.com/jfclere/tomcat-openshift/blob/master/conf/server.xml

DNSMembershipProvider is inspired by this i guess: 
http://jgroups.org/manual4/index.html#_dns_ping

Please read the first few lines

java/org/apache/catalina/tribes/membership/cloud/DNSMembershipProvider.java
martin-g: I am confused. 
`"One uses rhe Kubernetes API..."` - is this DNSMembershipProvider ?
`...the other one is "lightweight" and does the same with a DNS hack request` - is this KubernetesMembershipProvider ?
If this is the case then IMO those classes should swap their names.

rmaucher: DNSMembershipProvider does the DNS hack request, and KubernetesMembershipProvider is the one using the Kubernetes REST API.

PR #244, opened 2020-02-16
The '$' in the class name of Digester$EnvironmentPropertySource is no…

bohmber:
…t really shell frendly for unix environments. The class should not a inner class.

### DIFF
--- a/java/org/apache/tomcat/util/digester/EnvironmentPropertySource.java
+++ b/java/org/apache/tomcat/util/digester/EnvironmentPropertySource.java
@@ -1,0 +1,41 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.tomcat.util.digester;
+
+import java.security.Permission;
+
+import org.apache.tomcat.util.IntrospectionUtils;
+import org.apache.tomcat.util.security.PermissionCheck;
+
+public class EnvironmentPropertySource implements IntrospectionUtils.SecurePropertySource {
+
+    @Override
+    public String getProperty(String key) {
+        return null;
+    }
+
+    @Override
+    public String getProperty(String key, ClassLoader classLoader) {
+        if (classLoader instanceof PermissionCheck) {
+            Permission p = new RuntimePermission("getenv." + key, null);
+            if (!((PermissionCheck) classLoader).check(p)) {
+                return null;
+            }
+        }
+        return System.getenv(key);
+    }
+}
--- a/java/org/apache/tomcat/util/digester/EnvironmentPropertySource.java
+++ b/java/org/apache/tomcat/util/digester/EnvironmentPropertySource.java
@@ -21,6 +21,41 @@ import java.security.Permission;
 import org.apache.tomcat.util.IntrospectionUtils;
 import org.apache.tomcat.util.security.PermissionCheck;
 
+/**
+ * A {@link org.apache.tomcat.util.IntrospectionUtils.SecurePropertySource} that uses environment variables to resolve expressions.
+ *
+ * <p><strong>Usage example:</strong></p>
+ *
+ * Configure the certificate with environment variables.
+ *
+ * <pre>
+ *   {@code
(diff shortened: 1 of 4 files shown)


martin-g:
Just curious: why does it have to be Shell friendly ?

bohmber:
If you set it via CATALINA_OPTS is almost impossible to set this value.

martin-g:
I haven't needed to set this particular class but usually one just needs to quote such values.

bohmber:
> I haven't needed to set this particular class but usually one just needs to quote such values.

Did you try to quote it? I don't know how many \\\\\\ are needed until it's working this is not reliable. Therefore a '$' should be avoided if you are using a shell.

martin-g:
I meant quote it, e.g. `-Dsome.key="Some$Value"`, not _escape_ it.

bohmber:
The only reliable way was to add this to catalina.properties. Please test it yourself. I still belief it's a really bad idea to use '$' in a classname that is configurable via -D command line parameter.

martin-g:
I am not against using a proper class! I'm just trying to understand what kind of problem you have faced.

java/org/apache/tomcat/util/digester/EnvironmentPropertySource.java
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+package org.apache.tomcat.util.digester;
+
+import java.security.Permission;
+
+import org.apache.tomcat.util.IntrospectionUtils;
+import org.apache.tomcat.util.security.PermissionCheck;
+
+public class EnvironmentPropertySource implements IntrospectionUtils.SecurePropertySource {
```
ChristopherSchultz: Since you are doing a bit of re-factoring, could you add some javadoc? Your use of the class suggests that you understand its use and the need for its existence. No need for anything fancy. Just a one or two sentence description and maybe a reference to the users manual if appropriate.

bohmber: Super idea will do it. Give me time til Friday.

bohmber: Can you take a look please

java/org/apache/tomcat/util/digester/EnvironmentPropertySource.java
```diff
@@ -0,0 +1,76 @@
...
+ *                        certificateFile="${CERTIFICATE_FILE}"
+ *                        certificateChainFile="${CERTIFICATE_CHAIN_FILE}"
+ *                        type="RSA" />
+ *     </SSLHostConfig> }
+ * </pre>
+ *
+ * How to configure:
+ * <pre>
+ * {@code
+ *   echo "org.apache.tomcat.util.digester.EnvironmentPropertySource=org.apache.tomcat.util.digester.EnvironmentPropertySource" >> conf/catalina.properties}
```
martin-g: Are you sure that this is the right key ?
Looking at org.apache.tomcat.util.digester.Digester I see:
```
String className = System.getProperty("org.apache.tomcat.util.digester.PROPERTY_SOURCE");
```

martin-g:
If we are going to make this change then it makes sense to do it for `org.apache.tomcat.util.digester.Digester.SystemPropertySource` as well.

java/org/apache/tomcat/util/digester/EnvironmentPropertySource.java
bohmber: good catch fixed

bohmber:
> If we are going to make this change then it makes sense to do it for `org.apache.tomcat.util.digester.Digester.SystemPropertySource` as well.

That is the default implementation and it's always configured. It's a private class the change there is not needed.

webapps/docs/changelog.xml
```diff
@@ -92,6 +92,10 @@
       <fix>
         Missing store config attributes for Resources elements. (remm)
       </fix>
+      <scode>
+        Rename <code>org.apache.tomcat.util.digester.Digester$EnvironmentPropertySource</code> to
+        <code>org.apache.tomcat.util.digester.EnvironmentPropertySource</code> old class is still available.
```
martin-g: Punctuation is missing before ` old class is ...`

java/org/apache/tomcat/util/digester/EnvironmentPropertySource.java
```diff
@@ -0,0 +1,76 @@
...
+ *   {@code
+ *     <SSLHostConfig>
+ *           <Certificate certificateKeyFile="${CERTIFICATE_KEY_FILE}"
+ *                        certificateFile="${CERTIFICATE_FILE}"
+ *                        certificateChainFile="${CERTIFICATE_CHAIN_FILE}"
+ *                        type="RSA" />
+ *     </SSLHostConfig> }
+ * </pre>
+ *
+ * How to configure:
```
martin-g: Both samples to setup the PROPERTY_SOURCE below do not use environment variable. IMO both of them would be picked up by SystemPropertySource, not by EnvironmentPropertySource.

bohmber:
Ping

hamza86:
I was trying to use this feature on Tomcat 9.x and from my experience, I found it briefly documented. This feature is really important to run legacy apps on Kubernetes since most of the configurations are coming from config maps. 
What I would do, is to create one class that read properties form system properties and environment properties. Ofc a priority mechanism must be introduced to avoid overlapping/overriding properties.

PR #247, opened 2020-02-21
Tomcat Session Manager w/ Authentication Persistence

cklein05:
This enhancement adds a new attribute `persistAuthentication` to both `StandardManager` and `PersistentManager`. When set to `true`, sessions maintained by that manager do as well persist authentication information (if any) along with the session's base data and attributes, that is, fields `authType` and `principal` of the `StandardSession` class. Effectively, this adds _Authentication Persistence_ to Tomcat's _Session Persistence_ feature.

As a result, sessions remain authenticated when being restored from disk after an application restart (with `StandardManager`) or when (for whatever reason) being _swapped in_ by `PersistentManager`. That could be quite useful when a session-based authentication mechanism is used (e.g. FORM authentication). Since an application must be restarted in order to apply configuration changes, preserving authentication during that time may help not to bother users by constantly requiring to re-enter credentials.

This enhancement is primarily intended for smaller installations, for which the more powerful clustering feature with session and authentication replication seems being overdrawn. Indeed, this enhancement does not alter anything in, or interact with Tomcat's _High Availability_ capabilities (Clustering).

Since storing the sessions's principal do disk may be considered a security risk in some cases, this new attribute defaults to `false`, which completely disables authentication persistence.

This enhancement has already been discussed quite intensively on the tomcat-users mailing list: ['Enhancement: New option 'persistAuthentication' for session manager'](https://www.mail-archive.com/users@tomcat.apache.org/msg134167.html).


markt-asf:
Fixed in:
- master for 10.0.0-M2 onwards
- 9.0.x for 9.0.32 onwards
- 8.5.x for 8.5.52 onwards
- 7.0.x for 7.0.101 onwards

PR #249, opened 2020-03-02
Support for default values in property value expressions

bohmber:
It would be nice to have support for default values in property value expressions.
You can use it with ${key:default}. 
Any interest in this feature?
Where is the best place for document this? 
Where to add a junit test for this? 
I used a modified version of 
org.apache.catalina.startup.TestTomcatStandalone on my local env.

### DIFF
--- a/java/org/apache/tomcat/util/IntrospectionUtils.java
+++ b/java/org/apache/tomcat/util/IntrospectionUtils.java
@@ -280,7 +280,8 @@ public final class IntrospectionUtils {
         return sb.toString();
     }
 
-    private static String getProperty(String name, Hashtable<Object, Object> staticProp, PropertySource[] dynamicProp, ClassLoader classLoader) {
+    private static String getProperty(String name, Hashtable<Object, Object> staticProp,
+            PropertySource[] dynamicProp, ClassLoader classLoader) {
         String v = null;
         if (staticProp != null) {
             v = (String) staticProp.get(name);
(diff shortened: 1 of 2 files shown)


markt-asf:
Hmm. That makes ":" a reserved character in property values. That is going to have an impact on backwards compatibility. ":" is likely to be used for the address attribute for IPv6 addresses. I think it is going to be tricky to find a suitable character to use as a delimiter here.

bohmber:
I try to resolve the property with ":" first. Can you give me some examples of what goes wrong. I don't want i introduce something that is not backwards compatible.

markt-asf:
Of course, this is the name, not the value and if the name with ":" is tried first that should be fine.

martin-g:
The change looks good to me but it would be nice to have some new unit tests for it.

bohmber:
Unit test would be good. Did you have a good idea for a starting point for an unit test? I was playing with `TestTomcatStandalone`.

martin-g:
No need to start Tomcat. Just plain old unit test.

bohmber:
Added a simple unit test.

martin-g:
> Added a simple unit test.

Awesome!

Just one more thing: please update the javadoc of the `public` method that it supports default value.

bohmber:
Ping. Anything else needed for this pr?

markt-asf:
Nothing comes to mind immediately. We typically go through all the open PRs before we start the release process. We aim to start the release process at the beginning of the month so we'll be going through the PRs in the next day or so.

bohmber:
Thanks that is fine for me

markt-asf:
Thanks for thr PR. A variation has been applied to all currently supported branches.

PR #252, opened 2020-03-04
8.5.x

lkirchev:
Provide means for the cookie header generation to make decisions based on properties of the request.

### DIFF
--- a/java/org/apache/catalina/connector/Response.java
+++ b/java/org/apache/catalina/connector/Response.java
@@ -974,11 +974,11 @@ public class Response implements HttpServletResponse {
             return AccessController.doPrivileged(new PrivilegedAction<String>() {
                 @Override
                 public String run(){
-                    return getContext().getCookieProcessor().generateHeader(request, cookie);
+                    return getContext().getCookieProcessor().generateHeader(request.getRequest(), cookie);
                 }
             });
         } else {
-            return getContext().getCookieProcessor().generateHeader(request, cookie);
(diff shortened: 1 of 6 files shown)


markt-asf:
I can see why this is necessary and I support making the necessary changes to Tomcat to support solving this problem. The patch as proposed cannot be applied as the o.a.t.util package should not have a dependency on o.a.catalina. It should be possible to use the HttpServletRequest instead. I'll take a look at a variation on this patch for master and then back-port.

lkirchev:
Thanks a lot Mark! Yes, you are right, I overlooked the dependency I introduced. I added another commit with a proposal for replacing o.a.c.c.Request with HttpServletRequest. 
Only to mention that in the TestCookieProcessorGeneration the usage of the newly introduced dummy implementation of HttpServletRequest could be replace just with nulls - this reference is never used in the two implementations of CookieProcessor in Tomcat. 
However, I didn't like very much the idea to use nulls, that's why I introduced the dummy implementation.

markt-asf:
Thanks. While I'm not merging this specific PR I do want to assure you that the PR was very helpful in resolving this. Having a concrete proposal to work from is much easier than developing one from scratch. The end result isn't that different from what you proposed. I opted for `null` rather than dummy request objects as it simplified things a little. The only other difference was the changes to handle the deprecation so we end up with a single `generateHeader()` in Tomcat 10 onwards.

Thanks again for your contribution.

lkirchev:
Thanks a lot Mark! I didn't expect to merge the PR when I added the last commit, I just wanted to bring it most closely to a state you can reuse.

PR #253, opened 2020-03-05
add warning to upgrade

YYTVicky:


markt-asf:
Adding comments via a PR is not the way to register an enhancement request. Please see http://tomcat.apache.org/bugreport.html

PR #256, opened 2020-03-06

ThStock:
See https://github.com/apache/tomcat/pull/240

### DIFF
--- a/java/org/apache/catalina/ha/session/DeltaSession.java
+++ b/java/org/apache/catalina/ha/session/DeltaSession.java
@@ -106,7 +106,15 @@ public class DeltaSession extends StandardSession implements Externalizable,Clus
         super(manager);
         boolean recordAllActions = manager instanceof ClusterManagerBase &&
                 ((ClusterManagerBase)manager).isRecordAllActions();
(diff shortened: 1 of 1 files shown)


rmaucher:
I think merging #240 was a bad idea: we are not going to add extension points for just about anything, which are then supposed to be set in stone while looking rather useless to just about everyone and in great danger of being refactored away.
-1 for adding this.

ChristopherSchultz:
I think this is the same as #240; I just haven't back-ported it.

Can we discuss on #240 what your objections are? The alternative is for a user to re-implement 100% of e.g. DeltaManager just to change one aspect of it's behavior.

rmaucher:
The change implies all methods of concrete classes become supported stable APIs. I think this is too much and we cannot / should not promise that the added methods are not going to be refactored away.
Also, as I pointed out, there's pretty much an infinity of convenient hooks that can be added everywhere for extensibility.
When extending into custom components, I don't think some amount of code duplication is that bad. Sure, you may miss bugfixes but it's more likely to remain stable that way. And when things work and you have no problems, you don't need magic bugfixes and random behavior changes IMO.

PR #257, opened 2020-03-06

ThStock:
https://github.com/apache/tomcat/pull/240


markt-asf:
Thanks for the additional PR. I've back-ported the original fix to 9.0.x, 8.5.x and 7.0.x along with a comment that explains why the code has been added.

PR #260, opened 2020-03-12
remove top '/' from parameter

touchez-du-bois:
Remove top "/" from paramter of URLClassLoader#findResources().
I think that bad "/META-INF/services/...", good "META-INF/services/...".


rmaucher:
Not needed, see commit c9c52ba3ec3f07f019dbd246574088fb07891d24

PR #262, opened 2020-03-13
Fix and tests for tomcat bug #64226

gazzyt:
https://bz.apache.org/bugzilla/show_bug.cgi?id=64226

### DIFF
--- a/java/org/apache/tomcat/util/http/ConcurrentDateFormat.java
+++ b/java/org/apache/tomcat/util/http/ConcurrentDateFormat.java
@@ -34,33 +34,34 @@ public class ConcurrentDateFormat {
     private final String format;
     private final Locale locale;
     private final TimeZone timezone;
-    private final Queue<SimpleDateFormat> queue = new ConcurrentLinkedQueue<>();
+    private final Queue<SimpleDateFormat> formatQueue = new ConcurrentLinkedQueue<>();
+    private final Queue<SimpleDateFormat> parseQueue = new ConcurrentLinkedQueue<>();
 
     public ConcurrentDateFormat(String format, Locale locale, TimeZone timezone) {
         this.format = format;
         this.locale = locale;
         this.timezone = timezone;
-        SimpleDateFormat initial = createInstance();
-        queue.add(initial);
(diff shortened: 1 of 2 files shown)


martin-g:
IMO for the branches where Java 8 is minimum we should switch to Java 8 DateTime APIs. 
java.time.format.DateTimeFormatter is thread-safe.

gazzyt:
> IMO for the branches where Java 8 is minimum we should switch to Java 8 DateTime APIs.
> java.time.format.DateTimeFormatter is thread-safe.

Would you think to do this now or would that switch be better left to Tomcat 10 branch? There is some legacy stuff in FastHttpDateFormat marked for removal in Tomcat 10.

I don't mind having a go at this but it is a bigger change than the fix I submitted and there are no unit tests to fall back on.

gazzyt:
Can someone fix the CI please? The error isn't related to my code:

> The command "wget -q $JDK_URL && tar xzf OpenJDK*.tar.gz" failed and exited with 4 during .

martin-g:
> Can someone fix the CI please? The error isn't related to my code:

I've restarted it. It runs fine so far!

martin-g:
> Would you think to do this now or would that switch be better left to Tomcat 10 branch?

Let's see what the other devs think about this idea first. I could do it in a separate PR if there is interest.

ChristopherSchultz:
Please don't change something like this in Tomcat 9.0 or older. Those releases are far too stable to go around changing the date-related APIs.

gazzyt:
I did some hacking on the Java 8 `DateTimeFormatter` yesterday and concur with @ChristopherSchultz. I found that `DateTimeFormatter` parsing was more strict that `SimpleDateFormat` for the same format string and would throw exceptions where `SimpleDateFormat` returned success. These differences can be overcome - `DateTimeFormatter` does give you a lot more control but the risk of regression is there. I'll leave this code at https://github.com/gazzyt/tomcat/commits/java8-datetimeformatter in case it is useful to anyone.

rmaucher:
I preferred a simple reset of the timezone for now. Thanks for the PR.

PR #263, opened 2020-03-17
graal support for embedded jars

fhanik:
Don't merge yet, opening PR for purpose of review.

I will bring this up on the mailing list.

_include Graal files so that the native-image compiler picks it up
automatically_


rmaucher:
It doesn't seem bad to do that given how these jars are used.
Note: The resource bundles are now traced with Graal 20.0 so it's not 100% mandatory to have them (but I left that in the other resources because that was already there and it's documented to work with 19.3).

fhanik:
Close in favor of https://github.com/apache/tomcat/pull/274

PR #264, opened 2020-03-18
fix regression: only skip jar-scanning completely if scanSet is empty

Iridias:
The commit 21d08e1952d6612da9a712a173c81174958f4b8b introduced a regression to the behavior of jar-scanning. Since version 9.0.31 it skips all jars, if a wildcard is found in the skipSet - no matter the content of the scanSet!

The documentation for `jarsToScan` clearly states: 
> overrides the default jarsToSkip list above

This PR restores this behavior by also respecting the "jarsToScan".

### DIFF
--- a/java/org/apache/tomcat/util/scan/StandardJarScanFilter.java
+++ b/java/org/apache/tomcat/util/scan/StandardJarScanFilter.java
@@ -260,7 +260,7 @@ public class StandardJarScanFilter implements JarScanFilter {
     }
 
     private static boolean determineSkipAll() {
-       return defaultScanSet.isEmpty() &&
+        return defaultScanSet.isEmpty() &&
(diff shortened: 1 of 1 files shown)


martin-g:
Fixed with https://github.com/apache/tomcat/commit/9c6563bf6ac723cda4e78d0e8c1a996fa1b8ce56

Iridias:
> Fixed with 9c6563b

Hm, that would ignore the case, that `jarsToScan` is specified but empty. (Because then `defaultScan` would be an empty String instead of null)
Is that intended?

rmaucher:
I hadn't seen this PR at all as there was a BZ there already and it didn't mention it.
I will make the extra change (after all this is a stupid configuration, so it doesn't hurt to handle an even more stupid configuration) and will also give you credit since your patch was correct.

PR #265, opened 2020-03-31
Replace LOGGING_CONFIG BY CATALINA_LOGGING_CONFIG

netbeansuser2019:
To avoid collision e.g. with spring framework like spring-projects/spring-boot#1432

### DIFF
--- a/bin/catalina.bat
+++ b/bin/catalina.bat
@@ -94,9 +94,9 @@ rem   JSSE_OPTS       (Optional) Java runtime options used to control the TLS
 rem                   implementation when JSSE is used. Default is:
 rem                   "-Djdk.tls.ephemeralDHKeySize=2048"
 rem
-rem   LOGGING_CONFIG  (Optional) Override Tomcat's logging config file
+rem   CATALINA_LOGGING_CONFIG  (Optional) Override Tomcat's logging config file
 rem                   Example (all one line)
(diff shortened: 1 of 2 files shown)


markt-asf:
In what circumstances is this going to clash with Spring Boot?
My initial reaction to to reject this change as it is going to break existing configurations.

ChristopherSchultz:
Also, `LOGGING_CONFIG` is just an environment variable containing `-Dfoo=bar` arguments to the JVM. Merging `LOGGING_CONFIG` necessary for e.g. Spring with whatever you need for Tomcat should be easy. There should be no need to change anything, here.

martin-g:
> Merging `LOGGING_CONFIG` ... 

merge it how ?

Spring Boot reads this env variable to find the Logback/Log4j2 config file, e.g. `classpath:logback.xml` for Logback. Its value is not as the one Tomcat expects.

ChristopherSchultz:
Spring Boot is only configured via environment variables? Seems fragile and non-portable to me.

martin-g:
Spring Boot could be configured via config file(s), system properties and/or env variables. It is pretty flexible!
Using env variables is the preferred way in Docker/Kubernetes setups.

markt-asf:
Given containers seem to like environment variables, I don't see a way around this. I'm working on a backwards compatible solution.

markt-asf:
Fixed in a backwards compatible manner for all currently supported versions.

PR #266, opened 2020-04-04
Fix typo in exception message about quote character in repository path

biya-bi:
This is a very small commit that corrects a typo in the message of the
exception thrown when a quote character is used in path representing a
repository.


markt-asf:
Thanks for the PR. I'll get it back-ported to the other supported branches.

biya-bi:
All the pleasure is mine Mark!

PR #267, opened 2020-04-06
Remove reflection from Connector.java

fhanik:
I will start a discussion on dev@tomcat.apache.org

TL;DR; We are using Apache Tomcat embedded in native images (GraalVM). 
Reflection causes images to be unnecessarily large, and Apache Tomcat has many places where reflection can be avoided in an embedded scenario.


rmaucher:
The main target of Tomcat remains Tomcat standalone, server.xml and the digester. So this means tons of reflection here and there, and I expect it will stay that way.
You're in luck here since the Connector is a mess and some things could be improved. You need to continue working on the PR though, as the code changed to be less messy already.

fhanik:
AbstractProtocol and AbstractEndpoint use reflection for all property setters. This change is a step in the right direction, but only has minimal impact.

fhanik:
Closed per discussion on mailing list

PR #269, opened 2020-04-07
Use parameterized collection constructors where possible

larsgrefer:
This allows collections like ArrayList and HashSet to initialize their backing arrays
with the correct size.

### DIFF
--- a/java/jakarta/el/LambdaExpression.java
+++ b/java/jakarta/el/LambdaExpression.java
@@ -65,8 +65,7 @@ public class LambdaExpression {
         // Build the argument map
         // Start with the arguments from any outer expressions so if there is
         // any overlap the local arguments have priority
-        Map<String,Object> lambdaArguments = new HashMap<>();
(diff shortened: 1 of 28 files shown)


ChristopherSchultz:
Many of these instances are to protect private collections from being modified. For example, in `ServletSecurityElement.getHttpMethodConstraints()`, a new collection is being returned to avoid having the internal collection modified. Would it be safe-enough to return e.g. `Collections.unmodifiableSet(methodConstraints.values())` from that method, or is that too dangerous, as the "unmodifiable" collection includes a reference to the actual, potentially modifiable collection?

markt-asf:
`Collections.unmodifiable[Set|Map|List]` is probably an option in some cases but some (most?) of these instances are returned as part of a public API and if the API doesn't state they are modifiable then we can't use this approach. Probably a discussion for the dev@ list.
The approach used in the PR is sound. I just want to review the full diff and then I'll merge it.

PR #275, opened 2020-04-18
BZ 59203 - interrupt tomcat threads instead of stopping

govi20:
interrupt tomcat threads instead of stopping. It seems existing test cases are already checks whether the ThreadPoolExecutor is terminated.

### DIFF
--- a/java/org/apache/catalina/loader/WebappClassLoaderBase.java
+++ b/java/org/apache/catalina/loader/WebappClassLoaderBase.java
@@ -1813,7 +1813,7 @@ public abstract class WebappClassLoaderBase extends URLClassLoader
                         // stopped and check them at the end of the method.
                         executorThreadsToStop.add(thread);
                     } else {
-                        clearThread(thread);
+                        clearThread(thread, usingExecutor);
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


markt-asf:
Thanks for the PR.

There are multiple different changes in this commit. It needs to be broken down into one commit for the changed described in the PR summary and one commit for each type of clean-up.

The configuration option that controls this behaviour is called `clearReferencesStopThreads`. That makes no sense with this PR applied. However, the behaviour of stopping threads needs to be retained.

Note there are multiple places where stop() is called. This PR addresses only one of them.

It would be better to align the behaviour for Executor and non-Executor threads. i.e.:

- interrupt
- pause
- stop if still running

There was some debate whether there should be multiple iterations if interrupt/pause but I'm leaning towards not. We are dealing with broken web applications here.

govi20:
I will remove the cleanup changes. 
So apparently we should call interrupt, check if thread is still active, if yes then call stop. Correct me if I'm wrong here.

java/org/apache/catalina/loader/WebappClassLoaderBase.java
```diff
@@ -1813,35 +1813,40 @@ private void clearReferencesThreads() {
...
+        if (!t.isInterrupted()) {
+            t.interrupt();
+        }
+
+        // Give threads up to 2 seconds to shutdown
+        while (t.isAlive() && count < 100) {
+            try {
+                Thread.sleep(20);
+            } catch (InterruptedException e) {
+                // Quit the while loop
```
pzygielo, line 1839: https://rules.sonarsource.com/java/RSPEC-2142

markt-asf:
Thanks for the updates. Exactly the sort of thing I had in mind.
Looking at the updated code, I'm wondering whether a slightly wider refactoring makes sense. Something like:

- for each thread where the web application class loader is the context class loader (this code is already present)
    - if using executor, stop the executor (which should interrupt the thread) else interrupt the thread
    - add the thread to a list of threads that should be stopping
- wait until all the threads that should be stopping have stopped or a set amount of time passes (currently we use 2 seconds)
- call stop on any threads still running

i.e. interrupt all threads, wait for all threads, stop all remaining threads rather than interrupt, wait, stop one thread at a time.

govi20:
Looks like the first two steps are already covered in the code. Small refactoring would be to ensure interrupt is not called on executor thread.

```
if (!usingExecutor && !t.isInterrupted()) {
       t.interrupt();
}
```

correct me if I am not on a same page.

govi20:
@markt-asf could you please suggest if any further changes are required here.

markt-asf:
Thanks. I used a variation of your PR to fix this.

PR #276, opened 2020-04-20
Add s390x Travis support

prankkelkar:
As Travis CI [officially supports](https://blog.travis-ci.com/2019-11-12-multi-cpu-architecture-ibm-power-ibm-z) s390x builds, adding support for same.

Changes:
* Added s390x to build matrix
* Removed explicit steps to download java as they were redundant since using jdk tag from travis already provides expected java in the environment.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -36,7 +36,15 @@ addons:
         - libtool-bin
 
 install:
-    - export JAVA_HOME=$(dirname $(dirname $(readlink -f $(which javac))))
+    - ARCH=`uname -p`
+    - echo $ARCH
+    - JDK_X64="https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u242-b08/OpenJDK8U-jdk_x64_linux_hotspot_8u242b08.tar.gz"
+    - JDK_ARM64="https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u232-b09/OpenJDK8U-jdk_aarch64_linux_hotspot_8u232b09.tar.gz"
+    - JDK_s390x="https://github.com/AdoptOpenJDK/openjdk11-binaries/releases/download/jdk-11.0.5%2B10/OpenJDK11U-jdk_s390x_linux_hotspot_11.0.5_10.tar.gz"
+    - if test "X$ARCH" = "Xaarch64"; then JDK_URL=$JDK_ARM64; elif test "$ARCH" = "s390x"; then JDK_URL=$JDK_s390x; else JDK_URL=$JDK_X64; fi
+    - wget -q $JDK_URL && tar xzf OpenJDK*.tar.gz
+    - if test "$ARCH" = "s390x"; then mv jdk-11* jdk; else mv jdk8* jdk; fi
+    - export JAVA_HOME=`pwd`/jdk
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -32,14 +36,7 @@ addons:
...
 
 install:
-    - ARCH=`uname -p`
-    - echo $ARCH
-    - JDK_X64="https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u242-b08/OpenJDK8U-jdk_x64_linux_hotspot_8u242b08.tar.gz"
-    - JDK_ARM64="https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u232-b09/OpenJDK8U-jdk_aarch64_linux_hotspot_8u232b09.tar.gz"
-    - if test "X$ARCH" = "Xaarch64"; then JDK_URL=$JDK_ARM64; else JDK_URL=$JDK_X64; fi
-    - wget -q $JDK_URL && tar xzf OpenJDK*.tar.gz
-    - mv jdk8* jdk
-    - export JAVA_HOME=`pwd`/jdk
```
martin-g, line 42: The above was here for two reasons:
1) because TravisCI is really inconsistent in what is available in the different types of VMs
1.1) JAVA_HOME is exported for AMD64, but not for ARM64
1.2) `mvn` is available for AMD64, but not for ARM
1.3.) most important - to keep the version of the used JDK the same for all environments. The one provided by Travis for ARM64 was older than the one for AMD64

For the above reasons I'd prefer to keep the control

.travis.yml
```diff
@@ -15,8 +15,12 @@
 
 dist: bionic
 language: java
-jdk: oraclejdk8
-arch: arm64
+jobs:
+    include:
+      - arch: s390x
+        jdk: openjdk11
```
martin-g, line 21: any reason to use JDK 11 for s390x ?

.travis.yml
prankkelkar: I had shifted to use travis jdk to make things cleaner. Also since the version was exactly same i thought the steps were redundant. I will revert since you want keep it the way it is.

.travis.yml
prankkelkar: JDK8 is very slow for s390x and causes test case failures due to timeout.

martin-g: My concern is that Tomcat (even version 10.x) is supposed to be build with JDK 8. 
But since Travis is not the main CI I think it is OK to use JDK 11 as an exception here.

rainerjung:
Hi Martin,

the build.xml line numbers

/home/travis/build/apache/tomcat/build.xml:2789: The following error 
occurred while executing this line:

/home/travis/build/apache/tomcat/build.xml:3124: The following error 
occurred while executing this line:

/home/travis/build/apache/tomcat/build.xml:3240: 
javax.net.ssl.SSLHandshakeException: PKIX path validation failed:

point at the commons-daemon download via an Apache mirror. So I guess 
one of the mirrors uses an outdated certificate? If so, the error will 
be intermittent, depending on which mirror is chosen.

Regards,

Rainer

Am 23.04.2020 um 15:44 schrieb Martin Grigorov:

martin-g:
Thanks for the explanation, Rainer!
I've restarted the job at Travis. Let's see!

martin-g:
Thank you, @prankkelkar !

PR #278, opened 2020-04-21
display build status on the readme file

govi20:


markt-asf:
This assumes Travis is the primary CI system for Tomcat. It isn't. I'm not against this information being added somewhere to the readme file but it needs to make clear Travis is not the primary CI and, as such, I don;t think it belongs at the top of the file.

govi20:
My assumption was Travis is the only CI. We can add status' at bottom just like apache/spark readme.

markt-asf:
For the full list:
http://tomcat.apache.org/ci.html

PR #279, opened 2020-04-21
Optimize concurrency

KangZhiDong:
more performance

### DIFF
--- a/java/org/apache/tomcat/util/net/AprEndpoint.java
+++ b/java/org/apache/tomcat/util/net/AprEndpoint.java
@@ -1770,7 +1770,7 @@ public class AprEndpoint extends AbstractEndpoint<Long,Long> implements SNICallB
                     maintainTime = 0;
                     try {
                         synchronized (this) {
-                            if (sendfileRunning && sendfileCount < 1 && addS.size() < 1) {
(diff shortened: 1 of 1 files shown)


markt-asf:
What is the basis for this change?
While the thread is in `wait()`, `notify()` should only be called if one of the conditions in the surrounding `if(...)` is no longer true. I'd expect the change in this PR to make performance marginally worse, not better.

KangZhiDong:
> What is the basis for this change?
> While the thread is in `wait()`, `notify()` should only be called if one of the conditions in the surrounding `if(...)` is no longer true. I'd expect the change in this PR to make performance marginally worse, not better.

When this thread is notify , it do not check "(sendfileRunning && sendfileCount < 1 && addS.size() < 1)" ,but "while" check again,avoid other thread  may change it

KangZhiDong:
> What is the basis for this change?
> While the thread is in `wait()`, `notify()` should only be called if one of the conditions in the surrounding `if(...)` is no longer true. I'd expect the change in this PR to make performance marginally worse, not better.

i think because "while (sendfileRunning && sendfileCount < 1 && addS.size() < 1)"  in outer layer check again,so it  can use "if" there ，do not need check again in synchronized block

markt-asf:
Your analysis is not correct.

PR #280, opened 2020-04-22
Remove unnecessary boxing and unboxing

larsgrefer:


markt-asf:
These are all deliberate to keep the IDE (Eclipse) clear of warnings. We could disable that warning but that then runs the risk of us missing one of the more subtle autoboxing/unboxing related bugs. Given that these conversions are going to happen anyway, I'd prefer for them to be explicit.

PR #281, opened 2020-04-23
Change FilterChain holder from array to list

govi20:
Instead of resizing array manually, we can leverage `List` to hold filter-chain

### DIFF
--- a/java/org/apache/catalina/core/ApplicationFilterChain.java
+++ b/java/org/apache/catalina/core/ApplicationFilterChain.java
@@ -19,6 +19,8 @@ package org.apache.catalina.core;
 import java.io.IOException;
 import java.security.Principal;
 import java.security.PrivilegedActionException;
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Set;
 
 import jakarta.servlet.Filter;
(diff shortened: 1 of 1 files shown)


rmaucher:
This commit seems functionally useless but is solely motivated by personal preferences. At the very least, they should be accompanied by quick perf numbers that show equivalence or better.

ChristopherSchultz:
There *is* some code-reduction, and therefore a reduction in the possibility of bugs. Obviously, this code is well-tested and does indeed work.

Another possible advantage is that the List can become unmodifiable at some point, which may possibly improve security and/or code-safety.

govi20:
There are so many places in the codebase where we can replace array with List could reduce good amount of code.

markt-asf:
The question is "At what cost?" I have two concerns.
1. Regressions. The changes should be obvious and easy to check but there are plenty of examples in the history of Tomcat of seemingly simple clean-up triggering unintended regressions. That isn't a reason not to do it but there needs to be a clear benefit - such as reduced lines of code - of making the change.
2. Performance. I have in the past seen very significant performance drops changing from a simple array to an object (byte[] to ByteBuffer is the case I am thinking of). We need to be very careful of changes that are on the critical path (get executed for all/most requests). In those cases I would want to see repeatable performance numbers that demonstrated the performance cost/benefit of making the change including the impact on garbage generation. `ApplicationFilterChain` falls into this category.

rmaucher:
When I look at https://github.com/apache/tomcat/pull/271 in comparison, it has an extensive "why" demonstrating why it is actually better and it is not simply cosmetics.

PR #283, opened 2020-04-23
Remove implicit unboxing

larsgrefer:


markt-asf:
Thank you. That prompted me to figure out why I got the warning in the text editor but not in the problems tab. I now have ~1,500 errors and warnings to look at. But I'll merge this first. At least it will be one less ;)

PR #284, opened 2020-04-23
Improve General IDE support

larsgrefer:
see also: https://github.com/apache/tomcat/pull/282#pullrequestreview-399324287

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+root = true
+
+[*]
+end_of_line = lf
+indent_style = space
+trim_trailing_whitespace = true
+
+[*.java]
+indent_size = 4
+max_line_length = 100
+ij_java_block_brace_style = end_of_line
+ij_java_else_on_new_line = false
+
+[*.{txt,md}]
+max_line_length = 80
+
+[*.xml]
+indent_size = 2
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,18 @@
+root = true
+
+[*]
+end_of_line = lf
+indent_style = space
+trim_trailing_whitespace = true
+
+[*.java]
+indent_size = 4
+max_line_length = 100
```
markt-asf, line 10: I think we have more to 120 for this.

.editorconfig
```diff
@@ -0,0 +1,18 @@
+root = true
```
markt-asf, line 1: There probably needs to be some discussion on tge dev@ list whether or not we want to start using this. My own first impressions are positive.

.editorconfig
```diff
@@ -0,0 +1,18 @@
...
+indent_size = 4
+max_line_length = 100
+ij_java_block_brace_style = end_of_line
+ij_java_else_on_new_line = false
+
+[*.{txt,md}]
+max_line_length = 80
+
+[*.xml]
+indent_size = 2
```
markt-asf, line 18: I'm not sure how consistent this is throughout the code base. I think the docs tend to use 2 but elsewhere it tends to be 4.

.editorconfig
larsgrefer: I took the numbers from here: https://github.com/apache/tomcat/blob/master/CONTRIBUTING.md#L127

.editorconfig
larsgrefer: More information can be found here: https://editorconfig.org/

markt-asf:
Thanks for this. I want to do a few checks locally and then I'll back-port it.

PR #285, opened 2020-04-24
Avoid waste of resources due to reconstruction of objects

KangZhiDong:
avoid waste of resources due to reconstruction of objects


markt-asf:
What is the performance benefit here? I'm concerned that some error paths, particularly around unhandled exceptions might lead to memory leaks. The PR needs very careful analysis before merging.

markt-asf:
Re-opening, as I have done most of the analysis required.

markt-asf:
The only one I'm concerned about is `JspFactoryImpl` as this would open up the possibility of one web application being able to impact another. I'll merge this manually so I can skip that class.

markt-asf:
Merged

PR #286, opened 2020-04-26
optimized code,replace some c-style array

XavierChengZW:
optimized code,replace some c-style array  Brackets are a part of an Array type. The definition could be: String[] args


martin-g:
Thank you, @XavierChengZW !

rmaucher:
Hmmm :(
So, ok, I guess you have to start somewhere, but it is best to have PRs on something meaningful (like at least a small bugfix or real improvement of some sort). Here, maybe at least do the change to the entire codebase instead of a single class.
Also the change then usually needs to be cherry picked to all current branches.
Overall this sort of PR creates a ton of noise.

martin-g:
Thanks for cherry-picking the change, Remy! I was interrupted after merging the PR and I was going to do it at lunch break.

PR #287, opened 2020-04-29
Fix typos

k-tamura:
Trivial typo... 
Fill GC -> Full GC


markt-asf:
Thanks. We use POEditor.com to manage translations. I've updated the string there and we will pick up the change the next time we import the translations from POEditor.com to Tomcat.

PR #288, opened 2020-04-30
keyload update

YYTVicky:


markt-asf:
Adding comments to the code is note the way to recommend code changes.
Also, I don't agree with the recommendation.
This PR will not be applied.

PR #289, opened 2020-05-18
Add mimetype for WASM files

Thihup:


markt-asf:
Thanks for the PR. I patched this by hand as there are a couple of other files that also need to be updated.

PR #290, opened 2020-05-21
If the log is info level, t2、t1 is long type，no need to convert to long type

EruDev:


markt-asf:
This will introduce an IDE warning for auto-boxing. The Tomcat code deliberately uses explicit conversion rather than auto-boxing.

PR #291, opened 2020-05-26
fixed-chinese-encoding

xiaofanTo:
When I use Chinese, I find that the log printed by the console is garbled，like this :
26-May-2020 09:42:32.374 严重 [main] org.apache.catalina.startup.Catalina.initDirs å¨[catalina-home1/temp]æ¾ä¸å°æå®çä¸´æ¶æä»¶å¤¹
26-May-2020 09:42:33.259 信息 [main] org.apache.catalina.startup.VersionLoggerListener.log Server.æå¡å¨çæ¬: Apache Tomcat/@VERSION@
26-May-2020 09:42:33.259 信息 [main] org.apache.catalina.startup.VersionLoggerListener.log æå¡å¨æå»º:        @VERSION_BUILT@
26-May-2020 09:42:33.259 信息 [main] org.apache.catalina.startup.VersionLoggerListener.log æå¡å¨çæ¬å·(:     @VERSION_NUMBER@
26-May-2020 09:42:33.260 信息 [main] org.apache.catalina.startup.VersionLoggerListener.log æä½ç³»ç»åç§°:      Mac OS X

So I changed the encoding so that there is no problem.

### DIFF
--- a/java/org/apache/tomcat/util/res/StringManager.java
+++ b/java/org/apache/tomcat/util/res/StringManager.java
@@ -146,7 +146,13 @@ public class StringManager {
             //      a null check.
             str = null;
         }
-
+        // change the encoing from "ISO-8859-1" to "UTF-8"
+        // Therefore, there will be no garbled characters when using Chinese
+        try {
+            str = new String(str.getBytes("ISO-8859-1"),"UTF-8");
+        } catch (UnsupportedEncodingException e) {
+            e.printStackTrace();
+        }
         return str;
     }
 


gmshake:
Can you please provide details to help reproducing the garbled console ? Required but not limited to OS version, JVM version, environment varibles and the steps to reproduce.

xiaofanTo:
ok , all info in this .
tomcat version : 8.5.55
java version : 1.8.0_181-b13
os version : Mac OS X 10.15.4

and i  debug this problem this morning .
I found the final cause of this problem is because of the **PropertyResourceBundle** class .

My test code is as follows :
`
public class Test {
    public static void main(String[] args) throws IOException {
        //1. inputstream
        InputStream inputStream = Test.class.getResourceAsStream("/test.properties");
        PropertyResourceBundle prb1 = new PropertyResourceBundle(inputStream);
        String title1 = prb1.getString("title");
        System.out.println(title1);

        //2. reader
        InputStreamReader reader = new InputStreamReader(Test.class.getResourceAsStream("/test.properties"),"UTF-8");
        PropertyResourceBundle prb2 = new PropertyResourceBundle(reader);
        String title2 = prb2.getString("title");
        System.out.println(title2);
    }
}
`
the test.properties contens is : 
`title=测试`

When you initialize him, the input is an InputStream , then Properties will load this stream，The garbled problem arises here . 
` public PropertyResourceBundle (InputStream stream) throws IOException {
        Properties properties = new Properties();
        properties.load(stream);
        lookup = new HashMap(properties);
    }
`

So I think you should replace the above with the following initialization method. 
` public PropertyResourceBundle (Reader reader) throws IOException {
        Properties properties = new Properties();
        properties.load(reader);
        lookup = new HashMap(properties);
    }`

when you use this reader , you can set the charset to **utf-8**

gmshake:
@xiaofanTo 

Refer to the javadoc of `java.util.Properties`,  https://docs.oracle.com/javase/8/docs/api/java/util/Properties.html, 

> The load(Reader) / store(Writer, String) methods load and store properties from and to a character based stream in a simple line-oriented format specified below. The load(InputStream) / store(OutputStream, String) methods work the same way as the load(Reader)/store(Writer, String) pair, except the input/output stream is encoded in ISO 8859-1 character encoding. Characters that cannot be directly represented in this encoding can be written using Unicode escapes as defined in section 3.3 of The Java™ Language Specification; only a single 'u' character is allowed in an escape sequence. The native2ascii tool can be used to convert property files to and from other character encodings.

What you experienced is expected behavior :)

markt-asf:
I'm guessing you are running Tomcat directly from source in an IDE. As @gmshake pointed out, you'll need to convert the property files to ASCII to do that (the Tomcat build process does this). There are various other approaches you can use depending on exactly what you are trying to achieve. The users mailing list is the place to seek help with this.
http://tomcat.apache.org/

PR #293, opened 2020-06-04
Fix IPUtils.canonize() to return a valid IPv6 address

syarramsetty-skyhook:


markt-asf:
Thanks for the PR. I applied it manually:

- because I had renamed the test class
- so I could include a changelog entry
- because the changes to the test code wouldn't compile

PR #295, opened 2020-06-11
Fix BZ 64513 - bndlib is not needed, normalize the name to bnd everyw…

rotty3000:
…here


markt-asf:
Applied manually so I could add a change log entry.

rotty3000:
@markt-asf if there is some way for me to include the changelog update let me know and I will include that in the future :)

rotty3000:
ah never mind, I found the changelog. I will update that from now on.

PR #297, opened 2020-06-11
Fix BZ 64515 - no need to filter bnd files (bnd passes ant properties…

rotty3000:
… in)


markt-asf:
I added the bug number of the changelog entry (I had missed in in the example you copied). The Tomcat committer id is always going to be tricky as you don't know which committer is going to review and merge the request. We need to figure out a way to handle that.

rotty3000:
Understood, if it's better to let you deal with the changelog I can do that. I just want to avoid more work for you if I could.

PR #298, opened 2020-06-11
Fix BZ 64514 - classes missing from bootstrap.jar

rotty3000:
Signed-off-by: Raymond Augé <rotty3000@apache.org

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -366,6 +366,7 @@
     <include name="org/apache/catalina/startup/CatalinaProperties.*" />
     <include name="org/apache/catalina/startup/ClassLoaderFactory.*" />
     <include name="org/apache/catalina/startup/ClassLoaderFactory$*.*" />
+    <include name="org/apache/catalina/startup/Constants.*" />
     <include name="org/apache/catalina/startup/SafeForkJoinWorkerThreadFactory.*" />
     <include name="org/apache/catalina/startup/SafeForkJoinWorkerThreadFactory$*.*" />
     <include name="org/apache/catalina/startup/Tool.*" />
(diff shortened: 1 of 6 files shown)


markt-asf:
I don't think this is correct. Only those classes required during the bootstrap process need to be in the bootstrap JAR. If this was not correct, I'd expect to see reports of ClassNotFoundException on Tomcat start and there aren't any.

Those classes may well have additional dependencies that are not used during the bootstrap process. Those dependencies do not belong in bootstrap.jar.

Taking ExceptionUtils as an example, there is an explicit workaround in Bootstrap to account for that class not being accessible.

For Globals, most (all?) of those will be constants which will be compiled into the class meaning there should be no runtime dependency on Globals.

And so on.

Unless you can provide a configuration that triggers a ClassNotFoundException on Tomcat start, I'm minded to resolve this as invalid.

rotty3000:
So would you rather make the work around in the metadata calculation once/if we get that far? Personally I rather make a valid jar than work around the metadata but that's just personal preference.

rotty3000:
From a JPMS point of view the goal is always that jdeps results match up with the module-info.class (i.e. that the jar is completely resolvable from the module path) and if you don't clear that you will always have to add a workaround to get the module-info you want (like adding a static module requires for the missing packages or forcing a manual ignore of the required module).

rotty3000:
For the record I'm more interested in establishing a baseline for how to handle these types of cases with the tomcat project. I will obviously follow your direction :)

markt-asf:
There are trade-offs here.
Modules require "clean packages" with all code from one package in one JAR.
Bootstrap uses utility code from elsewhere in the Tomcat code base.
We want to keep the bootstrap JAR small.
We don't want to duplicate code.

Hmm...

This feels more like a discussion for the dev list. I'll move this discussion there.

markt-asf:
Fixed via back-port (cherry-pick) of 10.0.x fix

PR #299, opened 2020-06-11
Fix BZ 64515 - no need to filter bnd files (bnd passes ant properties…

rotty3000:
… in)


markt-asf:
I back-ported this from 10.0.x

PR #300, opened 2020-06-13
Fix BZ 64521 - avoid moving i18n translations into classes dir since …

rotty3000:
…they are packaged into separate jars


markt-asf:
Merged on command line due to conflicts. Thanks for the PR.

PR #303, opened 2020-06-17
Fix BZ 64532 - update to bnd 5.1.1

rotty3000:

### DIFF
--- a/build.properties.default
+++ b/build.properties.default
@@ -286,14 +286,14 @@ saaj-api.home=${base.path}/saaj-api-${saaj-api.version}
 saaj-api.jar=${saaj-api.home}/saaj-api-${saaj-api.version}.jar
 saaj-api.loc=${base-maven.loc}/javax/xml/soap/saaj-api/${saaj-api.version}/saaj-api-${saaj-api.version}.jar
 
-# ----- bnd & bndlib, version 4.0.0 or later  -----
+# ----- bnd, version 5.1.1 or later  -----
 # ----- provides OSGI metadata for JARs       -----
-bnd.version=5.1.0
+bnd.version=5.1.1
 
-# checksums for biz.aQute.bnd-5.1.0.jar
+# checksums for biz.aQute.bnd-5.1.1.jar
 bnd.checksum.enabled=true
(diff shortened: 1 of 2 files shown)


build.properties.default
```diff
@@ -286,14 +286,14 @@ saaj-api.home=${base.path}/saaj-api-${saaj-api.version}
 saaj-api.jar=${saaj-api.home}/saaj-api-${saaj-api.version}.jar
 saaj-api.loc=${base-maven.loc}/javax/xml/soap/saaj-api/${saaj-api.version}/saaj-api-${saaj-api.version}.jar
 
-# ----- bnd & bndlib, version 4.0.0 or later  -----
+# ----- bnd, version 5.1.1 or later  -----
 # ----- provides OSGI metadata for JARs       -----
-bnd.version=5.1.0
+bnd.version=5.1.1
 
 # checksums for biz.aQute.bnd-5.1.0.jar
```
markt-asf, line 293: You need to update the version in the comment for consistency.

build.properties.default
```diff
@@ -286,14 +286,14 @@ saaj-api.home=${base.path}/saaj-api-${saaj-api.version}
 saaj-api.jar=${saaj-api.home}/saaj-api-${saaj-api.version}.jar
 saaj-api.loc=${base-maven.loc}/javax/xml/soap/saaj-api/${saaj-api.version}/saaj-api-${saaj-api.version}.jar
 
-# ----- bnd & bndlib, version 4.0.0 or later  -----
+# ----- bnd, version 5.1.1 or later  -----
```
markt-asf, line 289: Is this really an increase in the minimum version? Or is 5.1.0 (with the fix for the Java 15 bug) sufficient?

rotty3000: There is that yes, but also in 5.1.1 specifically there is a bug fix for correctly generating the Manifest Name section which I'd like to use for the Spec jars.

markt-asf: Fair enough. Just checking.

build.properties.default
rotty3000: Done!

rotty3000:
@markt-asf did I miss any other changes you wanted to see?

PR #304, opened 2020-06-18
Use https URLs when possible

bekionel:
- This is a follow up to https://github.com/apache/tomcat/commit/82f764ccec2

### DIFF
--- a/res/maven/README.txt
+++ b/res/maven/README.txt
@@ -6,7 +6,7 @@
   (the "License"); you may not use this file except in compliance with
   the License.  You may obtain a copy of the License at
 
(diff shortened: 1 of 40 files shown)


markt-asf:
In its current state applying this PR will trigger build failures (license header validation will fail).

I am also not convinced that the effort required to convert the link in every license header to https is worth the effort. However, if the community does decide to do this - and a conversation on the dev list is where/how that decision should be made - then all the files need to be changed, not just a sub-set.

markt-asf:
Closing as this PR cannot be applied as-is. Anyone interested in following through on this should start a discussion on the dev list as per my previous comment.

PR #307, opened 2020-06-23
github actions

rotty3000:
Run github actions on branch pushes and PRs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,49 @@
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+
+name: Smoke Test
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+env:
+  LC_ALL: en_US.UTF-8
+
+jobs:
+  JDKxx_Matrix:
+    strategy:
+      fail-fast: false
+      matrix:
+        java: [ 8, 11, 14, 15-ea ]
+        os: [ ubuntu-latest ]
+        include:
+        - os: windows-latest
+          java: 8
+    name: JDK${{ matrix.java }} ${{ matrix.os }}
+    runs-on: ${{ matrix.os }}
+    steps:
+    - name: Git Checkout
+      uses: actions/checkout@v2
+    - name: Set up Java
+      uses: actions/setup-java@v1
+      with:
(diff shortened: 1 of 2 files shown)


rotty3000:
This could be applied to each of the key branches.

rotty3000:
Also we can add to the matrix as we go.

martin-g:
So far we have:
1) Apache BuildBot as the official CI
2) Travis - for testing non-x86_64 architectures (at the moment: aarch64 and s390x)

Do we need a third CI ?
Since GitHub Actions comes for free and there is nearly zero cost maintaining it I am fine to add it to the mix. But it might be confusing for someone.
The pros are:
* test on Windows
* test PRs (this is the biggest issue I have with Apache BuildBot)

rotty3000:
I do not disagree that adding a third could be confusing. What I like about actions in addition to the points you've made is:

- runs on forks as well (which means I can iterate with CI before even sending PR to main repo)
- easily test full matrix of Java versions (including EAs, i.e. `15-ea`)
- is super fast

To differentiate between the official Apache BuildBot (which runs everything and takes very long because it is absolutely comprehensive) I was thinking to make use of github actions to be a form of soak test which builds all the main artifacts, on all supported Java versions, runs a small but comprehensive battery of tests, and runs a set of baseline tests against the 4 runtime assemblies:
- **original** - catalina.sh on classpath
- **embed** - need to figure out how to assemble this for run
- **OSGi** - already have it locally
- **jpms** - already have it locally

Finally, i would like to see all that run in no more than 10 minutes to allow for pretty darn fast iterations with low risk of regression.

As you can note in my change I've actually trimmed to run just the `nio` suite and disabled the openssl tests toward the goal of good testing but fast. Of course I need lots more input about what constitutes a comprehensive (yet fast) set of tests.

Thoughts?

rotty3000:
FYI, tomcat seems to build and test (nio) fine on Java 8, 11, 14 & 15-ea! :+1:

rotty3000:
FYI the latest iteration of this runs on Java 8 - 15-ea + windows in 11 minutes https://github.com/rotty3000/tomcat/actions/runs/146537422

The arbitrary reduction was made by filtering all tests that take more than 5 seconds to run. The rest seem to make a reasonable smoke test.

rmaucher:
Looks good to me. Personally, I don't panic when I see a few random failures.

Since this is meant as a smoke test, and in an effort to not spam too much about failures, I would do the following:
- Only test trunk (development is supposed to happen there, so although there are backport problems sometimes, this is not that common and will be caught by the main CI rather quickly)
- Use "test-nio", for some reason you use "test-nio2" which is not the default connector; but good idea on not testing all three connectors (or rather all six if you also include the JSSE/OpenSSL options
- Exclude tests that need a long time to run

rotty3000:
@rmaucher I have made the requested changes:
- only trunk
- use nio
- excluded tests slower than 5 seconds

rmaucher:
Nice, so it would seem it takes about 8 minutes now, down from 30 minutes initially. Eventually more "unimportant" tests could be excluded to get this down. I got confused with the "Container.tests" but that was just another branch name.

rotty3000:
Is anyone up for merging this? If something is missing please let me know :) I wasn't sure if this warranted a changelog.

rotty3000:
Also note that since this is handling push/pr on _master_ (to be renamed branch) there is no point applying this to other branches.

PR #311, opened 2020-07-02
support schemes "wss" and "ws" in WsHandshakeRequest.buildRequestUri()

sephiroth-j:
requests with scheme "wss" or "ws", as outlined in [RFC 6455 section 11.1](https://tools.ietf.org/html/rfc6455#section-11.1), will cause an `IllegalArgumentException` with message "wsHandshakeRequest.unknownScheme" without this change.

this fixes a regression from e09e6b04e41d3dae27fcfe060e6f8f89f1c0d6b5 / BZ 62731

### DIFF
--- a/java/org/apache/tomcat/websocket/server/WsHandshakeRequest.java
+++ b/java/org/apache/tomcat/websocket/server/WsHandshakeRequest.java
@@ -156,6 +156,8 @@ public class WsHandshakeRequest implements HandshakeRequest {
             uri.append("ws");
         } else if ("https".equals(scheme)) {
             uri.append("wss");
+        } else if ("wss".equals(scheme) || "ws".equals(scheme)) {
+            uri.append(scheme);
         } else {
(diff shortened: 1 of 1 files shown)


sephiroth-j:
i observed the exception in Spring Boot Project with Vaadin with an enabled `ForwardedHeaderFilter`. The setup was Traefik as reverse proxy with TLS termination forwarding the request from the client to the backend listening on plain HTTP. The client starts a websocket request using the URI scheme "wss://". Traefik adds the header "X-Forwared-Proto: wss" (_among others_). `ForwardedHeaderFilter` alters the scheme property of the `HttpServletRequest` to the value from the "X-Forwared-Proto" header. The result is an internal server error (500) caused by that `IllegalArgumentException`.

markt-asf:
Thanks for the PR

sephiroth-j:
you are welcome

PR #312, opened 2020-07-03
Use StringBuilder instead of StringBuffer

martin-g:
There is no need of synchronization when it is a method local variable.

Append character instead of String when possible

### DIFF
--- a/java/org/apache/catalina/Container.java
+++ b/java/org/apache/catalina/Container.java
@@ -305,7 +305,7 @@ public interface Container extends Lifecycle {
      * @return the configuration path
      */
     public static String getConfigPath(Container container, String resourceName) {
-        StringBuffer result = new StringBuffer();
+        StringBuilder result = new StringBuilder();
         Container host = null;
         Container engine = null;
         while (container != null) {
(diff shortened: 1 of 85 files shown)


rmaucher:
The "Append character instead of String when possible" seems wrong to me, and you're editing generated code there.

martin-g:
@rmaucher Could you please comment on the changed lines ?
I am not sure which code is generated and which `append(char)` might be wrong.
Thanks!

pzygielo:
> and which `append(char)` might be wrong.

I suppose, there is change in produced result
```
(1)              retval.append("\\t");
(2)              retval.append('\t');
```
as in first case "\t" (i.e. two-char string: `\` followed by `t`) is appended, and in second case - tab character ('\t') is.
Unfortunately I can't tell if there is test to validate what is expected as all smoke test jobs pass.

martin-g:
Thanks, @pzygielo !

Also these seem to be the generated classes!

martin-g:
I've reverted the changes to the generated classes.

What about the DBCP2 classes ? Maybe they should be changed upstream and then copied over here ?

markt-asf:
Yes, DBCP2 changes need to be made upstream.

martin-g:
OK. I've reverted the DBCP classes!
And made some more changes from `.append(String)` to `.append(char)`

rmaucher:
Looks good to me now. The change for rewrite is likely the most significant, the rest is probably not doing anything though.

PR #316, opened 2020-07-04
Use lowercase HTML tags and attributes.  Add quotes around attributes.

jbampton:


kkolinko:
error.html:
1. Shouldn't it better be <form method="GET"?
2. Options shall be closed with </option>. The <br>s there are out of place and ignored. (validator.w3.org complains)

jbampton:
Hey @kkolinko I removed the `br` tags and also closed the `option` tags now.

I also validated the 'error.html' by direct upload and it validates now.

 https://validator.w3.org/#validate_by_input

webapps/ROOT/index.jsp
```diff
@@ -25,11 +25,11 @@ request.setAttribute("tomcatExamplesUrl", "/examples/");
 <!DOCTYPE html>
 <html lang="en">
     <head>
-        <meta charset="UTF-8" />
+        <meta charset="UTF-8">
```
martin-g, line 28: Why did you remove the auto-closing ? It doesn't do any harm. I, personally, prefer the style it was before.

webapps/examples/jsp/cal/cal1.jsp
```diff
@@ -15,13 +15,12 @@
   limitations under the License.
 --%>
 <%@page contentType="text/html; charset=UTF-8" %>
-<HTML>
-<HEAD><TITLE>
-    Calendar: A JSP APPLICATION
-</TITLE></HEAD>
+<html>
+<head>
+<title>Calendar: A JSP APPLICATION</title>
```
martin-g, line 20: There is no indentation here.

webapps/examples/jsp/cal/cal1.jsp
```diff
@@ -32,63 +31,60 @@
...
-<TD ALIGN=CENTER> Calendar:<%= table.getDate() %></TD>
-<TD ALIGN=CENTER> <A HREF=cal1.jsp?date=next> next </A>
-</TR>
-</TABLE>
+<center>
+<table width="60%" bgcolor="yellow" cellpadding="15">
+<tr>
+<td align="center"> <a href="cal1.jsp?date=prev"> prev </a>
+<td align="center"> Calendar:<%= table.getDate() %></td>
+<td align="center"> <a href="cal1.jsp?date=next"> next </a>
```
martin-g, line 39: no indentation

webapps/examples/jsp/cal/cal2.jsp
```diff
@@ -15,31 +15,30 @@
...
-<BR> <INPUT NAME="date" TYPE=HIDDEN VALUE="current">
-<BR> <INPUT NAME="time" TYPE=HIDDEN VALUE="<%= util.HTMLFilter.filter(time) %>">
-<BR> <h2> Description of the event <INPUT NAME="description" TYPE=TEXT SIZE=20> </h2>
-<BR> <INPUT TYPE=SUBMIT VALUE="submit">
-</FORM>
-
-</BODY>
-</HTML>
+<font size="5"> Please add the following event:
+<br> <h3> Date <%= table.getDate() %>
```
martin-g, line 31: indentation

webapps/ROOT/index.jsp
jbampton: In HTML5 you don't need the auto-closing.
I got the style from MDN.
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meta

martin-g: I know, but it does not do any harm either.
This change makes the diff bigger and harder to back-port to the other branches.
I prefer the XHTML style because it is easier to spot errors in the IDE/text editor without copying to W3C validator in the web.

webapps/ROOT/index.jsp
```diff
@@ -25,11 +25,11 @@ request.setAttribute("tomcatExamplesUrl", "/examples/");
...
         <title><%=request.getServletContext().getServerInfo() %></title>
-        <link href="favicon.ico" rel="icon" type="image/x-icon" />
-        <link href="favicon.ico" rel="shortcut icon" type="image/x-icon" />
-        <link href="tomcat.css" rel="stylesheet" type="text/css" />
+        <link href="favicon.ico" rel="icon" type="image/x-icon">
```
stokito, line 30: in fact we can remove the line but the favicon.ico will be anyway requested by a browser. The tag makes sense when the file name is another

jbampton:
> error.html:
> 
> 1. Shouldn't it better be <form method="GET"?
> 2. Options shall be closed with . The s there are out of place and ignored. (validator.w3.org complains)

MDN says -> Possible (case insensitive) values
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form

Screen Shot 2020-08-01 at 5 31 33 am

ChristopherSchultz:
I'm a -1 on this change mostly because it's so large with so many changes at the same time. If this PR had been the accumulation of several commits e.g. (1) re-format whitespace everywhere with no other changes then (2) remove unnecessary auto-close elements then (3) standardize quoting, etc. then it would have been easier to review, spot things we didn't like, etc. But this is one giant change I'm unwilling to read through to decide if I like every single one of them so, no.

PR #319, opened 2020-07-08
Fix bare URL

jbampton:
https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md034---bare-url-used


martin-g:
Thank you, @jbampton !

jbampton:
Thanks @martin-g 👍

PR #330, opened 2020-08-04
BZ-64644 throw an Idle Session Event on idle timeout

sakshamverma:
Bug ID Bug ID: https://bz.apache.org/bugzilla/show_bug.cgi?id=64644

The demo is also present there.
I found a bug while working on my product of IoT platform.
On network outage, the devices still showed connected due to the fact that tomcat didn't expire the existing connection.
KIndly review.

### DIFF
--- a/java/javax/websocket/OnIdleSession.java
+++ b/java/javax/websocket/OnIdleSession.java
@@ -1,0 +1,12 @@
+package javax.websocket;
+
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.METHOD)
+public @interface OnIdleSession {
+
+}
--- a/java/javax/websocket/Session.java
+++ b/java/javax/websocket/Session.java
@@ -78,6 +78,19 @@ public interface Session extends Closeable {
     void setMaxIdleTimeout(long timeout);
 
     /**
+     * Get the flag to decide whether the session will be closed if idle Time is passed
+     * @return true if the session is to be closed on expire
+     *         false if the session will remain open and an IdleSession event is thrown instead
+     */
+    boolean getCloseOnIdleTimeout();
+
+    /**
+     * Get the flag to decide whether the session will be closed if idle Time is passed
+     * @param clsoeOnIdleTimeout- true for closing the session on idle timeout
+     *                            false for keeping the session open and throwing IdleSession event instead.
+     */
+    void setCloseOnIdleTimeout(boolean clsoeOnIdleTimeout);
+    /**
      * Set the current maximum buffer size for binary messages.
      * @param max The new maximum buffer size in bytes
      */
--- a/java/javax/websocket/Session.java
+++ b/java/javax/websocket/Session.java
@@ -86,10 +86,11 @@ public interface Session extends Closeable {
 
     /**
      * Get the flag to decide whether the session will be closed if idle Time is passed
-     * @param clsoeOnIdleTimeout- true for closing the session on idle timeout
+     * @param closeOnIdleTimeout- true for closing the session on idle timeout
      *                            false for keeping the session open and throwing IdleSession event instead.
      */
-    void setCloseOnIdleTimeout(boolean clsoeOnIdleTimeout);
+    void setCloseOnIdleTimeout(boolean closeOnIdleTimeout);
+
     /**
      * Set the current maximum buffer size for binary messages.
      * @param max The new maximum buffer size in bytes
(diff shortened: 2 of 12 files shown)


java/javax/websocket/OnIdleSession.java
```diff
@@ -0,0 +1,12 @@
+package javax.websocket;
```
martin-g, line 1: I'm afraid this is not part of WebSocket specification: https://www.jcp.org/en/jsr/detail?id=356 and https://download.oracle.com/otndocs/jcp/websocket-1_1-mrel-eval-spec/index.html
So you cannot add it to `javax.websocket` package

java/javax/websocket/Session.java
```diff
@@ -77,6 +77,19 @@
...
+     *         false if the session will remain open and an IdleSession event is thrown instead
+     */
+    boolean getCloseOnIdleTimeout();
+
+    /**
+     * Get the flag to decide whether the session will be closed if idle Time is passed
+     * @param clsoeOnIdleTimeout- true for closing the session on idle timeout
+     *                            false for keeping the session open and throwing IdleSession event instead.
+     */
+    void setCloseOnIdleTimeout(boolean clsoeOnIdleTimeout);
```
martin-g, line 92: typo in the parameter name

java/org/apache/tomcat/websocket/WsSession.java
```diff
@@ -98,6 +99,8 @@
     private volatile int maxTextMessageBufferSize = Constants.DEFAULT_BUFFER_SIZE;
     private volatile long maxIdleTimeout = 0;
     private volatile long lastActive = System.currentTimeMillis();
```
martin-g, line 101: Probably this one should be renamed to `lastWrite` ?!

java/org/apache/tomcat/websocket/WsSession.java
```diff
@@ -98,6 +99,8 @@
...
+    private volatile long lastActiveRead = System.currentTimeMillis();
```
martin-g, line 102: `lastRead` ?

java/org/apache/tomcat/websocket/WsSession.java
```diff
@@ -809,20 +825,36 @@ protected void updateLastActive() {
...
+        if (isWriteTimeout || isReadTimeout) {
+        	if(closeOnIdle) {
+                String msg = sm.getString("wsSession.timeout", getId());
+                if (log.isDebugEnabled()) {
+                    log.debug(msg);
+                }
+                doClose(new CloseReason(CloseCodes.GOING_AWAY, msg),
+                        new CloseReason(CloseCodes.CLOSED_ABNORMALLY, msg));
+        	} else {
+        		String msg = sm.getString("wsSession.timeout", getId());
```
martin-g, line 850: indentation is off

java/javax/websocket/OnIdleSession.java
sakshamverma: Would it be okay if I move it to the package org.apache.tomcat.websocket?
And what about the other chagnes in the javax package?

markt-asf:
There needs to be a *really* good reason to essentially introduce a Tomcat specific extension to the Jakarta WebSocket specification and thie PR hasn't (yet) provided such a justification.
If there is a good reason for adding this feature then it would be better added via the Jakarta WebSocket project where the various edge cases can be addressed.
My reading of this patch is once the idle event has been fired once, it will fire every time `WsSession.checkExpiration()` is called unless the associated handler sends a message. That seems pointless so I assume the intention is that the handler will send a message when the event fires. The end result is that a message will be sent at least every timeout milliseconds and the session will never timeout. There are simpler ways to ensure an endpoint sends a message every x milliseconds.
Absent a suitable justification for adding the feature, this PR is likely to be closed.

sakshamverma:
@markt-asf 
I have tried to explain the problem in the video: https://youtu.be/BJf5-BpkVX8

Tomcat is not expiring the websocket channel even if there is no network. 
The way any application makes use of WS is to have some successful response from the other endpoint after sending a message.
So, if the idle session event is thrown that would mean that the network is not there. 

And the default behavior is also not changed. Its upto the products, when do they want to close the connection.

Let me know if we can have a zoom meeting to give another demo if required to understand the problem better.

markt-asf:
The proposed approach is invasive, would require significant changes to applications for them to make use of it and - effectively - extends the Jakarta WebSocket API in a proprietary way.  I am therefore closing this PR as the approach proposed is not going to be adopted in Tomcat without it first being included in the WebSocket specification.
I have proposed a simpler approach in https://bz.apache.org/bugzilla/show_bug.cgi?id=64644 and discussion should continue there.

sakshamverma:
@markt-asf 

I have raised another PR https://github.com/apache/tomcat/pull/346 as per your suggestion. 

Could you please let me know what are the next steps for it to merge? It will be a great help for our product (Thingworx) which uses tomcat as the webserver.

PR #331, opened 2020-08-04
Remove White Spaces and extra lines from the JSP files

kamnani:
The changes enables the compiler to remove excess white spaces and new lines from the JSP files & thus reduce the JVM metadata.
This can be controlled by providing context init params inside `web.xml` file. Example attached. 

By Default this remains false, and thus the source JSP files will not reflect any changes compared to the previous builds.

Note: 
1 ) The changes do not affect to the pre tags, to protect the behavioral changes on the tag. 
2 ) This change is for Tomcat 9 Version. 

If any official documentation is required, can you attach the links on the PR? 

```
<web-app>
<context-param>
    <param-name>JSPWhiteSpaceTrimming</param-name>
    <param-value>true</param-value>
</context-param> 
</web-app>

### DIFF
--- a/java/org/apache/jasper/compiler/NewlineReductionServletWriter.java
+++ b/java/org/apache/jasper/compiler/NewlineReductionServletWriter.java
@@ -1,0 +1,40 @@
+package org.apache.jasper.compiler;
+
+import java.io.PrintWriter;
+
+/**
+ * This class filters duplicate newlines instructions from the compiler output,
+ * and therefore from the runtime JSP. The duplicates typically happen because
+ * the compiler has multiple branches that write them, but they operate
+ * independently and don't realize that the previous output was identical.
+ *
+ * Removing these lines makes the JSP more efficient by executing fewer operations during runtime.
+ *
+ * @author Engebretson, John
+ * @author Kamnani, Jatin
+ *
+ */
+public class NewlineReductionServletWriter extends ServletWriter {
+    private static final String NEWLINE_WRITE_TEXT = "out.write('\\n');";
+    private boolean lastWriteWasNewline;
+
+    public NewlineReductionServletWriter(PrintWriter writer) {
+        super(writer);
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by kamnani** [comment]
@ChristopherSchultz  Are there any updates on this ?

**Comment by ChristopherSchultz** [comment]
Why tag me, specifically?

This seems fragile. The `<pre>` tag-handling looks like it will only work if all the text is in the same text block. That's probably only going to be true in some very special cases.

Adding this configuration should be done via a JSPServlet `<init-param>` and not a system property.

**Comment by kamnani** [comment]
I can remove this optimization for the pre tag, so pre tags are untouched.

File: java/org/apache/jasper/compiler/NewlineReductionServletWriter.java [hunk 1]
```diff
@@ -0,0 +1,40 @@
...
+ * independently and don't realize that the previous output was identical.
+ *
+ * Removing these lines makes the JSP more efficient by executing fewer operations during runtime.
+ *
+ * @author Engebretson, John
+ * @author Kamnani, Jatin
+ *
+ */
+public class NewlineReductionServletWriter extends ServletWriter {
+    private static final String NEWLINE_WRITE_TEXT = "out.write('\\n');";
```
markt-asf, line 18: There are no calls that match this pattern in the Tomcat code base. Due to this point alone, I don't see the justification for this feature.
Skipping blank lines like this is going to break SMAP support. That isn't a show stopper but needs to be noted in the documentation.

kamnani: SMAP Support defaults to false, and this feature has a property which can be default to false as well. 
Can you provide me the links where I can write the information on this patch ? @markt-asf

kamnani:
#351 Opened after re-work. Closing this PR.

PR #333, opened 2020-08-06
[Patch] Support INDEX.LIST in jar file to speed up resource finding

kelthuzadx:
`jdk.internal.util.jar.JarIndex` was introduced by JDK1.3, it's an old but useful technique. If third-party JAR files contain INDEX.LIST file(generated via `jar -i` command), these JAR files can be identified as loading into JVM. Class loader such as BuiltinClassLoader,URLClassLoader and subclasses of URLClassLoader can speed up their resource searching with the help of INDEX.LIST since it maps resource names to corresponding JAR names. 

The latest version of Tomcat does not aware of `INDEX.LIST`, when there are so many JAR files in `WEB-INF/lib`,  Tom cat have to search in all JAR files sequentially, which is obviously slow. This patch helps Tomcat to be aware of `INDEX.LIST`, speeds up the search of resources, especially when there are many JAR files in `WEB-INF/lib`

### DIFF
--- a/java/org/apache/catalina/WebResourceSet.java
+++ b/java/org/apache/catalina/WebResourceSet.java
@@ -18,6 +18,7 @@ package org.apache.catalina;
 
 import java.io.InputStream;
 import java.net.URL;
+import java.util.List;
 import java.util.Set;
 
 /**
@@ -35,6 +36,10 @@ public interface WebResourceSet extends Lifecycle {
      * @return  The object that represents the resource at the given path
      */
     WebResource getResource(String path);
+
+    default WebResource getResource(String path, List<List<WebResourceSet>> allResources){
+        return getResource(path);
+    }
 
     /**
      * Obtain the list of the names of all of the files and directories located
--- a/java/org/apache/catalina/webresources/JarResourceSet.java
+++ b/java/org/apache/catalina/webresources/JarResourceSet.java
@@ -16,11 +16,15 @@
  */
 package org.apache.catalina.webresources;
 
+import java.net.URL;
+import java.util.List;
 import java.util.jar.JarEntry;
 import java.util.jar.Manifest;
 
 import org.apache.catalina.WebResource;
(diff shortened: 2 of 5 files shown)


java/org/apache/tomcat/util/scan/JarIndex.java
```diff
@@ -0,0 +1,61 @@
+package org.apache.tomcat.util.scan;
```
markt-asf, line 1: No ALv2 Header
Why is this class in this package when it is only used in `org.apache.catalina.webresources`?

java/org/apache/catalina/WebResourceSet.java
```diff
@@ -36,6 +37,10 @@
      */
     WebResource getResource(String path);
 
+    default WebResource getResource(String path, List<List<WebResourceSet>> allResources){
```
markt-asf, line 40: This looks wrong. A single `WebResourceSet` shouldn't need to be passed a reference to `allResources`.

java/org/apache/catalina/webresources/JarResourceSet.java
```diff
@@ -59,6 +63,36 @@ public JarResourceSet(WebResourceRoot root, String webAppMount, String base,
...
+        }
+        String webAppMount = getWebAppMount();
+        if (path.startsWith(getWebAppMount())) {
+            String pathInJar = getInternalPath() + path.substring(
+                    webAppMount.length());
+            if (pathInJar.length() > 0 && pathInJar.charAt(0) == '/') {
+                pathInJar = pathInJar.substring(1);
+            }
+            List<String> jarFiles = jarIndex.get(pathInJar);
+            if(jarFiles!=null){
```
markt-asf, line 80: Why is this here rather than in `StandardRoot` ?

kelthuzadx:
I will fix these problems and give a fresh new patch, anyway, thanks to your review.

PR #334, opened 2020-08-06
Fix https://bz.apache.org/bugzilla/show_bug.cgi?id=64614

jfclere:
I was not able to convince the JVM in "FIPS mode" to use an alias, if there are several key/cert in the keystore it will use the latest one, therefore the following patch.


java/org/apache/tomcat/util/net/LocalStrings.properties
```diff
@@ -176,3 +176,4 @@ sslUtilBase.ssl3=SSLv3 has been explicitly enabled. This protocol is known to be
 sslUtilBase.tls13.auth=The JSSE TLS 1.3 implementation does not support authentication after the initial handshake and is therefore incompatible with optional client authentication
 sslUtilBase.trustedCertNotChecked=The validity dates of the trusted certificate with alias [{0}] were not checked as the certificate was of an unknown type
 sslUtilBase.trustedCertNotValid=The trusted certificate with alias [{0}] and DN [{1}] is not valid due to [{2}]. Certificates signed by this trusted certificate WILL be accepted
+sslUtilBase.alias_ignored=Alias name [{0}] is ignored
```
markt-asf, line 179: Maybe add a reason here. e.g.
"FIPS enabled so alias name [{0}] will be ignored. If there is more than one key in the key store, the key used will depend on the key store implementation"

jfclere: Sure I will change the message.

markt-asf:
I merged manually so i could add a change log entry.

PR #335, opened 2020-08-07
Message digest credential handler fix

MilovdZee:
For SSHA the order of salt and digest was reverse to the documentation. It was "{SSHA}<digest:20><salt:20>".

I added SSHA2 to also allow for SHA-256 and other algorithms. I introduced {SSHA2} for this. It could be confusing because it also support SHA-512 and others.

The same holds for tomcat-9 but let us first fix the 10.x.x branch. If accepted I'll also create a 9.x.x PR.


MilovdZee:
oops. wrong master

PR #338, opened 2020-08-12
Compare entity tag

stokito:
Sorry for being annoying, this is just refactoring and performance optimization.
Instead of parsing `If-Match` header to a HashSet we can make in-place comparison.
Thus we can avoid of generation of garbage.


markt-asf:
Thanks. Applied manually so I could tweak some of the Javadoc and code comments.

PR #340, opened 2020-08-17
Optimize server startup time

kamnani:
Tomcat Takes quite a bit of time to start the server if there are 1000's of Jar's in the class path. This wastes a lot of developer time, where we need to start and stop the server quite often. 
This pull request has a couple of optimizations that reduces the server startup time : 

1)  Accelerate classloading

Optimization 1:  *Cache and maintain file handles for each jar.  This adds 1MB of old generation memory but eliminates 100,000,000+ file open/close operations.

Optimization 2: *Cache the contents of each jar in memory via a Bloom filter.  This adds 1.5MB of old generation memory but eliminates 100,000,000+ archive lookups.

2) Parallelize annotation scans

Tomcat searches the entire classpath for all instances of certain annotations, such as @WebServlet, @WebFilter, @MultipartConfig, etc.  This is performed by scanning each classpath entry sequentially.

Optimization:  *Scan the jars in parallel.

### DIFF
--- a/conf/catalina.properties
+++ b/conf/catalina.properties
@@ -204,3 +204,6 @@ tomcat.util.buf.StringCache.byte.enabled=true
 #tomcat.util.buf.StringCache.char.enabled=true
 #tomcat.util.buf.StringCache.trainThreshold=500000
 #tomcat.util.buf.StringCache.cacheSize=5000
+
+# Flag represents whether we want to use multi-threading while starting the tomcat server.
+tomcat.fastServerStartup=true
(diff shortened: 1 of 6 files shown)


kamnani:
"Configuration should not be via system properties unless absolutely necessary. In this instance configuration should probably be on the Host." 
This configurations here affects the Java behavior, so the reason I mentioned it via the System Property route. Can you be more specific on that?

markt-asf:
New PRs #348 and #349 submitted.

PR #341, opened 2020-08-19
Update README.md

PoojaChandak:
Observed a few changes. Suggesting the same. Kindly check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,8 +2,8 @@
 
 ### What Is It?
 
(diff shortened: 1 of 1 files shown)


markt-asf:
The change from "logo" to "logos" is incorrect. The remainder are purely style changes that are inconsistent with the remainder of the Tomcat documentation. We use "open source" not "open-source". We don't use oxford commas.

PR #343, opened 2020-08-25
Ensure ExpressionFactory SPI registration is compatible with javax/jakarta API jar

rmannibucau:
https://github.com/javaee/el-spec/blob/master/api/src/main/java/javax/el/FactoryFinder.java#L151 only reads one line (jakarta el-api jar), tomcat license header breaks it and is not strictly required (there is a tolerance for these files which sometimes don't support headers) so let's drop it.

### DIFF
--- a/res/META-INF/jasper-el.jar/services/jakarta.el.ExpressionFactory
+++ b/res/META-INF/jasper-el.jar/services/jakarta.el.ExpressionFactory
@@ -1,16 +1,1 @@
-# Licensed to the Apache Software Foundation (ASF) under one or more
-# contributor license agreements.  See the NOTICE file distributed with
-# this work for additional information regarding copyright ownership.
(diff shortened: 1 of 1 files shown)


markt-asf:
That version of the EL API is broken. It should be using ServiceLoader. This has been fixed in Jakarta EE. It is a general principle that the Tomcat project doesn't implement work-arounds for bugs in other projects except in extreme circumstances. In this instance, a working version of the EL is available from the Tomcat project.
This PR will not be merged.

rmannibucau:
@markt-asf can you point out the fixed release? seems https://mvnrepository.com/artifact/jakarta.el/jakarta.el-api/3.0.3 is affected by that so this is an issue ATM and as soon as you use tomcat embedded or its flavors (spring boot with cxf for ex) it hurts.

PR #344, opened 2020-08-25
Minor performance improvements

martin-g:
I've noticed few small improvements to Tomcat code while profiling it with Java Flight Recorder and Yourkit.
Each improvement is a separate commit with an explanation of the change.
This PR is just for discussion and will be marked as a draft PR so it won't be merged.
If the separate commits are approved I will push them directly to the active branches.


martin-g:
Thank you for the review!

The commits have been applied to `master`, `9.0.x`, `8.5.x` and `7.0.x` branches.

PR #346, opened 2020-08-27
Bug 64644 add custom properties to close connections on readIdletimeout

sakshamverma:


sakshamverma:
@markt-asf 

Could you please review this PR
Its related to the PR https://github.com/apache/tomcat/pull/330 which was not appropriate. 

The current PR is as per your suggestion on the bug https://bz.apache.org/bugzilla/show_bug.cgi?id=64644

Added 2 custom properties for session 
IdleReadTimeout
IdleWriteTimeout

Need to check if I can remove the 2 other properties I added : closeOnReadIdleTimeout and closeOnWriteIdleTimeout if they are not needed.

java/org/apache/tomcat/websocket/Constants.java
```diff
@@ -114,6 +114,22 @@
...
 
+    // Configuration for idle read timeout on websocket channel
+    public static final String WS_READ_IDLE_TIMEOUT =
+            "org.apache.tomcat.websocket.WS_READ_IDLE_TIMEOUT";
+
+    // Configuration for idle write timeout on websocket channel
+    public static final String WS_WRITE_IDLE_TIMEOUT =
+            "org.apache.tomcat.websocket.WS_WRITE_IDLE_TIMEOUT";
+
+    // Configuration for idle write timeout on websocket channel
```
martin-g, line 125: This comment is exactly the same as for the previous constant

java/org/apache/tomcat/websocket/Constants.java
```diff
@@ -114,6 +114,22 @@
...
+    public static final String WS_CLOSE_READ_IDLE_TIMEOUT =
+            "org.apache.tomcat.websocket.WS_CLOSE_READ_IDLE_TIMEOUT";
+
+    // Configuration for idle write timeout on websocket channel
```
martin-g, line 129: Same copy/paste error here

martin-g (approved):

java/org/apache/tomcat/websocket/Constants.java
martin-g: What is the purpose of the following two constants ? They are not used below.

sakshamverma: These were part of initial draft. 
I intended to remove them. 
Will do in next commit.

java/org/apache/tomcat/websocket/Constants.java
sakshamverma: Removed the parameters which are not used

markt-asf:
Thanks for the updated PR. This needs to be applied to 10.0.x (master) first and then back-ported. There are also a few small things I'd like to tweak so I'll apply this manually rather than merging it. I'll start on that process now.

sakshamverma:
@markt-asf 

Thanks for your comment. 
Do you want me to raise the PR for master branch? or would you be raising that with the tweaks you have in mind?

markt-asf:
No need for another PR. I'll just work directly on master.

java/org/apache/tomcat/websocket/WsSession.java
```diff
@@ -816,16 +823,62 @@ protected void checkExpiration() {
...
+        Long userReadTimeout = null;
+        if (readTimeout instanceof Long) {
+            userReadTimeout = (Long) readTimeout;
+        }
+        return userReadTimeout;
+    }
+
+    private Long getIdleWriteTimeout() {
+        Object writeTimeout = this.getUserProperties().get(Constants.WS_WRITE_IDLE_TIMEOUT);
+        Long userWriteTimeout = null;
```
markt-asf, line 876: No need for the additional reference here

java/org/apache/tomcat/websocket/WsSession.java
```diff
@@ -816,16 +823,62 @@ protected void checkExpiration() {
...
+            closeConnection = idleWriteTimeout != null && (currentTime - lastWriteTime > idleWriteTimeout);
+
+            if (closeConnection) {
+                String msg = sm.getString("wsSession.writeIdleTimeout", getId());
+                if (log.isDebugEnabled()) {
+                    log.debug(msg);
+                }
+                doClose(new CloseReason(CloseCodes.GOING_AWAY, msg),
+                        new CloseReason(CloseCodes.CLOSED_ABNORMALLY, msg));
+            }
```
markt-asf, line 861: Don't need to perform this check if the read timeout has already triggered. Don't want the possibility of multiple close messages.

java/org/apache/tomcat/websocket/WsSession.java
```diff
@@ -805,8 +807,13 @@ protected MessageHandler getBinaryMessageHandler() {
...
 
 
-    protected void updateLastActive() {
+    protected void updateLastActive(boolean isWriteOperation) {
         lastActive = System.currentTimeMillis();
+        if (isWriteOperation) {
+            lastWriteTime = lastActive;
+        } else {
+            lastReadTime = lastActive;
+        }
```
markt-asf, line 816: On reflection I decided to split this into two methods for clarity when reading the calling code.

test/org/apache/tomcat/websocket/TestWsWebSocketContainer.java
```diff
@@ -779,6 +779,104 @@ public void testSessionExpirySession() throws Exception {
         Assert.assertEquals(0, getOpenCount(setA));
     }
 
+    @Test
+    public void testSessionExpiryOnUserPropertyReadIdleTimeout() throws Exception {
+        final String readIdleTimeoutParameter = org.apache.tomcat.websocket.Constants.WS_READ_IDLE_TIMEOUT;
```
markt-asf, line 784: I opted to use the Constants class (as it is in the same package) and use the fully qualified name with the Constants from the server package.

markt-asf:
Applied manually.

PR #348, opened 2020-08-31
Optimize server startup time using Bloom Filter

kamnani:
Bloom Filter : https://llimllib.github.io/bloomfilter-tutorial/#:~:text=A%20Bloom%20filter%20is%20a,may%20be%20in%20the%20set.

This code enables a faster lookup of resources in the jar by eliminating the time needed to search for a resource in a Jar, where it is not going to be present. The Flag can be set by using catalina properties.

Flag defaults to false.

### DIFF
--- a/java/org/apache/catalina/webresources/AbstractArchiveResourceSet.java
+++ b/java/org/apache/catalina/webresources/AbstractArchiveResourceSet.java
@@ -36,11 +36,12 @@ public abstract class AbstractArchiveResourceSet extends AbstractResourceSet {
 
     private URL baseUrl;
     private String baseUrlString;
-
+    private final boolean FAST_STARTUP = Boolean.getBoolean("tomcat.fastServerStartup");
     private JarFile archive = null;
     protected Map<String,JarEntry> archiveEntries = null;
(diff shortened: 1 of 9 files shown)


java/org/apache/catalina/webresources/AbstractArchiveResourceSet.java
```diff
@@ -283,7 +310,7 @@ public final WebResource getResource(String path) {
     protected abstract boolean isMultiRelease();
 
     protected abstract WebResource createArchiveResource(JarEntry jarEntry,
-            String webAppPath, Manifest manifest);
+                                                         String webAppPath, Manifest manifest);
```
markt-asf, line 313: Revert change to indent

markt-asf (changes requested):
It is disappointing that I find myself repeating comments I have made on previous versions of this PR.

- PRs should be against master. They will be back-ported to earlier versions as appropriate
- Configuration should not be via system properties unless absolutely necessary. In this instance configuration should probably be on the Host.

Two separate approaches have been defined to improve performance in this area. This approach and jar indexes. We really need to compare the two to see which is best.

PR #349, opened 2020-08-31
Optimize Server Startup time

kamnani:
Tomcat searches the entire classpath for all instances of certain annotations, such as @WebServlet, @webfilter, @MultipartConfig, etc. This is performed by scanning each classpath entry sequentially.

Optimization: *Scan the jars in parallel

The default behavior stays as is, but this could be triggered by defining the flag inside catalina properties

### DIFF
--- a/java/org/apache/catalina/startup/ContextConfig.java
+++ b/java/org/apache/catalina/startup/ContextConfig.java
@@ -111,6 +111,10 @@ import org.apache.tomcat.util.res.StringManager;
 import org.apache.tomcat.util.scan.JarFactory;
 import org.xml.sax.InputSource;
 import org.xml.sax.SAXParseException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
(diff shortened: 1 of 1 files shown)


markt-asf (changes requested):
It is disappointing that I find myself repeating comments I have made on previous versions of this PR.

- PRs should be against master. They will be back-ported to earlier versions as appropriate
- Comments should wrap at 80 chars, code at 120.
- Configuration should not be via system properties unless absolutely necessary. In this instance configuration should probably be on the Host.
- Similar changes have been proposed previously and have been rejected for the same reason. Using one thread per core is not acceptable as that will impact processing of requests (where automatic deployment is enabled). The number of threads used needs to be configurable (probably on the Host since it is the Host that is doing the deploying and it needs to default to 1.

I'll add that applying this PR may cause additional issues where multiple web applications are deployed as web application deployment on startup can also be performed in parallel. Again, the number of threads used for annotation scanning needs to be configurable and needs to default to 1.

markt-asf:
Updated PR provided at #354

PR #352, opened 2020-09-02
Optimizing Resource Lookup using  Bloom Filter

kamnani:
This is a redo of Previous PR: https://github.com/apache/tomcat/pull/348

The following changes have been made based on the suggestions earlier: 
1) Flag can be passed through Host Configuration. By default it remains false and Tomcat's default behavior will be untouched. See example below. 
2) Jar Contents will be refreshed every 60 seconds if corresponding jar file has been modified.  
3) PR is against master. 

Apologies in case something is missed. 

Example on Bloom Filter : https://llimllib.github.io/bloomfilter-tutorial/#:~:text=A%20Bloom%20filter%20is%20a,may%20be%20in%20the%20set.

```
            unpackWARs="true" autoDeploy="true" fastClasspathScanning="true">

### DIFF
--- a/java/org/apache/catalina/webresources/AbstractArchiveResourceSet.java
+++ b/java/org/apache/catalina/webresources/AbstractArchiveResourceSet.java
@@ -27,20 +27,26 @@ import java.util.jar.JarEntry;
 import java.util.jar.JarFile;
 import java.util.jar.Manifest;
 
+import org.apache.catalina.Host;
 import org.apache.catalina.WebResource;
 import org.apache.catalina.WebResourceRoot;
 import org.apache.catalina.util.ResourceSet;
 import org.apache.tomcat.util.compat.JreCompat;
+import java.util.logging.Logger;
 
 public abstract class AbstractArchiveResourceSet extends AbstractResourceSet {
 
+    private final static Logger log = Logger.getLogger(AbstractArchiveResourceSet.class.getName());
     private URL baseUrl;
     private String baseUrlString;
-
     private JarFile archive = null;
     protected Map<String,JarEntry> archiveEntries = null;
     protected final Object archiveLock = new Object();
     private long archiveUseCount = 0;
+    private JarContents jarContents;
+    private JarFile jar;
+    private long prevJarOpenTime;
+    private final int jarRefreshTime = 60000;
 
 
     protected final void setBaseUrl(URL baseUrl) {
@@ -210,6 +216,50 @@ public abstract class AbstractArchiveResourceSet extends AbstractResourceSet {
         checkPath(path);
         String webAppMount = getWebAppMount();
         WebResourceRoot root = getRoot();
+
+
+        /*
+         * This initializes (when necessary) and checks the jarContents, which
+         * is a highly efficient index of the files stored in the jar. If
+         * jarContents reports that this resource definitely does not contain
+         * the path, we can end this method and move on to the next jar.
+         *
+         * Note: the initialization is thread-safe because multiple simultaneous
+         * threads will create a complete and valid copy, then set the shared
+         * pointer. This guarantees the shared pointer will always go to a
+         * valid object. The cost of multiple copies is small since only one of
+         * them will be long-lived.
+         */
+        try {
+            if ((root.getContext() != null) && (root.getContext().getParent()) != null &&
+                    (((Host) root.getContext().getParent()).getFastClasspathScanning())) {
+
+                if (jarContents == null ||
+                        ((System.currentTimeMillis() - prevJarOpenTime > jarRefreshTime))) {
+
+                    if (!openJarFile().equals(jar)) {
+                        try {
+                            jar = openJarFile();
+                            jarContents = new JarContents(jar);
+                            prevJarOpenTime = System.currentTimeMillis();
+                            closeJarFile();
+                        } catch (IOException ioe) {
+                            throw new RuntimeException("Unable to parse contents of JAR", ioe);
+                        } finally {
+                            closeJarFile();
+                        }
+                    } else {
+                        closeJarFile();
+                    }
+                }
+                if (!jarContents.mightContainResource(path, webAppMount)) {
+                    return new EmptyResource(root, path);
+                }
+            }
+        } catch (Exception e) {
+            log.info(sm.getString("abstractArchiveResourceSet.jarContentWarning") + e.getMessage());
+        }
+
 
         /*
          * Implementation notes
--- a/java/org/apache/catalina/webresources/JarContents.java
+++ b/java/org/apache/catalina/webresources/JarContents.java
@@ -1,0 +1,119 @@
+package org.apache.catalina.webresources;
+
+import java.util.BitSet;
+import java.util.Enumeration;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
+
+/**
+ * This class represents the contents of a jar by determining whether a given
+ * resource <b>might</b> be in the cache, based on a bloom filter. This is not a
+ * general-purpose bloom filter because it contains logic to strip out
+ * characters from the beginning of the key.
+ *
+ * The hash methods are simple but good enough for this purpose.
+ */
+public final class JarContents {
+    private final BitSet bits1;
+    private final BitSet bits2;
+    /**
+     * Constant used by a typical hashing method.
+     */
+    private static final int HASH_PRIME_1 = 31;
+
+    /**
+     * Constant used by a typical hashing method.
+     */
+    private static final int HASH_PRIME_2 = 17;
+
+    /**
+     * Size of the fixed-length bit table. Larger reduces false positives,
+     * smaller saves memory.
+     */
+    private static final int TABLE_SIZE = 2048;
+
+    /**
+     * Parses the passed-in jar and populates the bit array.
+     *
+     * @param jar
+     */
+    public JarContents(JarFile jar) {
+        Enumeration<JarEntry> entries = jar.entries();
+        bits1 = new BitSet(TABLE_SIZE);
+        bits2 = new BitSet(TABLE_SIZE);
+
+        // Enumerations. When will they update this API?!
+        while (entries.hasMoreElements()) {
+            JarEntry entry = entries.nextElement();
+            String name = entry.getName();
+            int startPos = 0;
+
+            // If the path starts with a slash, that's not useful information.
+            // Skipping it increases the significance of our key by
+            // removing an insignificant character.
+            boolean precedingSlash = name.charAt(0) == '/';
+            if (precedingSlash) {
+                startPos = 1;
+            }
+
+            // Find the correct table slot
+            int pathHash1 = hashcode(name, startPos, HASH_PRIME_1);
+            int pathHash2 = hashcode(name, startPos, HASH_PRIME_2);
+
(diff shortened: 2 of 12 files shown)

### CODE REVIEW

File: java/org/apache/catalina/webresources/AbstractArchiveResourceSet.java [hunk 1]
```diff
@@ -211,6 +216,43 @@ public final WebResource getResource(String path) {
...
+         */
+        try {
+            if ((root.getContext() != null) && (root.getContext().getParent()) != null &&
+                    (((Host) root.getContext().getParent()).getFastClasspathScanning())) {
+
+                if (jarContents == null ||
+                        ((System.currentTimeMillis() - prevJarOpenTime > jarRefreshTime) &&
+                            !openJarFile().equals(jar))) {
+                    try {
+                        jar = openJarFile();
```
martin-g, line 240: You call `openJarFile()` twice here. This will increase `org.apache.catalina.webresources.AbstractArchiveResourceSet#archiveUseCount` twice and later it won't be GC-ed by `org.apache.catalina.webresources.AbstractArchiveResourceSet#gc`.
I think you need to call `org.apache.catalina.webresources.AbstractArchiveResourceSet#closeJarFile` after creating the JarContents

martin-g (changes requested):

java/org/apache/catalina/webresources/JarContents.java
```diff
@@ -0,0 +1,144 @@
...
+     * substring, but memory and runtime speed are critical.
+     *
+     * @param content
+     *            Wrapping String.
+     * @param startPos
+     *            First character in the range.
+     * @return hashcode of the range.
+     */
+    private int hashcode2(String content, int startPos) {
+        int h = 7;
```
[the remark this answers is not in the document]
kamnani: Resolved with latest commit

java/org/apache/catalina/webresources/AbstractArchiveResourceSet.java
kamnani: Resolved with latest commit

kamnani:
@martin-g   Thanks for your time on this : )
I have marked all of the conversations as resolved after the latest commits. Do we have any more feedback on this PR?

java/org/apache/catalina/webresources/AbstractArchiveResourceSet.java
```diff
@@ -211,6 +216,46 @@ public final WebResource getResource(String path) {
...
+         * valid object. The cost of multiple copies is small since only one of
+         * them will be long-lived.
+         */
+        try {
+            if ((root.getContext() != null) && (root.getContext().getParent()) != null &&
+                    (((Host) root.getContext().getParent()).getFastClasspathScanning())) {
+
+                if (jarContents == null ||
+                        ((System.currentTimeMillis() - prevJarOpenTime > jarRefreshTime) &&
+                            !openJarFile().equals(jar))) {
```
martin-g, line 238: This will still leak an `opening` of the same jar. If the `equals()` returns `true` then it won't enter the body of the `if` and `closeJarFile()` won't be called.

java/org/apache/catalina/webresources/JarContents.java
```diff
@@ -0,0 +1,122 @@
...
+            // If the path starts with a slash, that's not useful information.
+            // Skipping it increases the significance of our key by
+            // removing an insignificant character.
+            boolean precedingSlash = name.charAt(0) == '/';
+            if (precedingSlash) {
+                startPos = 1;
+            }
+
+            // Find the correct table slot
+            int pathHash1 = hashcode(name, startPos,HASH_PRIME_1);
```
martin-g, line 60: missing space before the third argument

martin-g (changes requested):
Please add unit tests for JarContents class.

java/org/apache/catalina/webresources/AbstractArchiveResourceSet.java
```diff
@@ -211,6 +216,50 @@ public final WebResource getResource(String path) {
...
+                    } else {
+                        closeJarFile();
+                    }
+                }
+                if (!jarContents.mightContainResource(path, webAppMount)) {
+                    return new EmptyResource(root, path);
+                }
+            }
+        } catch (Exception e) {
+            // Do nothing
```
martin-g, line 259: why nothing ? 
IMO it should be logged as WARNING, or at least DEBUG

test/org/apache/catalina/webresources/TestJarContents.java
```diff
@@ -0,0 +1,61 @@
...
+            // If this JAR was in a web application, this is equivalent to how it
+            // would be added
+            JarResourceSet test =
+                    new JarResourceSet(root, "/", jar.getAbsolutePath(), "/META-INF/resources");
+            test.setStaticOnly(true);
+            root.addJarResources(test);
+
+            JarContents testJarContentsObject = new JarContents(new JarFile("test/webresources/dir1.jar"));
+            Assert.assertTrue(testJarContentsObject.mightContainResource(
+                    "/d1/d1-f1.txt", jar.getAbsolutePath()));
```
martin-g, line 53: Please also add test assertions for:
- `null`
- "" (i.e. empty path)
- "/", i.e. the root
- path without leading `/`
- any other possible weird value

test/org/apache/catalina/webresources/TestJarContents.java
```diff
@@ -0,0 +1,61 @@
...
+
+            Assert.assertFalse(testJarContentsObject.mightContainResource(
+                    "/d7/d1-f1.txt", jar.getAbsolutePath()));
+        } catch (Exception e) {
+            Assert.fail("Error Happened while Testing JarContents, " + e);
```
martin-g, line 58: No need of capital letters for `H`appened and `T`esting.

martin-g: Use `e.getMessage()` instead of `e.toString()` (implicitly)

kamnani: Thanks for the suggestions :) 
Resolved with the latest commit.

test/org/apache/catalina/webresources/TestJarContents.java
kamnani: Added more test cases in the latest commit.

java/org/apache/catalina/webresources/AbstractArchiveResourceSet.java
kamnani: Added a logger message to this. Thanks for the suggestions :)

java/org/apache/catalina/webresources/JarContents.java
kamnani: Resolved with the latest commit

java/org/apache/catalina/webresources/AbstractArchiveResourceSet.java
kamnani: Resolved with the latest commit.

java/org/apache/catalina/webresources/JarContents.java
```diff
@@ -0,0 +1,122 @@
...
+
+        if (path.charAt(startPos) == '/') {
+            // ignore leading slash
+            startPos++;
+        }
+
+        // Find the correct table slot
+        int pathHash1 = hashcode(path, startPos, HASH_PRIME_1);
+        int pathHash2 = hashcode(path, startPos, HASH_PRIME_2);
+        boolean probablyPresent = (bits1.get(pathHash1 % TABLE_SIZE) && bits2.get(pathHash2 % TABLE_SIZE));
```
FSchumacher, line 117: Wouldn't it be even faster, when you calculate pathHash2 lazyly, i.e. inlining the two int variables and use the fact, that the `&&` operator is lazy?
The parenthesis seem to be not needed here, too.
And last, why not return the result directly?

kamnani: Thanks for the suggestion @FSchumacher 👍 
I have made this change in the latest commit.

kamnani:
@martin-g Thanks for your valuable feedback. 
I have added the test cases and marked the conversations as resolved after the latest commits. Do we have anything else on this PR?

rmaucher:
I am rather skeptical on this PR overall.
As with your other PR, please remove all the changes to the Host, the flag should be on the Context. Things are then much simpler without any unnecessary API changes.

kamnani:
@rmaucher Thanks for the comment. 
We tested this optimization on an application with 3500+ jars which performed 60,000+ class loads. This sliced off 180+ seconds from the startup time. I am sure this is a really good optimizations for large applications. 
I can add those flags to the Context as you did with the other PR.

rmaucher:
I made the change to use the Context instead of the Host, no issue there.
However, I'm not ok with the proposed changes as is. The problem for me is that I don't understand the weird code added to AbstractArchiveResourceSet. I am ok with adding the creation of that JarContents filter in openJarFile() [even with the sync ...], basically, but not elsewhere, and without the strange custom refresh mechanism on lookup.
Then, since it is *likely* quite harmless, it could be enabled by default (and maybe the flag can be removed altogether ...).

kamnani:
> I made the change to use the Context instead of the Host, no issue there.
> However, I'm not ok with the proposed changes as is. The problem for me is that I don't understand the weird code added to AbstractArchiveResourceSet. I am ok with adding the creation of that JarContents filter in openJarFile() [even with the sync ...], basically, but not elsewhere, and without the strange custom refresh mechanism on lookup.
> Then, since it is _likely_ quite harmless, it could be enabled by default (and maybe the flag can be removed altogether ...).


@rmaucher  I agree for the creation of JarContent in OpenJarFile(), however the crux of the optimization will be when we are checking for the resource in a Jar. So we still need to add this code inside getResource() call, the way its been in the PR after we create the JarContent.

```
                if (!jarContents.mightContainResource(path, webAppMount)) {
                    return new EmptyResource(root, path);
                }
```

For the custom refresh mechanism, we added that with regard to PR #340  where Mark(comment attached below) had asked us about the cache validation mechanism. 

```
I don't see any cache validation meaning updated JARs (with the same name) will be ignored. That is likely to trigger all sorts of unexpected bugs.
```

I would expect OpenJarFile already must be tackling the jar-refresh issue and in that scenario your comment would be the best way to go forward with this. 

Thanks again for your valuable time on this. I look forward for this optimization to get shipped soon :)

rmaucher:
There's no cache here: the AbstractArchiveResourceSet already has a "JarFile archive" field, so it adds a new JarContents field with the same data and both get created and cleared at the same time. Looking back at the other #340 there were problems with the loader/classloader modifications which IMO look like a big hack and have a JAR cache.

kamnani:
> There's no cache here: the AbstractArchiveResourceSet already has a "JarFile archive" field, so it adds a new JarContents field with the same data and both get created and cleared at the same time. Looking back at the other #340 there were problems with the loader/classloader modifications which IMO look like a big hack and have a JAR cache.

I think, your solution does make sense to me. I will still need these code lines in AbstactArchiveResourceSet just to skip checking a jar if the resource is not going to be found there. 

```
   if (!jarContents.mightContainResource(path, webAppMount)) {
                    return new EmptyResource(root, path);
                }
```

rmaucher:
I tried it, with some changes. Will see the feedback about an actual performance benefit. I measured the cost of building the filter, not too big overall so there's likely something.

PR #354, opened 2020-09-04
Optimize Server startup time using multi-threading for annotation scanning

kamnani:
This is a redo of Previous PR: https://github.com/apache/tomcat/pull/349

The following changes have been made based on the suggestions earlier: 
1) Flag (parallelAnnotationScanning) can be passed through Host Configuration. See example below. 
2) By default parallelAnnotationScanning remains disabled and thus default tomcat behavior is untouched.
3) Number of threads could be triggered by startStopThreads parameter. 
3) PR is against master. 

Apologies in case something is missed. 

```
            unpackWARs="true" autoDeploy="true" parallelAnnotationScanning="true" >

### DIFF
--- a/java/org/apache/catalina/startup/ContextConfig.java
+++ b/java/org/apache/catalina/startup/ContextConfig.java
@@ -39,6 +39,10 @@ import java.util.Map.Entry;
 import java.util.Properties;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
+import java.util.concurrent.Callable;
 
 import jakarta.servlet.MultipartConfigElement;
 import jakarta.servlet.ServletContainerInitializer;
@@ -121,7 +125,6 @@ import org.xml.sax.SAXParseException;
 public class ContextConfig implements LifecycleListener {
 
     private static final Log log = LogFactory.getLog(ContextConfig.class);
-
 
     /**
      * The string resources for this package.
@@ -1374,7 +1377,14 @@ public class ContextConfig implements LifecycleListener {
     protected void processClasses(WebXml webXml, Set<WebXml> orderedFragments) {
         // Step 4. Process /WEB-INF/classes for annotations and
         // @HandlesTypes matches
-        Map<String, JavaClassCacheEntry> javaClassCache = new HashMap<>();
+
+        Map<String, JavaClassCacheEntry> javaClassCache;
+
+        if (context.getParent() instanceof Host && ((Host) context.getParent()).isParallelAnnotationScanning()) {
+            javaClassCache = new ConcurrentHashMap<>();
+        } else {
+            javaClassCache = new HashMap<>();
+        }
 
         if (ok) {
             WebResource[] webResources =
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by rmaucher** [comment]
I don't agree with this. There is a "utility" executor (Server.getUtilityExecutor()), it could be used for that instead of adding special purpose threads. Then you can add a flag to do the scanning as executor tasks or the usual way.

**Comment by kamnani** [comment]
Thanks @rmaucher for your comment. I have made the necessary changes as suggested. It would be great if you could verify that change. I have also updated the PR description since the flags have now been changed with this change. 
Is there any other change needed on this PR you want to suggest ?

**Comment by rmaucher** [comment]
I still don't really like some items, like the explicit flag (but I don't see yet how to do it) or some of the "optimizations".
By default the utility thread count is 2, so not too much parallelism.

File: java/org/apache/catalina/startup/ContextConfig.java [hunk 1]
```diff
@@ -122,6 +126,11 @@
 
     private static final Log log = LogFactory.getLog(ContextConfig.class);
 
+    private static final int DEFAULT_CLASS_CACHE_SIZE = 16384;
+
+    private static final float DEFAULT_LOAD_FACTOR = .75f;
+
+    private static int CONCURRENCY_LEVEL = 1;
```
rmaucher, line 133: Ok for the map size, maybe (it seems very large, 16k classes ?), I don't quite see the point of the rest. The concurrency may be up to the utility executor thread count, and I don't think this is an optimization which makes any difference.

java/org/apache/catalina/mbeans/MBeanFactory.java
```diff
@@ -490,6 +491,7 @@ public String createStandardContext(String parent,
     public String createStandardHost(String parent, String name,
                                      String appBase,
                                      boolean autoDeploy,
+                                     boolean parallelAnnotationScanning,
```
rmaucher, line 494: I disagree with this API change. IMO this new feature does not justify it.

java/org/apache/catalina/startup/ContextConfig.java
```diff
@@ -1374,7 +1383,19 @@ protected void webConfig() {
...
+        if (context.getParent() instanceof Host) {
+           Host host = (Host) context.getParent();
+            Container container = host.getParent();
+            CONCURRENCY_LEVEL = container.getStartStopThreads();
+            javaClassCache = new ConcurrentHashMap<>(DEFAULT_CLASS_CACHE_SIZE, DEFAULT_LOAD_FACTOR,
+                    CONCURRENCY_LEVEL);
+        } else {
+            javaClassCache = new ConcurrentHashMap<>(DEFAULT_CLASS_CACHE_SIZE, DEFAULT_LOAD_FACTOR,
+                    CONCURRENCY_LEVEL);
+        }
```
rmaucher, line 1398: Well, if it's changed to be always a ConcurrentHashMap, it probably won't be a measurable performance difference.

java/org/apache/catalina/startup/ContextConfig.java
```diff
@@ -2136,26 +2157,98 @@ protected InputSource getWebXmlSource(String filename, boolean global) {
...
+            for (WebXml fragment : fragments) {
+                Callable<Void> task = new AnnotationScanTask(fragment, handlesTypesOnly, javaClassCache);
+                futures.add(pool.submit(task));
+            }
+            try {
+                for (Future<?> future : futures) {
+                    future.get();
+                }
+            } catch (Exception e) {
+                throw new RuntimeException("Parallel execution failed", e);
```
rmaucher, line 2246: Localization missing.

rmaucher (commented):
I will have to test and run the whole thing too and see if it's possible to simplify some more.

kamnani:
> I still don't really like some items, like the explicit flag (but I don't see yet how to do it) or some of the "optimizations".
> By default the utility thread count is 2, so not too much parallelism.

Yes, You're right with the default Value, but that can be changed through the Host Configuration.. right?

java/org/apache/catalina/startup/ContextConfig.java
kamnani: Resolved with the latest changes.

java/org/apache/catalina/startup/ContextConfig.java
kamnani: Resolved with latest commit.

java/org/apache/catalina/startup/ContextConfig.java
kamnani: Have added the default behavior without the flag in the latest commit.

kamnani:
@rmaucher I actually tested the optimization with the flag in a large application having 1000's of Jar in the classpath and the server startup gains are significant and the stats are as follows : 
2816 classes -  44% improvement 
1800 classes - 40% improvement 
1000 classes - 40% improvement

The default tomcat app shows an improvement of 6% with the flag.

java/org/apache/catalina/mbeans/MBeanFactory.java
kamnani: Thanks @rmaucher for the comment. 
I have little experience working on Tomcat so if you can you explain how this becomes irrelevant that would be great. Also, do we not want the API's to be able to pass this flag when the optimizations are quite significant?

rmaucher: This is not enough to justify changing this API. You can create the host, then set the flag just after. Good enough.

kamnani: Alright. I have removed these changes. 
Resolved with the latest commit. 
Thanks 👍

java/org/apache/catalina/startup/LocalStrings_fr.properties
```diff
@@ -69,6 +69,7 @@ contextConfig.missingRealm=Aucun royaume (realm) n'a été configuré pour réal
 contextConfig.processAnnotationsDir.debug=Balayage du répertoire pour trouver des fichiers de classe avec annotations [{0}]
 contextConfig.processAnnotationsJar.debug=Analyse du fichier jars pour des classes annotées avec [{0}]
 contextConfig.processAnnotationsWebDir.debug=Balayage du répertoire d''applications web, pour fichiers de classe avec annotations [{0}]
+contextConfig.processAnnotationsInParallelFailure=exécution parallèle a échoué
```
martin-g, line 72: Should `exécution` be capital cased like the other messages ?

java/org/apache/catalina/startup/ContextConfig.java
```diff
@@ -2136,26 +2146,98 @@ protected InputSource getWebXmlSource(String filename, boolean global) {
...
+            try {
+                for (Future<?> future : futures) {
+                    future.get();
+                }
+            } catch (Exception e) {
+                throw new RuntimeException(sm.getString("contextConfig.processAnnotationsInParallelFailure"), e);
+            }
+        } finally {
+            if (pool != null) {
+                pool.shutdownNow();
```
martin-g, line 2239: This is not really needed because org.apache.tomcat.util.threads.ScheduledThreadPoolExecutor#shutdown methods do nothing, but anyway it is a good practice to do it.

java/org/apache/catalina/startup/ContextConfig.java
```diff
@@ -2136,26 +2146,98 @@ protected InputSource getWebXmlSource(String filename, boolean global) {
...
+    }
+
+    /**
+     * Executable task to scan a segment for annotations. Each task does the
+     * same work as the for loop inside processAnnotations();
+     *
+     * @author Engebretson, John
+     * @author Kamnani, Jatin
+     */
+    private class AnnotationScanTask implements Callable<Void> {
```
martin-g, line 2190: Why `Callable<Void>` ? 
You can use `Runnable`

java/org/apache/catalina/startup/LocalStrings_fr.properties
rmaucher: Yes, the translation is meh. As a rule, I'd say never commit any localization IMO, just let others handle in with poeditor later.

java/org/apache/catalina/startup/ContextConfig.java
kamnani: Thanks for the suggestion. 
Resolved with the latest commit.

java/org/apache/catalina/startup/LocalStrings_fr.properties
kamnani: Thanks for the suggestion. 
Resolved with the latest commit.

martin-g (approved):
LGTM!

rmaucher:
Since the release cycle is done, I'll work on this feature.

rmaucher:
I merged the PR with some changes and a minor fix. The flag was moved to the Context since that's where it belongs, and it avoids the unnecessary API changes.

kamnani:
Thank you @rmaucher @martin-g to get this moving fast 👍 . 
Will this be backported to version 9 ?

pbodnar:
> Will this be backported to version 9 ?

Actually, it seems like this was merged right into the version 9 branch - see 101476c3536ac09a105d0b603b2c51dd0f3770e9. So it is available since Tomcat 9.0.39.

I can confirm about 1/3 scanning speed improvement with `parallelAnnotationScanning="true"` on a `<context>`. Thank you, guys. :)

PR #355, opened 2020-09-09
[perf] Lazy init ELContext#listeners to avoid unnecessary ArrayList and ArrayList$Itr

tandraschko:
in case of a JSF view, this methods are called quite often and creates ArrayList and ArrayList$Itr instances, also if no listener is added.
This reduces the created objects of a simple view demo view (-54k objects in 400 requests)

We could even improve it a bit by using a index-loop, which also avoids a ArrayList$Itr instance, if listeners are available. Not sure if Tomcat maintainers likes such changes.

It would also be great if you could port it to older branches

### DIFF
--- a/java/jakarta/el/ELContext.java
+++ b/java/jakarta/el/ELContext.java
@@ -35,7 +35,7 @@ public abstract class ELContext {
 
     private ImportHandler importHandler = null;
 
-    private List<EvaluationListener> listeners = new ArrayList<>();
+    private List<EvaluationListener> listeners;
 
     private Deque<Map<String,Object>> lambdaArguments = new LinkedList<>();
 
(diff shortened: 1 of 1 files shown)


java/jakarta/el/ELContext.java
```diff
@@ -139,6 +140,10 @@ public void setLocale(Locale locale) {
      * @since EL 3.0
      */
     public void addEvaluationListener(EvaluationListener listener) {
+        if (listeners == null) {
```
martin-g, line 143: Is it possible that this method is called by multiple threads ? In that case you will need to add locking here.

tandraschko: i dont think so. At least in JSF and JSP it should be used "request scoped". Also currently it doesnt use concurrent list impl.

PR #356, opened 2020-09-10
Micro optimization: replace "".equals(a) with a.isEmpty()

martin-g:
Blog post with microbenchmarks: https://medium.com/javarevisited/micro-optimizations-in-java-string-equals-22be19fd8416
Proposal for similar change in JDK: https://github.com/openjdk/jdk/pull/29


stokito:
In fact Tomcat has hundreds of such places where simple, safe, and more readable optimizations can be applied. Just open sources in IntelliJ, open Settings/Editor/Inspections/Java/Performarmance and enable all and then run Analyze.
Most hot places are unefficient working with maps like containsKey() then get() call and collections ans StringBuilder without initial capacity.
It would be great just to fix all these problems at once and this will give a real impact on performance

martin-g:
> It would be great just to fix all these problems at once

I prefer to make one kind of change at a time. It is easier to detect regressions and impact this way.

PR #358, opened 2020-09-10
Fix 64735 ServletContext.addJspFile() always fails with SecurityManager

stiemannkj1:
Fix for: https://bz.apache.org/bugzilla/show_bug.cgi?id=64735 9.0.x

### DIFF
--- a/java/org/apache/catalina/core/ApplicationContextFacade.java
+++ b/java/org/apache/catalina/core/ApplicationContextFacade.java
@@ -117,6 +117,7 @@ public class ApplicationContextFacade implements ServletContext {
         classCache.put("getAttribute", clazz);
         classCache.put("log", clazz);
         classCache.put("setSessionTrackingModes", new Class[]{Set.class} );
+        classCache.put("addJspFile", new Class[]{String.class, String.class});
     }
 
 


rmaucher:
Well, there's a problem: more methods are affected too. It would seem all from Servlet 4, plus maybe one from Servlet 3 (declareRoles).

stiemannkj1:
@rmaucher, I think you are right about `declareRoles` and Servlet 4.0 methods. However, I think only the following methods would be broken:

* `doPrivileged("declareRoles", new Object[]{roleNames});`
* `doPrivileged("setSessionTimeout", new Object[] { Integer.valueOf(sessionTimeout) });`
* `doPrivileged("setRequestCharacterEncoding", new Object[] { encoding });`
* `doPrivileged("setResponseCharacterEncoding", new Object[] { encoding });`

Other methods that have no args won't have this issues, and anything that calls the 3-arg `doPrivileged` method won't have this issue.

Maybe I'll add some tests and fix the other methods as well.

stiemannkj1:
Closing this PR. I'll add tests and fix the other `SecurityManager` issues in this class and resend.

stiemannkj1:
@rmaucher, updated PR fixing all the broken methods and adding tests: https://github.com/apache/tomcat/pull/360

PR #359, opened 2020-09-13
Fix 64735 ServletContext.addJspFile() always fails with SecurityManager

stiemannkj1:


markt-asf:
Many thanks for this. I particularly like the way you implemented the unit test. I am currently leaning towards packaging the test slightly differently so I'll apply this manually rather than just clicking merge.

markt-asf:
Thanks. Applied manually so I could make a few minor tweaks (mainly for consistency with existing tests). Also back-ported to 8.5.x and 7.0.x,

PR #361, opened 2020-09-13
Add a GitHub action to lint the Markdown and YAML files.

jbampton:
Add a basic `markdownlint` config file.
Add a `yamllint` config file.
Lint both Markdown and YAML files.


.github/workflows/lint.yml
```diff
@@ -0,0 +1,18 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - name: yaml-lint
+        uses: ibiqlik/action-yamllint@v1
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: '12.x'
+      - run: npm install -g markdownlint-cli@0.23.2
+      - run: markdownlint '**/*.md' --ignore node_modules
```
martin-g, line 18: Is it possible to pass the config file as an argument ?
The idea is to move `.markdownlint.yml` from the root folder to be next to `lint.yml`.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,18 @@
...
+on: [push, pull_request]
+
+jobs:
+  lint:
+    name: 🧹 Markdown and YAML Lint
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: yaml-lint
+        uses: ibiqlik/action-yamllint@v1
```
martin-g, line 12: Please move `yamllint.yml` from the root folder to `.github/workflows/` and use [config_file](https://github.com/ibiqlik/action-yamllint#optional-parameters) to specify it.
The idea is to not pollute the root folder with too many config files which are used for very specific needs.

jbampton:
Hey @martin-g I have moved both config files to the `.github` folder. 

If we put them in the `.github/workflows/` folder, GitHub will think they are both workflows.

martin-g (approved):

martin-g:
Thanks! Looks better now!
Let's see what others think about this linters.

jbampton:
Hey @markt-asf what do you think about this PR ?

markt-asf:
A couple of those files appear to be missing ALv2 headers. On that topic, it looks like the markdown files need an ALv2 header if possible as well.

jbampton:
Hey @markt-asf I have rebased and moved the two config files back to the repo root.

The 2 files are not specific to GitHub so say we moved to GitLab we wouldn't want our whole repo config files in a GitHub folder.

I have also now added the licenses to all new files.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,36 @@
...
+      - master
+
+jobs:
+  lint:
+    name: 🧹 Markdown and YAML
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: yaml-lint
+        uses: ibiqlik/action-yamllint@v3
```
jbampton, line 30: YAML Lint is not certified by GitHub. It is provided by a third-party and is governed by separate terms of service, privacy policy, and support documentation.  The new ASF rules mean we can't use third party scripts for GitHub Actions.  https://github.com/marketplace/actions/yaml-lint

This YAML check with `yamllint` can be refactored to run on the official Python GitHub action by installing and running yamllint with pip.

.github/workflows/lint.yml
```diff
@@ -0,0 +1,36 @@
...
+
+name: Lint
+
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
jbampton, line 21: Linters should probably only rull on pull request to speed up build times.  
Linting is a more minor test etc.

PR #363, opened 2020-09-14
Fix Chinese translation of word "war"

reliveyy:
This PR fixes Chinese translation of word "war". The Chinese word "战争" means the real war not a web archive.


bzhaoopenstack (commented):
Looks good to me. That's the right translation. Thanks.

martin-g (approved):
A Chinese friend of mine confirmed that this is better!
Thanks, @reliveyy !

markt-asf:
Note to Tomcat committers. If you merge PRs like this you need to be careful because poeditor.com is the master copy for translations. You need to:
- import the latest translations from poeditor
- apply the PR
- export the translations to poeditor (and check it only changes translations affected by the PR)
In some cases it may be easier to manually apply the changes in poeditor and import.

markt-asf:
Thanks. Merged. Please note that our preferred route for translation updates is: https://poeditor.com/projects/view?id=221603

PR #364, opened 2020-09-15
Fix case of HTML and SHTML

jbampton:


michael-o (approved):

martin-g:
Thank you, @jbampton !

PR #366, opened 2020-09-17
Change exception from Throwable to IOException for ServerInfo.java

nicholasgribanov:
In this section may produce only IOException. Is not necessarily use Throwable there are.


markt-asf:
That code is deliberately defensive. I can see the possibility for at least `IllegalArgumentException` and possibly `NullPointerException` (I haven't checked in the JRE source to be sure of that one).

PR #368, opened 2020-09-29
fix: CachedResource add path length

blueSky1825821:
cachedResource should add webapp path length

### DIFF
--- a/java/org/apache/catalina/webresources/CachedResource.java
+++ b/java/org/apache/catalina/webresources/CachedResource.java
@@ -412,10 +412,15 @@ public class CachedResource implements WebResource {
     // Assume that the cache entry will always include the content unless the
     // resource content is larger than objectMaxSizeBytes. This isn't always the
     // case but it makes tracking the current cache size easier.
+    // The entry size should contain the webapp resource path and contents,
+    // because cached resources can contain very, very long error paths,
+    // which can cause the JVM to fail in garbage cleanup.
     long getSize() {
         long result = CACHE_ENTRY_SIZE;
-        if (getContentLength() <= objectMaxSizeBytes) {
-            result += getContentLength();
+        long webappPathLength = getWebappPath().getBytes().length;
+        long size = getContentLength() + webappPathLength;
+        if (size <= objectMaxSizeBytes) {
+            result += size;
         }
         return result;
     }


markt-asf:
What is the justification for this change? What problem does it solve?

blueSky1825821:
> What is the justification for this change? What problem does it solve?

When our application is loaded in another path or attacked in an unknown way, webapp path could take up a lot of space.
org.apache.catalina.webresources.Cache#resourceCache can't be cleaned up, this happens in a production environment

blueSky1825821:
> > What is the justification for this change? What problem does it solve?
> 
> When our application is loaded in another path or attacked in an unknown way, webapp path could take up a lot of space(a few megabytes).
> org.apache.catalina.webresources.Cache#resourceCache can't be cleaned up, this happens in a production environment

markt-asf:
I've done some calculations and the absolute upper limit (the actual upper limit will be lower) is that a cache with a 10MB limit on size will actually use a little under 175MB. That isn't great but it isn't terrible either. I'll tweak the resource size calculation so it includes an approximate of the memory required to store the path.
The proposed patch has a couple of issues:

- It converts the string to bytes to determine the size.  This is both expensive and wrong. String length *2 +38 is a much closer approximation and I intend to ignore the 38 as that is already accounted for in the 500 byte overhead for each entry.
- It includes the path size when determining if the resource exceeds the max object size (and hence won't have the content cached) but does not include the path size when performing the same check to determine when deciding whether to cache the content or not.

I should have a fix for this committed shortly.

markt-asf:
Fixed in:
- master for 10.0.0-M9 onwards
- 9.0.x for 9.0.39 onwards
- 8.5.x for 8.5.59 onwards

blueSky1825821:
> I've done some calculations and the absolute upper limit (the actual upper limit will be lower) is that a cache with a 10MB limit on size will actually use a little under 175MB. That isn't great but it isn't terrible either. I'll tweak the resource size calculation so it includes an approximate of the memory required to store the path.
> The proposed patch has a couple of issues:
> 
> * It converts the string to bytes to determine the size.  This is both expensive and wrong. String length *2 +38 is a much closer approximation and I intend to ignore the 38 as that is already accounted for in the 500 byte overhead for each entry.
> * It includes the path size when determining if the resource exceeds the max object size (and hence won't have the content cached) but does not include the path size when performing the same check to determine when deciding whether to cache the content or not.
> 
> I should have a fix for this committed shortly.

Let me add that, after jdk8, e.g jdk9 String has been changed
private final char value[]; -->  private final byte[] value; 
but length *2 is also fine

PR #370, opened 2020-10-12
Polish LocalStrings_zh_CN.properties

ijliym:
Remove unexpected `(`.


markt-asf:
Thanks for the PR. The Tomcat project manages i18n files via poeditor.com I've made these changes there (as well as several others I found) and I'll import them shortly.

PR #372, opened 2020-10-22
Update FileHandler.java

dao-jun:
remove static keyword from  `DELETE_FILES_SERVICE` to avoid class `ThreadPoolExecutor` loaded earlier than agent start.
It can cause agent enhance  class `ThreadPoolExecutor` failed.

### DIFF
--- a/java/org/apache/juli/FileHandler.java
+++ b/java/org/apache/juli/FileHandler.java
@@ -100,7 +100,7 @@ public class FileHandler extends Handler {
     public static final int DEFAULT_BUFFER_SIZE = -1;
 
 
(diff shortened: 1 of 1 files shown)


markt-asf:
The DELETE_FILES_SERVICE executor is intended to be a singleton. This PR breaks that and is, therefore, unlikely to be applied.

dao-jun:
> The DELETE_FILES_SERVICE executor is intended to be a singleton. This PR breaks that and is, therefore, unlikely to be applied.

Yep, I forgot that. The pr will be closed and I would use inner class to keep the field singleton.

PR #373, opened 2020-10-22
lazy set filed `DELETE_FILES_SERVICE` 

dao-jun:
Lazy set field `DELETE_FILES_SERVICE` to avoid class `ThreadPoolExecutor` loaded earlier than tomcat start.

Actually, some `APM` systems enhance class `ThreadPoolExecutor` with `java agent`,  the field `DELETE_FILES_SERVICE` may cause enhance failed.

### DIFF
--- a/java/org/apache/juli/FileHandler.java
+++ b/java/org/apache/juli/FileHandler.java
@@ -18,37 +18,20 @@
 
 package org.apache.juli;
 
-import java.io.BufferedOutputStream;
-import java.io.File;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.OutputStream;
-import java.io.OutputStreamWriter;
-import java.io.PrintWriter;
-import java.io.UnsupportedEncodingException;
+import java.io.*;
 import java.nio.file.DirectoryStream;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.security.AccessController;
-import java.security.PrivilegedAction;
 import java.sql.Timestamp;
 import java.time.DateTimeException;
 import java.time.LocalDate;
 import java.time.format.DateTimeFormatter;
 import java.time.temporal.ChronoUnit;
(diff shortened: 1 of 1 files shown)


martin-g:
Please provide more information what kind of issue you have now!
The description is a bit vague at the moment.

dao-jun:
> Please provide more information what kind of issue you have now!
> The description is a bit vague at the moment.

  We use `java agent` to enhance class `ThreadPoolExecutor` to provide `cross thread trace monitor ability`, but we found that class `ThreadPoolExecutor`  loaded before agent start, it caused by class `AsyncFileHandler` which defined in `logging.properties`.  So that `java agent` could not enhance class `ThreadPoolExecutor`.  Event we replaced `org.apache.juli.AsyncFileHandler` with `java.util.logging.FileHandler`  to avoid the issue, but I want to fix it.

markt-asf:
How does this fix the issue? It only delays the creation of the executor from the point the class is loaded to the point where the first instance is created. Instance creation is also triggered by the parsing of logging.properties and occurs immediately the class is loaded. The class loading is triggered by the instance creation so I don't see how this proposed PR can help.

dao-jun:
> How does this fix the issue? It only delays the creation of the executor from the point the class is loaded to the point where the first instance is created. Instance creation is also triggered by the parsing of logging.properties and occurs immediately the class is loaded. The class loading is triggered by the instance creation so I don't see how this proposed PR can help.

Yes, I just noticed it create an instance in `ClassLoaderLogManager#readConfiguration` by invoking `classLoader.loadClass(handlerClassName).getConstructor().newInstance()`, I'm sorry for that.     How about using a single `Thread` instead of `ThreadPoolExecutor` ?

markt-asf:
Switching a single thread would just require the re-implementation of a chunk of the Executor functionality.
One option would be to move lazy init to clean() and then configure maxDays=0 if you want to use your agent.

PR #376, opened 2020-11-05
Implement Bug 64877

pmouawad:
Hello,
First thanks for your great product I've been using for years.

Recently, I worked on a project that uses JDBCStore to store sessions.
I had to run performance tests on it.
Those tests revealed a high contention point in this component due to the coarse granularity of synchronization. 
The impact was that 1 thread could block hundreds of others leading to high response time at higher scale.

So I implemented (based on existing one) a new one with following differences:

- I kept only the dataSourceName based configuration
- I removed stateful instance variables in order to be able to lift synchronization

This new implementation improved drastically performances.

I am aware that there is duplication with JDBCStore that could be improved but it's a first shot and I lack of time unfortunately.

I hope you can take into account this contribution.

The associated bugzilla enhancement is:

- https://bz.apache.org/bugzilla/show_bug.cgi?id=64877

Regards
Philippe M.

### DIFF
--- a/java/org/apache/catalina/session/DataSourceJDBCStore.java
+++ b/java/org/apache/catalina/session/DataSourceJDBCStore.java
@@ -1,0 +1,808 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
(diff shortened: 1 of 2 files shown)


rmaucher:
At least for Tomcat 10, modifying the existing realm is the way to go. For example, the JDBCRealm was also removed over DataSourceRealm for similar reasons.

rmaucher:
I meged the PR, sorf of, in Tomcat 10. Thanks ! Also added in 9.0 and 8.5.

pmouawad:
Thanks Remy !

PR #377, opened 2020-11-13
Set "Secure" session cookie attribute if called under SSL.

akurth:
Missing "Secure" flag may lead to redirect loops under certain
conditions, e.g. when sameSiteCookies=none has been set at the
CookieProcessor and browser is Chrome.

Fixes <https://bz.apache.org/bugzilla/show_bug.cgi?id=64921>.

### DIFF
--- a/test/org/apache/catalina/valves/TestLoadBalancerDrainingValve.java
+++ b/test/org/apache/catalina/valves/TestLoadBalancerDrainingValve.java
@@ -238,6 +238,8 @@ public class TestLoadBalancerDrainingValve {
                 expectedCookie.setPath(cookieConfig.getPath());
                 expectedCookie.setMaxAge(0);
 
+                EasyMock.expect(request.isSecure()).andReturn(true);
+
                 // These two lines just mean EasyMock.expect(response.addCookie) but for a void method
                 response.addCookie(expectedCookie);
                 EasyMock.expect(ctx.getSessionCookieName()).andReturn(sessionCookieName); // Indirect call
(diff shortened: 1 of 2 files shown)


test/org/apache/catalina/valves/TestLoadBalancerDrainingValve.java
```diff
@@ -238,6 +238,8 @@ private void runValve(String jkActivation,
                 expectedCookie.setPath(cookieConfig.getPath());
                 expectedCookie.setMaxAge(0);
 
+                EasyMock.expect(request.isSecure()).andReturn(true);
```
ChristopherSchultz, line 241: Should this be `expect(request.isSecure() || cookieConfig.isSecure())`?

akurth: Feel free to add it. Makes no difference though, since || is short-circuit.

martin-g: Actually, I think we need a second test method to have coverage for the `cookieConfig.isSecure()` case. E.g. `runValve()` could have an additional boolean/enum parameter that decides whether to prepare `request.isSecure()` or `cookieConfig.isSecure()`.

In addition I don't see any new code that verifies that the `expectedCookie` has `isSecure() == true`

markt-asf:
Thanks for the PR. I have implemented a variation of this in 10.0.x and will back-port to 9.0.x and 8.5.x.

PR #378, opened 2020-11-19
use a single `Thread` to do `async-delete-file` works  instead of `ThreadPoolExecutor`

dao-jun:
see: https://github.com/apache/tomcat/pull/373


java/org/apache/juli/FileHandler.java
```diff
@@ -18,37 +18,20 @@
...
 
-import java.io.BufferedOutputStream;
-import java.io.File;
-import java.io.FileOutputStream;
-import java.io.IOException;
-import java.io.OutputStream;
-import java.io.OutputStreamWriter;
-import java.io.PrintWriter;
-import java.io.UnsupportedEncodingException;
+import java.io.*;
```
martin-g, line 21: Please don't use wildcard imports

java/org/apache/juli/FileHandler.java
```diff
@@ -58,98 +41,48 @@
...
+ * Default value: <code>ALL</code></li>
+ * <li><code>filter</code> - The <code>java.util.logging.Filter</code>
+ * implementation class name for this Handler. Default value: unset</li>
+ * <li><code>formatter</code> - The <code>java.util.logging.Formatter</code>
+ * implementation class name for this Handler. Default value:
+ * <code>java.util.logging.SimpleFormatter</code></li>
+ * <li><code>maxDays</code> - The maximum number of days to keep the log
+ * files. If the specified value is <code>&lt;=0</code> then the log files
+ * will be kept on the file system forever, otherwise they will be kept the
+ * specified maximum days. Default value: <code>-1</code>.</li>
```
martin-g, line 77: Whitespace changes which are not really needed.

java/org/apache/juli/FileHandler.java
```diff
@@ -585,4 +517,34 @@ private String obtainDateFromPath(Path path) {
...
+        private BlockingQueue<Runnable> tasks;
+
+        AsyncTaskExecutor() {
+            this.tasks = new LinkedBlockingQueue<>();
+            this.thread = new Thread(() -> {
+                try {
+                    this.tasks.take().run();
+                } catch (Exception e) {
+                    System.err.println("take task failed");
+                    e.printStackTrace();
```
martin-g, line 534: Please use a Logger

java/org/apache/juli/FileHandler.java
```diff
@@ -585,4 +517,34 @@ private String obtainDateFromPath(Path path) {
...
+
+    static class AsyncTaskExecutor {
+
+        private Thread thread;
+        private BlockingQueue<Runnable> tasks;
+
+        AsyncTaskExecutor() {
+            this.tasks = new LinkedBlockingQueue<>();
+            this.thread = new Thread(() -> {
+                try {
```
martin-g, line 530: This should probably be wrapped in a `while (true)` loop. At the moment it will execute just the first task and finish.

java/org/apache/juli/FileHandler.java
```diff
@@ -585,4 +517,34 @@ private String obtainDateFromPath(Path path) {
...
+                    this.tasks.take().run();
+                } catch (Exception e) {
+                    System.err.println("take task failed");
+                    e.printStackTrace();
+                }
+            });
+            this.thread.setName("tomcat-async-delete-file");
+            this.thread.start();
```
martin-g, line 538: The thread should be set as a daemon, or the class should be a lifecycle listener and stop the thread when Tomcat shuts down.

martin-g (changes requested):
The SecurityManager related code is dropped.

The commits should be squashed. No need of the old commits in this PR.

PR #380, opened 2020-11-26
Fix BZ 64110 - request attr for requested ciphers and protocols

markt-asf:
https://bz.apache.org/bugzilla/show_bug.cgi?id=64110

I'm providing a PR for this one as the additional attributes to add some additional overhead. The overhead is limited to keeping two lists of strings: one of requested protocol names and one of requested cipher names. In both cases the names are already in memory so the footprint will be limited to the lists and array of object references. The exceptions are the relatively new GREASE values (see RFC 8701). I don't think they'll cause too much overhead (and there are ways to workaround that if they do - namely do exactly what RFC 8701 doesn't want us to do and enumerate them).

### DIFF
--- a/java/org/apache/catalina/util/TLSUtil.java
+++ b/java/org/apache/catalina/util/TLSUtil.java
@@ -33,11 +33,18 @@ public class TLSUtil {
      *         information, otherwise {@code false}
      */
     public static boolean isTLSRequestAttribute(String name) {
-        return Globals.CERTIFICATES_ATTR.equals(name) ||
-                Globals.CIPHER_SUITE_ATTR.equals(name) ||
-                Globals.KEY_SIZE_ATTR.equals(name)  ||
-                Globals.SSL_SESSION_ID_ATTR.equals(name) ||
-                Globals.SSL_SESSION_MGR_ATTR.equals(name) ||
-                SSLSupport.PROTOCOL_VERSION_KEY.equals(name);
+        switch (name) {
+            case Globals.CERTIFICATES_ATTR:
+            case Globals.CIPHER_SUITE_ATTR:
+            case Globals.KEY_SIZE_ATTR:
+            case Globals.SSL_SESSION_ID_ATTR:
+            case Globals.SSL_SESSION_MGR_ATTR:
+            case SSLSupport.PROTOCOL_VERSION_KEY:
+            case SSLSupport.REQUESTED_PROTOCOL_VERSIONS_KEY:
+            case SSLSupport.REQUESTED_CIPHERS_KEY:
+                return true;
+            default:
+                return false;
+        }
     }
 }
--- a/java/org/apache/tomcat/util/net/SecureNio2Channel.java
+++ b/java/org/apache/tomcat/util/net/SecureNio2Channel.java
@@ -23,7 +23,9 @@ import java.nio.channels.AsynchronousSocketChannel;
 import java.nio.channels.CompletionHandler;
 import java.nio.channels.WritePendingException;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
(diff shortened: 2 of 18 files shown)


java/org/apache/catalina/util/TLSUtil.java
```diff
@@ -38,6 +38,8 @@ public static boolean isTLSRequestAttribute(String name) {
                 Globals.KEY_SIZE_ATTR.equals(name)  ||
                 Globals.SSL_SESSION_ID_ATTR.equals(name) ||
                 Globals.SSL_SESSION_MGR_ATTR.equals(name) ||
-                SSLSupport.PROTOCOL_VERSION_KEY.equals(name);
+                SSLSupport.PROTOCOL_VERSION_KEY.equals(name) ||
+                SSLSupport.REQUESTED_PROTOCOL_VERSIONS_KEY.equals(name) ||
+                SSLSupport.REQUESTED_CIPHERS_KEY.equals(name);
```
martin-g, line 43: I think this would look and read nicer if we use `switch (String)`. Extra benefit: it will be faster!

java/org/apache/tomcat/util/net/SecureNioChannel.java
```diff
@@ -68,6 +71,8 @@
     protected boolean closed = false;
     protected boolean closing = false;
 
+    private Map<String,List<String>> additionalTlsAttributes = new HashMap<>();
```
martin-g, line 74: `final`

java/org/apache/tomcat/util/net/SecureNio2Channel.java
```diff
@@ -70,6 +73,8 @@
     protected boolean closed;
     protected boolean closing;
 
+    private Map<String,List<String>> additionalTlsAttributes = new HashMap<>();
```
martin-g, line 76: could be `final`

martin-g: I see there are some `volatile` member fields in this class.  Should `sniComplete` be `volatile` as well ?

java/org/apache/tomcat/util/net/TLSClientHelloExtractor.java
```diff
@@ -193,6 +214,15 @@ public String getSNIValue() {
     }
 
 
+    public List<String> getClientRequestedCipherNames() {
+        if (result == ExtractorResult.COMPLETE || result == ExtractorResult.NOT_PRESENT) {
+            return clientRequestedCipherNames;
+        } else {
+            throw new IllegalStateException();
```
martin-g, line 221: Intentionally without a message ?

ChristopherSchultz: -1 to having no message

java/org/apache/tomcat/util/net/SecureNio2Channel.java
ChristopherSchultz: Is there a reason to use a Map of String-Lists here, instead of having a more concrete object with well-defined members? Hash maps are quick, but since we know the "names" of these things in advance, why not use a custom class for this purpose?

java/org/apache/catalina/util/TLSUtil.java
markt-asf: Will do.

java/org/apache/tomcat/util/net/SecureNioChannel.java
markt-asf: Will do

java/org/apache/tomcat/util/net/SecureNio2Channel.java
markt-asf: Yes, sniComplete should be volatile.

The reason to use Map<String,List<String>> is that the attribute is exposed to user code so I don;t want it to be an internal Tomcat class. If we get this added to the spec then we could define a class for it in the spec.

java/org/apache/tomcat/util/net/TLSClientHelloExtractor.java
markt-asf: I'll add a message. Force push to follow shortly.

java/org/apache/tomcat/util/net/SecureNio2Channel.java
ChristopherSchultz: 👍

martin-g (approved):

PR #383, opened 2020-12-02
Fix so ResponseUtil:addVaryFieldName() stop adding duplicate values

shorinji:
Currently addVaryFieldName() adds a value twice when there already is named value (not asterisk).
The existing tests did not catch this, since they assert using a HashMap thus containing no duplicates.
I have updated one of the test case, inlining parts of the doTestAddVaryFieldName() that can be reused.

Adding `System.out.println("adding " + name + " with " + adapter.getHeaders(VARY_HEADER) + " already present = " + varyHeader);` at the end of addVaryFieldName() we can track what happens when running the tests:

    test-nio:
    [junit] Running org.apache.tomcat.util.http.TestResponseUtil
    [junit] adding too with [foo, bar] already present = too,bar,too,foo
    [junit] adding too with [{{{, bar] already present = too,bar,too
    [junit] adding foo with [foo, bar] already present = foo,bar,foo
    [junit] adding bar with [{{{, bar] already present = bar,bar
    [junit] adding too with [foo, bar] already present = too,bar,too,foo
    [junit] adding too with [foo, bar] already present = too,bar,too,foo
    [junit] adding foo with [foo, bar] already present = foo,bar,foo
    ...

After applying my patch, this looks like:

    test-nio:
    [junit] Running org.apache.tomcat.util.http.TestResponseUtil
    [junit] adding too with [foo, bar] already present = bar,too,foo
    [junit] adding too with [{{{, bar] already present = bar,too
    [junit] adding foo with [foo, bar] already present = bar,foo
    [junit] adding bar with [{{{, bar] already present = bar
    [junit] adding too with [foo, bar] already present = bar,too,foo
    [junit] adding too with [foo, bar] already present = bar,too,foo
    [junit] adding foo with [foo, bar] already present = bar,foo

Perhaps a better change to the tests would be to rewrite doTestAddVaryFieldName() so none of the tests use HashSet. 
I assume the order of fields listed as Vary should not matter. I don't think the rfc mentions either order or what duplicates should mean.

### DIFF
--- a/java/org/apache/tomcat/util/http/ResponseUtil.java
+++ b/java/org/apache/tomcat/util/http/ResponseUtil.java
@@ -95,13 +95,15 @@ public class ResponseUtil {
 
         // Build single header to replace current multiple headers
         // Replace existing header(s) to ensure any invalid values are removed
-        fieldNames.add(name);
-        StringBuilder varyHeader = new StringBuilder();
-        varyHeader.append(name);
-        for (String fieldName : fieldNames) {
-            varyHeader.append(',');
-            varyHeader.append(fieldName);
+        if (!fieldNames.contains(name)) {
+            fieldNames.add(name);
         }
+        StringBuilder varyHeader = new StringBuilder();
+        for (String fieldName : fieldNames) {
+            varyHeader.append(fieldName);
+            varyHeader.append(',');
+        }
+        varyHeader.deleteCharAt(varyHeader.length() - 1);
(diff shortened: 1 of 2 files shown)


markt-asf:
Thanks for the pull request. I've applied a slightly different version as:

- we patch 10.0.x first and then back-port
- I agree it is better if the tests don't use Set
- On reflection, I think it is better not to modify the existing header order and add any new value to the end

That last point isn't strictly necessary but it strikes me as what users will expect.

PR #384, opened 2020-12-04
Fix BZ 60781 Escape access log output where httpd does

markt-asf:
https://bz.apache.org/bugzilla/show_bug.cgi?id=60781

It isn't quite as simple just escape where httpd does as httpd and Tomcat support slightly different things and validate at different points. See the code comments.

I was concerned about performance but after checking which elements need escaping (see code comments) I think this should be OK. I used Felix's original patch as my starting point but used chars rather than bytes and appended directly to the buffer. Give the potential performance issues, doing this as a PR.

### DIFF
--- a/java/org/apache/catalina/valves/AbstractAccessLogValve.java
+++ b/java/org/apache/catalina/valves/AbstractAccessLogValve.java
@@ -1869,37 +1869,37 @@ public abstract class AbstractAccessLogValve extends ValveBase implements Access
 
         for (char c : input.toCharArray()) {
             switch (c) {
-                // " and \
-                case '\\':
-                    dest.append("\\\\");
-                    break;
-                case '\"':
-                    dest.append("\\\"");
+            // " and \
+            case '\\':
+                dest.append("\\\\");
                 break;
-                // Standard C escapes for whitespace (not all standard C escapes)
-                case '\f':
-                    dest.append("\\f");
-                    break;
-                case '\n':
-                    dest.append("\\n");
-                    break;
-                case '\r':
-                    dest.append("\\r");
-                    break;
-                case '\t':
-                    dest.append("\\t");
-                    break;
-                case '\u000b':
-                    dest.append("\\v");
-                    break;
-                default:
-                    // Control, delete (127) or above 127
-                    if (c < 32 || c > 126) {
-                        dest.append("\\u");
-                        dest.append(HexUtils.toHexString(c));
-                    } else {
-                        dest.append(c);
-                    }
+            case '\"':
+                dest.append("\\\"");
+            break;
+            // Standard C escapes for whitespace (not all standard C escapes)
+            case '\f':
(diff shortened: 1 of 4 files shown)


java/org/apache/catalina/valves/AbstractAccessLogValve.java
```diff
@@ -1805,4 +1808,99 @@ protected AccessLogElement createAccessLogElement(char pattern) {
...
+
+        for (char c : input.toCharArray()) {
+            switch (c) {
+                // " and \
+                case '\\':
+                    dest.append("\\\\");
+                    break;
+                case '\"':
+                    dest.append("\\\"");
+                break;
```
martin-g, line 1878: indentation is off

test/org/apache/catalina/valves/TestAbstractAccessLogValveEscape.java
```diff
@@ -0,0 +1,65 @@
...
+        List<Object[]> parameters = new ArrayList<>();
+
+        parameters.add(new String[] { null, "-" });
+        parameters.add(new String[] { "", "-" });
+        parameters.add(new String[] { "ok", "ok" });
+        parameters.add(new String[] { "o\tk", "o\\tk" });
+        parameters.add(new String[] { "o\u0002k", "o\\u0002k" });
+        parameters.add(new String[] { "o\u007fk", "o\\u007fk" });
+        parameters.add(new String[] { "o\u0080k", "o\\u0080k" });
+        parameters.add(new String[] { "o\u00ffk", "o\\u00ffk" });
```
martin-g, line 47: I've expected to see an input with `"`, to cover the bug report.

martin-g (approved):

java/org/apache/catalina/valves/AbstractAccessLogValve.java
markt-asf: Personal style choice (I prefer case statements to be indented from the switch). The Tomcat code is inconsistent. Looks like the consensus amongst Java developers is for no indents. I don't mind changing this.

We have a commented out check for indentation in checkstyle. Looks like it still triggers ~5k errors.

test/org/apache/catalina/valves/TestAbstractAccessLogValveEscape.java
markt-asf: Good point. I'll add one.

java/org/apache/catalina/valves/AbstractAccessLogValve.java
martin-g: I meant the `break` statement. It is not aligned with the other `break`s

martin-g: I see there was a new commit that moved left everything but the problematic `break` :-)

martin-g (approved):

PR #385, opened 2020-12-06
Add peerAddress to coyote request

rainerjung:
It contains the IP address of the direct connection peer.
If a reverse proxy sits in front of Tomcat and the protocol
used is AJP or HTTP in combination with the RemoteIp(Valve|Filter),
the peer address might differ from the remoteAddress.
The latter then contains the address of the client in front of the
reverse proxy, not the address of the proxy itself.

Support for the peer address has been added to the
RemoteAddrValve and RemoteCIDRValve with the new attribute
"usePeerAddress". This can be used to restrict access
to Tomcat bsed on the reverse proxy IP address, which is especially
useful to harden access to AJP connecrtors.

The peer address can also be logged in the access log
using the new %{peer}a syntax.

### DIFF
--- a/java/org/apache/catalina/valves/AbstractAccessLogValve.java
+++ b/java/org/apache/catalina/valves/AbstractAccessLogValve.java
@@ -165,6 +165,13 @@ public abstract class AbstractAccessLogValve extends ValveBase implements Access
         LOCAL, REMOTE
     }
 
+    /**
+     * The list of our ip address types.
+     */
+    private enum RemoteAddressType {
+        REMOTE, PEER
+    }
+
     //------------------------------------------------------ Constructor
     public AbstractAccessLogValve() {
         super(true);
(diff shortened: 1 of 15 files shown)


java/org/apache/catalina/valves/AbstractAccessLogValve.java
```diff
@@ -861,19 +868,50 @@ public void addElement(CharArrayWriter buf, Date date, Request request,
      * write remote IP address - %a
      */
     protected class RemoteAddrElement implements AccessLogElement, CachedElement {
+        /**
+         * Type of address to log
+         */
+        private static final String remoteAddress = "remote";
```
martin-g, line 874: s/remoteAddress/REMOTE_ADDRESS/, to follow Java conventions ?! Same for `peerAddress` below.

java/org/apache/coyote/Constants.java
```diff
@@ -96,4 +96,11 @@
      * the X-Forwarded-For HTTP header.
      */
     public static final String REMOTE_ADDR_ATTRIBUTE = "org.apache.tomcat.remoteAddr";
+
+    /**
+     * The request attribute set by the RemoteIpFilter, RemoteIpValve (and may
+     * be set by other similar components) that identifies for the connector the
+     * conection peer IP address.
```
martin-g, line 103: con`n`ection

java/org/apache/coyote/Request.java
```diff
@@ -95,6 +95,7 @@ public Request() {
 
     // remote address/host
     private final MessageBytes remoteAddrMB = MessageBytes.newInstance();
+    private final MessageBytes peerAddrMB = MessageBytes.newInstance();
```
martin-g, line 98: Do we care that it is either `remote` or `peer`? So one of them wastes memory. Or it is negligible ?

webapps/docs/changelog.xml
```diff
@@ -117,6 +117,22 @@
...
+        the IP address of the direct connection peer. If a reverse proxy
+        sits in front of Tomcat and the protocol used is AJP or HTTP
+        in combination with the <code>RemoteIp(Valve|Filter)</code>,
+        the peer address might differ from the <code>remoteAddress</code>.
+        The latter then contains the address of the client in front of the
+        reverse proxy, not the address of the proxy itself.
+        Support for the peer address has been added to the
+        RemoteAddrValve and RemoteCIDRValve with the new attribute
+        <code>usePeerAddress</code>. This can be used to restrict access
+        to Tomcat bsed on the reverse proxy IP address, which is especially
```
martin-g, line 131: s/bsed/based/

webapps/docs/changelog.xml
```diff
@@ -117,6 +117,22 @@
...
+        useful to harden access to AJP connecrtors. The peer address can also
```
martin-g, line 132: s/connecrtors/connectors/

java/org/apache/catalina/valves/AbstractAccessLogValve.java
rainerjung: AbstractAccessLogValve consistently uses camel case for static Strings. I sticked to the local concentions in ther file. I would prefer to change the static string nameing conventions in an independent commit.

java/org/apache/coyote/Constants.java
rainerjung: Thanks.

java/org/apache/coyote/Request.java
rainerjung: I think in many cases where one would need peer, remote can ba ueseful as well.

webapps/docs/changelog.xml
rainerjung: Thanks.

webapps/docs/changelog.xml
rainerjung: Thanks.

PR #387, opened 2020-12-10
Fix typo for message 'jsp.exception' (french locale)

ghusta:
Error in french translation for message 'jsp.exception'

_durent_ instead of _durant_


rmaucher:
I have now improved the string in poeditor. Thanks.

You should do this kind of fix in poeditor rather than in a PR.

ghusta:
OK thanks. I didn't know poeditor.

martin-g:
I think we need to improve our "How to contribute" docs. Both at Tomcat website and here in GitHub PR templates.
I've seen several times poeditor being mentioned here and in the mailing lists but never seen how to use it. 
One can easily find https://poeditor.com/ but from there on I have no information how to use it and how to find Tomcat resources in it. And then once I find and change them then I guess I need to export them and import them in Git again.

So, before providing this information we should not blame our contributors.

rmaucher:
http://tomcat.apache.org/getinvolved.html -> click on "translations" and it looks good enough to me.

markt-asf:
I've updated that text to reflect the recent changes to escaping.

martin-g:
> http://tomcat.apache.org/getinvolved.html -> click on "translations" and it looks good enough to me.

Thanks!
I've grepped Tomcat webapps/docs for poeditor and it didn't return any hits.
I'll read the wiki and add GH template.

PR #388, opened 2020-12-12
Update ide-support package version

zzk0:
**Problem:**

I followed the steps [here](https://tomcat.apache.org/tomcat-10.0-doc/building.html) to build tomcat. I run the command `ant ide-intellij`, then open the IDEA. The package aQute will not be found.

**What I have done:**

Just modify the version of that package.


isapir:
This update matches the version in build.properties.default so it looks fine to me:
https://github.com/apache/tomcat/blob/master/build.properties.default#L295

martin-g (approved):

PR #391, opened 2020-12-17
Optimize if condition

lanicc:
Optimize if condition in `AbstractProtocol.startAsyncTimeout()`,
from `if (timeoutFuture == null || (timeoutFuture != null && timeoutFuture.isDone()))`
 to `if (timeoutFuture == null || timeoutFuture.isDone())`


martin-g (approved):

martin-g:
Thank you, @lanicc !

PR #395, opened 2020-12-18
 Replace C-style array declaring with Java-style array declaring

lanicc:


martin-g:
There are many of those. A quick search shows around 400 occurrences of the C-style.
I am not sure this is worth it. It will make backporting very hard.

zzk0:
> There are many of those. A quick search shows around 400 occurrences of the C-style.
> I am not sure this is worth it. It will make backporting very hard.

What if we write new code? Should we write the Java-style array?

ChristopherSchultz:
Yes, please use Java-style array declaration syntax in new code.

lanicc:
I think code style should be unified,especially obvious ones like this

martin-g:
@lanicc We agree that the code style should be consistent and that it should follow Java conventions. But we will do it only when editing such code for other functional changes.

PR #398, opened 2020-12-27
Ensure StandardContext#postWorkDirectory is protected

rmannibucau:
Goal is to enable to run tomcat without any temporary files.
For programmatic cases it is often possible and desired (note it is ok to skip jakarta.servlet.context.tempdir for such case).


martin-g (approved):

markt-asf:
Applied manually so I could add a changelog entry

PR #400, opened 2021-01-13
adding exitOnFirstListenerFailure flag in standardcontext to enable to stop at first listener error

rmannibucau:
follow up of the list discussion, trying to propose a new flag.

### DIFF
--- a/java/org/apache/catalina/core/StandardContext.java
+++ b/java/org/apache/catalina/core/StandardContext.java
@@ -172,6 +172,12 @@ public class StandardContext extends ContainerBase
     // ----------------------------------------------------- Instance Variables
 
     /**
+     * Allow to fail as soon as the first context listener failed.
+     * Note that it does not impact stop phase since some listeners can just cleanup some state.
+     */
+    protected boolean exitOnFirstListenerFailure = false;
+
+    /**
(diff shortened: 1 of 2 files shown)


rmaucher:
Ideally, this should be resolved one way or the other by the specification. Also I noticed there is a failCtxIfServletStartFails attribute already, which can easily be abused without too many problems. This is for load on startup Servlets. Overall I don't think it's a great plan to add flags for everything.

rmannibucau:
@rmaucher agree it would be great if spec can clarifies it (in a way making the flag true OOTB ;)) but merging with the servlet flag is kind of different, listener being used to setup env (or symmetrically cleanup), you can need to go through them even if some fails (for auditing for ex) wheras servlets being the runtime it is often an all or none so guess we should keep both split.

markt-asf:
Can you expand on why this is an issue? The web application doesn't start either way and - generally - I think I'd prefer to see the scale of the problem (how many errors are there in total) rather than face an unknown number of repetitions of "start app, fails, fix error, repeat".
I don't recall any EG discussion on this topic but 'll look to see if anything happened before my time or if I've forgotten something.

rmannibucau:
issue didnt pop up again and workarounds are working so closing if there is no more request of it

PR #402, opened 2021-02-01
Add support for Unix Domain Sockets (APR Protocol)

minfrin:
Add support and test case for Unix Domain Sockets in
org.apache.coyote.http11.Http11AprProtocol, to support
users using Java versions older than 16.

Requires tomcat-native 1.2.26 and above.

Backport of the following:

https://github.com/apache/tomcat/commit/c400239cb17d82f8e10f12e40881293b7d2d70b9
https://github.com/apache/tomcat/commit/a616bf385a350175a33a0ebf09d8b6688344e9e3


markt-asf:
Merged manually so I could add the change log entry to the correct version.

exabrial:
Hey guys, My apologies to comment on a year-old PR, but I cannot get `unixDomainSocketPath` to work on Tomcat9 with the latest APR, Java11, Ubuntu 18.04. What happens is the second request to the server causes the server to hang on the socket. I was trying to find _any_ example of a working server.xml, but it's not part of the PR. Any chance you guys could post your working server.xml? Otherwise I may have spotted a bug and can start a new discussion with a test case. Thank you!

aooohan:
> Hey guys, My apologies to comment on a year-old PR, but I cannot get `unixDomainSocketPath` to work on Tomcat9 with the latest APR, Java11, Ubuntu 18.04. What happens is the second request to the server causes the server to hang on the socket. I was trying to find _any_ example of a working server.xml, but it's not part of the PR. Any chance you guys could post your working server.xml? Otherwise I may have spotted a bug and can start a new discussion with a test case. Thank you!
> 
> prob: [http://stackoverflow.com/questions/72975059/tomcat-crashe…](http://stackoverflow.com/questions/72975059/tomcat-crashe%E2%80%A6)

I think the reason for your problem is that your JDK version is too low, you should use version 16+.
You can take a look at  the [commit](https://github.com/apache/tomcat/commit/a616bf385a350175a33a0ebf09d8b6688344e9e3) ,hopes for helping you.

exabrial:
Thank you for the reply. So it looks like if you are below Java 16, it uses APR, and if above Java16, it uses the Java APIs. I think there is a bug in APREndpoint, as my example is using Java11 + APR. When I commented out the "duplicate accept check" code in APREndpoint.java, Java11 + APR = Unix Sockets working.

aooohan:
> Thank you for the reply. So it looks like if you are below Java 16, it uses APR, and if above Java16, it uses the Java APIs. I think there is a bug in APREndpoint, as my example is using Java11 + APR. 

Indeed, i have reproduced this bug.  You can report this bug again on BZ.

> When I commented out the "duplicate accept check" code in APREndpoint.java, Java11 + APR = Unix Sockets working.

This is use for protecting  against a known OS bug.

aooohan:
> What happens is the second request to the server causes the server to hang on the socket.
Inaddtion, this is a second bug. It is purely by chance that you encountered this bug. I've found that this bug exists for a long time, and not putting the wrapper in connections before throwing an exception to prevent a known OS bug has led to it not being closed in time.

aooohan:
> What happens is the second request to the server causes the server to hang on the socket.

I've found this bug for a long time, and not putting the wrapper in connections before throwing an exception to prevent a known OS bug has led to it not being closed in time.

PR #403, opened 2021-02-01
Allow the Manager application to deploy applications in parallel

markt-asf:
As per this thread:
https://tomcat.markmail.org/thread/33xw2jkp4fmcpl7o

Details in the commit messages but the short version was "serviced" wasn't doing what was intended in all cases. Improving serviced enabled the Manager app to perform deploy different web apps in parallel.

This is a fairly large change so submitting as a PR for review before I commit it.


rmaucher:
Seems like a straightforward and logical expansion of the concept.

PR #410, opened 2021-03-04
Thank you! Apologies if unable to use right now...

mcroteau:
.war extensions I believe may turn some people off so I refactored much of the deployment routines to accept additional extensions. Much of the heavy lifting is finished but needs additional testing. Thank you for all that you provide, I have been using Tomcat for 12 years now. I am truly grateful. 

Please let me know if you would like me to proceed. There is a small issue with the Manager app that I needs to be addressed. The ManagerServlet keeps the original extension and adds .war. 

Thank you.
Mike

Edit: I just noticed the smoke checks failing. I will address if you would like.


markt-asf:
I understand where you are coming from with this PR but you can't break backwards compatibility the way this PR does. It also breaks things as can be seen from the CI checks.

Bundle has its own meaning in this space - see OSGi - and I think using it in this context will cause user confusion.

This is a very big change. Too big, I would suggest, for a single PR.

My suggestion would be to close this PR and look at the user visible uses of "war" - at least as a starting point - as far more people will see those than the internal code. While I don't think it will be possible to remove every reference right away, I am sure there are places where web archive could be used instead. I suggest starting small, maybe with a single documentation page.

mcroteau:
So I take it as a "no" to continue? The java source is pretty much covered. I went line by line and replaced with Bundle/bundle and will do the same for test if you would like. 

Thank you.

markt-asf:
Closing as per my previous suggestion.

To repeat what I said previously:
> My suggestion would be ... look at the user visible uses of "war" - at least as a starting point - as far more people will see those than the internal code. While I don't think it will be possible to remove every reference right away, I am sure there are places where web archive could be used instead. I suggest starting small, maybe with a single documentation page.

mcroteau:
I’m sorry.

PR #411, opened 2021-03-07
Add Health Check Valve

git001:
To offer the current container Orchestrator  the option to check the tomcat health let's add and activate the Health Check Valve.

It would be nice to back port this to TC 10+9


markt-asf:
The health check valve is not appropriate all all configurations so it should not be enabled by default.

PR #412, opened 2021-03-08
Adding default manager roles in tomcat users config.

dagnelies:
Every time I install a tomcat, I first add the tomcat manager users, and I think a lot of people do this as well. However, currently, it's kind of a pain because you have to look it up first what exactly should be written and create the roles/users manually. This here makes it trivial.

### DIFF
--- a/conf/tomcat-users.xml
+++ b/conf/tomcat-users.xml
@@ -20,15 +20,26 @@
               xsi:schemaLocation="http://tomcat.apache.org/xml tomcat-users.xsd"
               version="1.0">
 <!--
-  NOTE:  By default, no user is included in the "manager-gui" role required
+  By default, no user is included in the "manager-gui" role required
   to operate the "/manager/html" web application.  If you wish to use this app,
-  you must define such a user - the username and password are arbitrary. It is
-  strongly recommended that you do NOT use one of the users in the commented out
-  section below since they are intended for use with the examples web
-  application.
+  you must define such a user - the username and password are arbitrary.
+
+  Tomcat manager roles:
+    - manager-gui    - allows access to the HTML GUI and the status pages
+    - manager-script - allows access to the text interface and the status pages
+    - manager-jmx    - allows access to the JMX proxy and the status pages
+    - manager-status - allows access to the status pages only
 -->
 <!--
-  NOTE:  The sample user and role entries below are intended for use with the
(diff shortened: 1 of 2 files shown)


martin-g (approved):

dagnelies:
...I guess the travis CI is currently broken

martin-g:
For some reason `getParentStream()` returns null on s390x (JDK11).
 https://github.com/apache/tomcat/blob/efe0fe8b3d6b10b7e416c32b59f60bb441831c39/java/org/apache/coyote/http2/AbstractNonZeroStream.java#L125

ChristopherSchultz:
> For some reason `getParentStream()` returns null on s390x (JDK11).
> 
> https://github.com/apache/tomcat/blob/efe0fe8b3d6b10b7e416c32b59f60bb441831c39/java/org/apache/coyote/http2/AbstractNonZeroStream.java#L125

🤔

dagnelies:
Hi, since the Travis CI possibly failed due to some unrelated instability, can we perhaps simply try to re-run it?

martin-g:
>  can we perhaps simply try to re-run it?

Re-scheduled it!

kkolinko (changes requested):
The "manager-jmx" role and "manager-gui" roles should not be used together. (As I wrote is a comment earlier. Resubmitting via "Code review" panel.)

dagnelies:
@kkolinko Regarding:

> There rarely is a need to explicitly create roles like the above. When parsing the tomcat-users.xml file, all roles mentioned in users are created automatically.

I wasn't aware of that. If roles are automatically generated, what's the use of the annotation at all? I just assumed it was required since it's present in all examples.

I'll remove the JMX part.

kkolinko:
> 
> 
> @kkolinko Regarding:
> 
> > There rarely is a need to explicitly create roles like the above. When parsing the tomcat-users.xml file, all roles mentioned in users are created automatically.
> 
> I wasn't aware of that. If roles are automatically generated, what's the use of the `role` tag at all? I just assumed it was required since it's present in all examples.

- If you need to declare a role that has no users assigned to it, the role can be declared with a "role" element. Such use case is rare (e.g. if users are managed via some GUI and you want to be able to list all available roles).
- When a user database is saved (written out), "role" elements are written as well, for completeness. This operation can be triggered via JMX.

PR #413, opened 2021-03-12
Supplemental fix for https://github.com/apache/tomcat/commit/c232657

bekionel:
Add optimizations classes to tomcat-embed-jasper as well

Thanks to @csutherl for spotting :)


markt-asf:
Thanks. I applied this manually so I could add a change log entry.

PR #414, opened 2021-03-17
don't log undesired warnings for webapp classloader configuration

rmannibucau:


rmaucher:
This was introduced by https://github.com/apache/tomcat/commit/c38e9ccabc0d20e1f88b389ab531dcbf632b0ed8 and I understand the WebappLoader changes part of it.

In StandardContext, I don't see why the properties should not be set simply after using if (getLoader().getClassLoader() instanceof WebappClassLoaderBase) since the StandardContext class already uses WebappLoader (to create a default loader). This would avoid the useless reflection that is causing you problems.

rmannibucau:
@rmaucher due to the code structure I assumed it was intended for custom loader to be able to benefit from the same configuration without having to inherit from the default webapp classloader, if this assumption is true I agree with you it is better to do "if we know then do, else ignore". Not sure how an API it is since it is somehow publicly available.

rmaucher:
@rmannibucau That wasn't the case before that commit, but it's possible after it. Personally, I would prefer dropping the reflection. Let's see if Mark wants to comment on that.

rmaucher:
I added the simple reflection free version, so the properties now need WebappCLBase and we'll see what happens. I didn't merge the test since it's quite obvious when there's no reflection and heuristic, but you can add it if you'd like.

PR #417, opened 2021-04-28
Fix BZ 65272. Restore the use of LF as an HTTP line terminator

markt-asf:
Potential fix for https://bz.apache.org/bugzilla/show_bug.cgi?id=65272

Needs careful review, hence using a PR.
If you spot any potential ways an invalid HTTP request line or header could be:
 - accepted as valid
 - rejected later than necessary
 - rejected for the 'wrong' reason

or any other potential parsing error please report it. For bonus points provide a test case that demonstrates the issue.

As always, anything that might have security implications for a current Tomcat release should go to security@a.o rather than here.

### DIFF
--- a/java/org/apache/coyote/http11/Http11InputBuffer.java
+++ b/java/org/apache/coyote/http11/Http11InputBuffer.java
@@ -550,9 +550,14 @@ public class Http11InputBuffer implements InputBuffer, ApplicationBufferHandler 
                 prevChr = chr;
                 chr = byteBuffer.get();
                 if (chr == Constants.CR) {
-                    // Possible end of request line. Need LF next.
+                    // Possible end of request line. Need LF next else invalid.
                 } else if (prevChr == Constants.CR && chr == Constants.LF) {
+                    // CRLF is the standard line terminator
                     end = pos - 1;
+                    parsingRequestLineEol = true;
+                } else if (chr == Constants.LF) {
+                    // LF is an optional line terminator
+                    end = pos;
                     parsingRequestLineEol = true;
                 } else if (prevChr == Constants.CR || !HttpParser.isHttpProtocol(chr)) {
                     String invalidProtocol = parseInvalid(parsingRequestLineStart, byteBuffer);
@@ -841,7 +846,8 @@ public class Http11InputBuffer implements InputBuffer, ApplicationBufferHandler 
 
             if (chr == Constants.CR && prevChr != Constants.CR) {
                 // Possible start of CRLF - process the next byte.
-            } else if (prevChr == Constants.CR && chr == Constants.LF) {
+            } else if (chr == Constants.LF) {
+                // CRLF or LF is an acceptable line terminator
                 return HeaderParseStatus.DONE;
             } else {
                 if (prevChr == Constants.CR) {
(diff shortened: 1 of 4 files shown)


test/org/apache/coyote/http11/TestHttp11InputBuffer.java
```diff
@@ -687,24 +691,6 @@ public void testInvalidEndOfRequestLine01() {
     }
 
 
-    @Test
```
ChristopherSchultz, line 690: I would leave this test in, but change its nature (i.e. require that the response code is 200). Or is your assertion that the "correct" test is below in `test/org/apache/coyote/http11/TestHttp11InputBufferCRLF.java `?

webapps/docs/changelog.xml
```diff
@@ -143,6 +143,12 @@
         request line, ensure that all the available data is included in the
         error message. (markt)
       </fix>
+      <fix>
+        <bug>65272</bug>: Restore the optional HTTP feature that allows
+        <code>LF</code> to be treated as a line terminator as well as the
```
ChristopherSchultz, line 148: This is specifically for a line-terminator for request-line and headers, correct? Is it worth mentioning that specifically?

java/org/apache/coyote/http11/Http11InputBuffer.java
```diff
@@ -550,10 +550,15 @@ boolean parseRequestLine(boolean keptAlive, int connectionTimeout, int keepAlive
                 prevChr = chr;
                 chr = byteBuffer.get();
                 if (chr == Constants.CR) {
-                    // Possible end of request line. Need LF next.
+                    // Possible end of request line. Need LF next else invalid.
                 } else if (prevChr == Constants.CR && chr == Constants.LF) {
+                    // CRLF is the standard line terminator
                     end = pos - 1;
                     parsingRequestLineEol = true;
+                } else if (chr == Constants.LF) {
```
ChristopherSchultz, line 558: Is it worth "fingerprinting" a message to determine if it is self-consistent? That is, a message is allowable with LF-terminated request/headers but _only_ if *all* of its headers and the request line are terminated with LF (only)? Or is that being unnecessarily picky and complicating the code?

java/org/apache/coyote/http11/Http11InputBuffer.java
```diff
@@ -841,7 +846,8 @@ private HeaderParseStatus parseHeader() throws IOException {
 
             if (chr == Constants.CR && prevChr != Constants.CR) {
```
ChristopherSchultz, line 847: This appears to essentially ignore (all?) CR characters in headers. I'm not sure that's what we want.

webapps/docs/changelog.xml
markt-asf: Can do.

java/org/apache/coyote/http11/Http11InputBuffer.java
markt-asf: It would make the code a little more complex and there is nothing in the spec to suggest that self-consistency is required. If we were going to make the behaviour configurable I can see how this would fit into an option that allow varying degrees of LF use. Eg: none, consistent, any.
My current thinking is to essentially hard-code the any option unless someone identifies a good reaaon not to.

java/org/apache/coyote/http11/Http11InputBuffer.java
markt-asf: It shouldn't. What it should do is detect if a line starts with CRLF or LF and if it does set the HeaderParseStatus.DONE state. Anything else is passed through to be parsed as the start of the header name. That particular line is handling the case where the lines starts with CR which may be the start of a CRLF sequence.

test/org/apache/coyote/http11/TestHttp11InputBuffer.java
markt-asf: Yes, the correct test is below.

rmaucher:
+1 looking good.

PR #421, opened 2021-06-03
Change log level from error to debug

rohini2021:
Currently, in reconnectIfExpired(), reconnection failure is producing an error level log. This causes a lot of false positive in our alerting systems. This should be changed to debug level log.


markt-asf:
A (re-)connection failure doesn't strike me as something that we'd want to log at debug level. Logging at error level looks appropriate to me.

markt-asf:
No further response. Closing.

PR #431, opened 2021-07-03
Bug 65426 - HTTP2 Request is Stalled due to Exception on Tomcat. (PING FRAME)

bmistry13:
Hello Tomcat Team,

I am getting STALLED requests for PING FRAME occasionally and the chrome browser gets stuck for a long period of time.  It was due to the parent stream being null.  Since PING request is stream-id 0.  The parent is NULL. I was getting NPE.

Here is Bug: https://bz.apache.org/bugzilla/show_bug.cgi?id=65426

### DIFF
--- a/java/org/apache/coyote/http2/Http2UpgradeHandler.java
+++ b/java/org/apache/coyote/http2/Http2UpgradeHandler.java
@@ -1294,7 +1294,7 @@ class Http2UpgradeHandler extends AbstractStream implements InternalHttpUpgradeH
                     // Don't try and remove Stream 0 as that is the connection
                     // Don't try and remove 'newer' streams. We'll get to them as we
                     // work through the ordered list of streams.
-                    while (toClose > 0 && parent.getIdAsInt() > 0 && parent.getIdAsInt() < stream.getIdAsInt() &&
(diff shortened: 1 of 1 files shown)


markt-asf:
This PR will not be applied because:
1. It addresses the symptom, not the root cause
2. The root cause was fixed in 10.0.5, 9.0.45 and 8.5.65

bmistry13:
@markt-asf , can you please provide commit or bug # that address this issue?

rmaucher:
https://bz.apache.org/bugzilla/show_bug.cgi?id=65195

bmistry13:
@markt-asf ,  thanks for bug #.  Is it possible to get a commit id that fixed the Root cause?

markt-asf:
5e3c0ce7 although there was a fair amount of related refactoring at the time so it may depend on additional commits

bmistry13:
@markt-asf ,  Thank you I have tested it works with 9.0.50.   What could cause the following?  HTTP2 request gets stalled due to reset:

I am occasionally  running into following exception:

Caused by: java.io.IOException: Stream reset
	at org.apache.coyote.http2.Stream$StreamInputBuffer.doRead(Stream.java:1086)
	at org.apache.coyote.Request.doRead(Request.java:640)
	at org.apache.catalina.connector.InputBuffer.realReadBytes(InputBuffer.java:317)

PR #432, opened 2021-07-09
Allow subclassing CorsFilter getAllowedOrigins() and isAnyOriginAllow…

ekkelenkamp:
I would like to be able to subclass the CorsFilter clases to be able to provide a custom list of allowed origins, instead of configure them from the init parameters. This will allow providing the allowedOrigins at runtime and not hardcoded in the web.xml init parameters.
Unfortunately the method **isOriginAllowed** uses the instance variable allowedOrigns directly. This makes it impossible to override the available function: getAllowedOrigins and isAnyOriginAllowed.

See also:
https://bz.apache.org/bugzilla/show_bug.cgi?id=65443


markt-asf:
I went a little further and ensured getters were used for all configurable fields.

PR #433, opened 2021-07-12
Fix typo

pachulisk:
Add missing words


markt-asf:
Thanks for the PR.
Please note that trailing whitespace should be removed from all lines. If you run checkstyle using the provided configuration, it should catch issues like that.

PR #434, opened 2021-07-14
Optionally inherit request parameters

Pingvin235:
Greetings to All,

the proposed fix makes optional taking requests parameters for internal requests. 
Which produced when using JSTL directives `<c:import`. 

**Explanation case** 

Processing `REQUEST1` with parameters `param1=value11` and `param2=value12`. 
Processing forwards to a JSP page, containing internal sub-request:
``` jsp
<c:import url="URL2?param2=value22&param3=value23"
```
In called `REQUEST2` will be presented:
* `param1=value11`
* `param2=value12&param2=value22`
* `param3=value23`

Such behavior is not always wanted because of collisions in similar named parameters. 

**Solution**

The proposed change has been implemented for a long time ago and successfully tested internally. Would be great to include it in main project. Kindly ask to help with passing through all the requirements for that.

### DIFF
--- a/java/org/apache/catalina/core/ApplicationHttpRequest.java
+++ b/java/org/apache/catalina/core/ApplicationHttpRequest.java
@@ -92,6 +92,15 @@ class ApplicationHttpRequest extends HttpServletRequestWrapper {
 
     private static final int SPECIALS_FIRST_FORWARD_INDEX = 6;
 
+    private static final boolean INHERIT_REQUEST_PARAMETERS;
+
+    /**
+     * Take parameters from {@link #getRequest()}.
+     */
+    static {
+        INHERIT_REQUEST_PARAMETERS = Boolean.parseBoolean(System.getProperty(
+                "org.apache.catalina.core.ApplicationHttpRequest.INHERIT_REQUEST_PARAMETERS", "true"));
+    }
 
     // ----------------------------------------------------------- Constructors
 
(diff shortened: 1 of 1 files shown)


markt-asf:
Such an option would be in direct violation of the Servlet specification. That makes me very hesitant to agree to such a change. The Servlet specification already states that the parameters from the include take precedence. I'd rather add clarification of any ambiguities (if necessary) to the spec and align Tomcat's behaviour with that than add an option for non-compliant behaviour.

Regarding the implementation, I am -1 on the proposed implementation as it is based on an environment variable. If it were implemented, It would need to be a per context configuration attribute.

Pingvin235:
Could you please point me the place in the specification, describing that behavior? For 4.0 could found only section 9.3 regarding includes, but didn't notice anything about parameters inheritance / precedence..

Pingvin235:
> Regarding the implementation, I am -1 on the proposed implementation as it is based on an environment variable. If it were implemented, It would need to be a per context configuration attribute.

Can do it this way of course case this option is in principle allowed.

markt-asf:
You want the last paragraph of section 9.1.1. A RequestDispatcher is what performs the include or forward.

Pingvin235:
Thanks for pointing. Assuming this is the conflicting part:

"Parameters specified in the query string used to create the RequestDispatcher take precedence over other parameters of the same name passed to the included servlet."

Perhaps I understand it wrong, but if other parameters are completely excluded, do query string parameters loose their precedence? Actually that we need is to have the query string parameters only. May be you can recommend another way to archive such behavior?

markt-asf:
This is not going to be applied. The users mailing list is the place to discuss your requirement and possible solutions.

As a starting point, how about ensuring the target servlets take only the first value for any parameter. Then you can rely on the precedence rules in the specification.

PR #436, opened 2021-07-22
ant ide-eclipse lib update

Dmole:


markt-asf:
LGTM. Thanks for the patch.
Just a thought. If the versions were replaced with the relevant ${...} properties, and filtering enabled for the copy, there would be no need to keep these files in sync.

PR #439, opened 2021-08-04
Minimal fixes to documentation of Context and Host

rrodewald:
- Documented default value of unpackWARs attribute
- Tagged all boolean values als code
- Punctuation

### DIFF
--- a/webapps/docs/config/host.xml
+++ b/webapps/docs/config/host.xml
@@ -110,21 +110,21 @@
       </attribute>
 
       <attribute name="createDirs" required="false">
-        <p>If set to true, Tomcat will attempt to create the directories defined
+        <p>If set to <code>true</code>, Tomcat will attempt to create the directories defined
         by the  attributes <code>appBase</code> and <code>xmlBase</code> during
         the startup phase. The default value is <code>true</code>. If set to
-        true, and directory creation fails, an error message will be printed out
+        <code>true</code>, and directory creation fails, an error message will be printed out
         but will not halt the startup sequence.</p>
       </attribute>
 
       <attribute name="autoDeploy" required="false">
         <p>This flag value indicates if Tomcat should check periodically for new
-        or updated web applications while Tomcat is running. If true, Tomcat
+        or updated web applications while Tomcat is running. If <code>true</code>, Tomcat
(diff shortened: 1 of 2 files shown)


webapps/docs/config/host.xml
```diff
@@ -302,7 +302,7 @@
         placed in the <code>appBase</code> directory as web application
         archive (WAR) files to be unpacked into a corresponding disk directory
         structure, <code>false</code> to run such web applications directly
-        from a WAR file. See
+        from a WAR file. The flag's value defaults to <code>true</code>. See
```
michael-o, line 305: The default is...

michael-o (changes requested):

webapps/docs/config/host.xml
rrodewald: Shall I change this for all other occurences of the same wording?

michael-o (approved):

kkolinko:
Reviewing...
1. context.xml L187  `s/docBase/<code>docBase</code>/ `as well
2. I am more used to "The default value is" rather than "The default is".
I wonder what wording is better. Tomcat documentation uses both.

rrodewald:
> Reviewing...
> 
>     1. context.xml L187  `s/docBase/<code>docBase</code>/ `as well
> 
>     2. I am more used to "The default value is" rather than "The default is".
>        I wonder what wording is better. Tomcat documentation uses both.

1. Done
2. I am no native speaker so I do whatever I'm told to do...

ChristopherSchultz:
To a native English speaker, either one is completely understandable. 
"The default value is ___" is slightly more clear, but I would actually 
defer to NON-native English speakers because it may be more 
understandable to them one way or the other.

I'm happy to listen to the Russian :)

rrodewald:
If we want to standardize the wording for default values, I would suggest to open a new pull request for this, so that the main purpose of the present pull request (documentation of the default value of unpackWARs) can be considered as done.

Here is an incomplete list of wordings found:
- Defaults to ...
- Default [value] is ...
- The default [value] is ...

markt-asf:
Thanks for the PR. +1 tp a new PR to make the default wording consistent.
Applied manually so I could a) add a change log entry and b) broadly maintain the line length limit of 80 chars for xml files.

PR #441, opened 2021-08-10
Fix build fail when locale is not en

dachuan9e:
It will build fail when the locale is not set to en. Apache ant cannot support different locale format of time stamp. Showing the root cause by the following code in Ant. It needs to use the 24-hour format to avoid this issue.

```java
    public static final ThreadLocal<DateFormat> EN_US_DATE_FORMAT_MIN =
            ThreadLocal.withInitial(() -> new SimpleDateFormat("MM/dd/yyyy hh:mm a", Locale.US));

    public static final ThreadLocal<DateFormat> EN_US_DATE_FORMAT_SEC =
            ThreadLocal.withInitial(() -> new SimpleDateFormat("MM/dd/yyyy hh:mm:ss a", Locale.US));
```


markt-asf:
Thanks for the PR. I applied it manually so I could add a change log entry.

PR #442, opened 2021-08-10
Update http.xml, clarified compressionMinSize and compressibleMimeType

crisgeek:
Based on 21-July-2021 conversation with Mark Thomas in Tomcat users group


markt-asf:
Applied manually so I could add a change log entry.

PR #443, opened 2021-08-10
Delegate check for preemptive authentication from AuthenticatorBase to affected Authenticators

rrodewald:
The main purpose of the proposed refactoring is to give an individual `Authenticator` the possibility to decide if preemptive authentication is possible (e.g. if a completely different header is used for authentication).

In addition it yields cleaner code as the certificate handling code and the header name for basic, digest and spnego auth can now be moved to the relevant `Authenticator`s and does not pollute the `AuthenicatorBase`. `FormAuthenticator` and `NonLoginAuthenticator` don't need to override `isPreemptiveAuthRequest()` as preemptive is not supported/needed.


rrodewald:
Mixed up my branches. Will reopen without the Bugfix.

PR #445, opened 2021-08-17
Improve Travis CI build Performance

YunLemon:
According to [Build times out because no output was received](https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received), we should carefully use travis_wait, as it may make the build unstable and extend the build time.

[Caching Dependencies and Directories](https://docs.travis-ci.com/user/caching/) Travis CI can cache content that does not often change, to speed up the build process.

=====================
If there are any inappropriate modifications in this PR, please give me a reply and I will change them.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,7 +80,7 @@ install:
 
 script:
     - ant -q clean
-    - travis_wait 120 "./.travis/antTest.sh"
+    - "./.travis/antTest.sh"
 
 after_failure:
     - tail -n 5000 ant-test.log
@@ -90,3 +90,6 @@ after_failure:
 notifications:
     email:
       - dev@tomcat.apache.org
+cache:
+  directories:
+  - $HOME/.m2


markt-asf:
The use of `travis_wait` is required.

.travis.yml
```diff
@@ -80,7 +80,7 @@ install:
 
 script:
     - ant -q clean
-    - travis_wait 120 "./.travis/antTest.sh"
+    - "./.travis/antTest.sh"
```
martin-g, line 83: https://app.travis-ci.com/github/apache/tomcat/jobs/531899762 (the build for this PR) fails with `No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.`. So yes, `travis_wait` is needed!
Maybe the duration should be reduced though.
AFAIK Travis jobs could run for at most 50 mins but I see some which succeeded even after that...

.travis.yml
```diff
@@ -90,3 +90,6 @@ after_failure:
 notifications:
     email:
       - dev@tomcat.apache.org
+cache:
+  directories:
+  - $HOME/.m2
```
martin-g, line 95: Tomcat does not use Maven so I see no reason to cache this folder.

YunLemon: I add this because I see this repo contains pom.xml....

.travis.yml
YunLemon: This command will not exceed 60 mins? [https://app.travis-ci.com/github/apache/tomcat/jobs/529778525](https://app.travis-ci.com/github/apache/tomcat/jobs/529778525)

martin-g: I cannot say how long it might take. It depends on how loaded the Travis VM is. Sometimes jobs take twenty something minutes, sometimes fifty something minutes...
According to https://docs.travis-ci.com/user/customizing-the-build/#sts=Build%20Timeouts the max timeout for a job should be 50 mins, but it seems `travis_wait` may extend it. I cannot find any reference in their docs for its maximum.

.travis.yml
martin-g: Where is it ?
Did you confuse it with `build.xml` (Ant) ?

martin-g: Or maybe the ones in https://github.com/apache/tomcat/blob/main/modules/ ?
I am not sure whether Ant executes those. So the cache might be useful after all.

markt-asf:
It doesn't. It won't.

PR #447, opened 2021-09-03
Remove catalina ssi from catalina.jar.tmp.bnd file

huiwan:
References to org.apache.catalina.ssi are still contained in catalina.jar after it was moved to a separate catalina-ssi.jar. It should be cleaned up.


markt-asf:
Many thanks for the PR.

PR #449, opened 2021-09-09
Fix ContentRange parser

SchwingSK:
It failed to parse valid Content-Range headers
Closes #65563


markt-asf:
Thanks for the PR. I applied the PR manually as I wanted to make a few changes:
- add a comment explaining why we don't explicitly parse for SP
- add some additional tests
- add a change log entry

PR #451, opened 2021-09-15
make threadNameCache actually useful

t-gergely:


markt-asf:
Thanks for the PR. I'll add a change log entry and then back-port it.

PR #452, opened 2021-09-19
Introduce logs in RestCsrfPreventionFilter to improve troubleshooting.

PolinaGeorgieva:
Add some logs in RestCsrfPreventionFilter that would improve troubleshooting in case of failed CSRF validation.
Note that the RequestUtil.filter method is used only to apply some basic sensitive characters filtering of the requested path.
But as this method is deprecated in v.8.5.x and removed in v.9.0.x, would you recommend an alternative method I could use in this case?

### DIFF
--- a/java/org/apache/catalina/filters/RestCsrfPreventionFilter.java
+++ b/java/org/apache/catalina/filters/RestCsrfPreventionFilter.java
@@ -31,6 +31,10 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpSession;
 
+import org.apache.catalina.util.RequestUtil;
+import org.apache.juli.logging.Log;
+import org.apache.juli.logging.LogFactory;
+
 /**
  * Provides basic CSRF protection for REST APIs. The filter assumes that the
  * clients have adapted the transfer of the nonce through the 'X-CSRF-Token'
@@ -78,7 +82,8 @@ public class RestCsrfPreventionFilter extends CsrfPreventionFilterBase {
     private enum MethodType {
         NON_MODIFYING_METHOD, MODIFYING_METHOD
     }
-
+    private final Log log = LogFactory.getLog(RestCsrfPreventionFilter.class);
+    
     private static final Pattern NON_MODIFYING_METHODS_PATTERN = Pattern.compile("GET|HEAD|OPTIONS");
 
     private Set<String> pathsAcceptingParams = new HashSet<>();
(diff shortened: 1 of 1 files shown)


java/org/apache/catalina/filters/RestCsrfPreventionFilter.java
```diff
@@ -155,17 +160,29 @@ protected void storeNonceToSession(HttpSession session, String key, Object value
...
+                    nonceFromRequest,
+                    nonceFromSession)) {
                 return true;
             }
 
             storeNonceToResponse(response, Constants.CSRF_REST_NONCE_HEADER_NAME,
                     Constants.CSRF_REST_NONCE_HEADER_REQUIRED_VALUE);
             response.sendError(getDenyStatus(),
                     sm.getString("restCsrfPreventionFilter.invalidNonce"));
+            if (log.isErrorEnabled()) {
```
ChristopherSchultz, line 177: I don't think this should be logged at ERROR level, because an attacker could fill-up your logs just by repeatedly sending-in garbage. I would recommend using DEBUG log-level for this.

java/org/apache/catalina/filters/RestCsrfPreventionFilter.java
```diff
@@ -155,17 +160,29 @@ protected void storeNonceToSession(HttpSession session, String key, Object value
...
+                log.error("CSRF validation for REST failed! Request with method [" + request.getMethod() + "] and URI ["
+                        + RequestUtil.filter(request.getRequestURI())
```
ChristopherSchultz, line 179: I don't think there is a reason to use RequestUtil.filter() for this value; it's not going to be embedded in HTML so it's okay to use the value as-is.

java/org/apache/catalina/filters/RestCsrfPreventionFilter.java
```diff
@@ -217,6 +238,10 @@ public boolean apply(HttpServletRequest request, HttpServletResponse response) {
                 }
                 storeNonceToResponse(response, Constants.CSRF_REST_NONCE_HEADER_NAME,
                         nonceFromSessionStr);
+                if (log.isDebugEnabled()) {
```
ChristopherSchultz, line 241: I might use TRACE, here.

java/org/apache/catalina/filters/RestCsrfPreventionFilter.java
```diff
@@ -192,6 +209,10 @@ private String extractNonceFromRequestParams(HttpServletRequest request) {
                 String nonce = params[0];
                 for (String param : params) {
                     if (!Objects.equals(param, nonce)) {
+                        if (log.isErrorEnabled()) {
```
ChristopherSchultz, line 212: Same: use DEBUG not ERROR.

ChristopherSchultz (changes requested):
The PR looks good, but I have some suggestions for improvement.

java/org/apache/catalina/filters/RestCsrfPreventionFilter.java
```diff
@@ -217,6 +238,10 @@ public boolean apply(HttpServletRequest request, HttpServletResponse response) {
                 }
                 storeNonceToResponse(response, Constants.CSRF_REST_NONCE_HEADER_NAME,
                         nonceFromSessionStr);
+                if (log.isDebugEnabled()) {
+                    log.debug("CSRF Fetch request is succesfully handled - nonce is added to the response."
```
martin-g, line 242: succes`s`fully

java/org/apache/catalina/filters/RestCsrfPreventionFilter.java
martin-g: I agree with @ChristopherSchultz !

markt-asf:
It has been 2 months and the requested changes have not been made. I made the changes and a number of other fixes and applied this manually.

PR #453, opened 2021-10-08
AsyncFileHandler: try to flush on close

t-gergely:
When shutting down Tomcat on a single core server, some log entries can be lost. (E.g. the last ones added by `contextDestroyed`.) This patch tries to prevent that. It's not very elegant, but it works for me.

(Should you want to rewrite `AsyncFileHandler` instead, you might also want to drop the `org.apache.juli.AsyncLoggerPollInterval` property, since using that doesn't make any sense to me.)

### DIFF
--- a/java/org/apache/juli/AsyncFileHandler.java
+++ b/java/org/apache/juli/AsyncFileHandler.java
@@ -87,6 +87,8 @@ public class AsyncFileHandler extends FileHandler {
         if (closed) {
             return;
         }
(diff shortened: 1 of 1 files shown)


markt-asf:
Thanks for the PR. I applied a slightly different fix with an eye to a larger refactoring that would start/stop the thread as handlers were open/closed.
I also removed the system property. I agree there is no need for it.

PR #455, opened 2021-10-19
Optimize the comment about class loader

tianshuang:
In HotSpot JVM, javaseClassLoader is actually Extension ClassLoader, The comment of javaseClassLoader mentioned:
```java
    /**
     * The bootstrap class loader used to load the JavaSE classes. In some
     * implementations this class loader is always <code>null</code> and in
     * those cases {@link ClassLoader#getParent()} will be called recursively on
     * the system class loader and the last non-null result used.
     */
    private ClassLoader javaseClassLoader;
```
So I think it is more appropriate to use bootstrap classLoader in the comments.


markt-asf:
Thanks for the PR. I agree bootstrap is a better description of the class loader being used.

PR #460, opened 2021-11-22
Cookie updates for Servlet 6.0. Only support RFC 6265. Remove others.

markt-asf:
This removes support for Netscape cookies and RFC 2109 cookies.


rmaucher:
I think there will be a lot of complaints ;)

markt-asf:
I suspect you are right. I suspect this change will mean a longer than otherwise beta period for 10.1.x. It may be we end up adding a few more options to `Rfc6265CookieProcessor`. I'm planning on looking back through the archives to see if there are any obvious changes we'd want to make now.

ChristopherSchultz:
Just let MSIE die already. :) Oh, and all those custom clients whose devs never bothered to look at the HTTP or Cookie specs.

markt-asf:
The forced push just updated any since values to include "Servlet"

PR #461, opened 2021-12-06
Add null check for BaseNCodec decodning

Mikhail2048:
Add null check for BaseNCodec decodning

### DIFF
--- a/java/org/apache/tomcat/util/codec/binary/BaseNCodec.java
+++ b/java/org/apache/tomcat/util/codec/binary/BaseNCodec.java
@@ -373,10 +373,14 @@ public abstract class BaseNCodec {
      *
      * @param pArray
      *            A String containing Base-N character data
-     * @return a byte array containing binary data
+     * @return a byte array containing binary data, or
+     *         {@literal null} if the passed pArray is null
      */
     public byte[] decode(final String pArray) {
-        return decode(StringUtils.getBytesUtf8(pArray));
+        if (pArray != null) {
+            return decode(pArray.getBytes(StandardCharsets.UTF_8));
+        }
+        return pArray;
     }
 
     /**


markt-asf:
Why?

Mikhail2048:
Becuase this code, for example : `new Base64().decode(myString)` will produce NPE in case of myString is `null`. This is becuase of inner delegation to `public byte[] decode(final byte[] pArray)` method (in `BaseNCodec` class), which cannot handle null values.

BlueSkyT:
这是一封自动回复邮件。已经收到您的来信，我会尽快回复。

zhengdi1992:
<自动回复>
你好，你的来信我已收到，尽快会给你回复。

markt-asf:
So don't call `decode((String) null)`.

gmshake:
> Becuase this code, for example : `new Base64().decode(myString)` will produce NPE in case of myString is `null`. This is becuase of inner delegation to `public byte[] decode(final byte[] pArray)` method (in `BaseNCodec` class), which cannot handle null values.

It make sense to `encode` a empty string, but it does not make sense to `encode` or `decode` a `null` string.

Mikhail2048:
But guys, I am, as a framework user, expect that the framework will take care of this check. Do you really think I am doing the  
`decode((String) null)` - I have a string in application, which can be null, and I, as a framework user, do not want to take care about making checks about it.

Indeed, fellows, I guess you have used Apache Commons. Take a look into its utility classes - they withdraw this boring responsibilities from the framework client.

Mikhail2048:
> > Becuase this code, for example : `new Base64().decode(myString)` will produce NPE in case of myString is `null`. This is becuase of inner delegation to `public byte[] decode(final byte[] pArray)` method (in `BaseNCodec` class), which cannot handle null values.
> 
> It make sense to `encode` a empty string, but it does not make sense to `encode` or `decode` a `null` string.

I agree with that, but it forces the client the be aware of the fact that string is not null. I am responsible to check it, as a Client - why? I genuinely do not understand the logic behind it

markt-asf:
1. Please avoid use of gender specific terminology such as "guys" or "fellows" when referring to members of the Tomcat community.

2. This code is not part of a framework. It is Tomcat's internal API. Applications should not be calling this code.

The logic behind the current design is simple: Tomcat never calls that method with `null` so a null check is unnecessary.

Mikhail2048:
Ok, I am sorry, I will reference you directly :) I disagree with you, @markt-asf. The class Base64 is used **_directly_** by the clients. Just a couple of examples: [this](https://stackoverflow.com/questions/10744715/how-to-base64-encode-a-java-object-using-org-apache-commons-codec-binary-base64/10747284), and [this](https://www.baeldung.com/java-base64-encode-and-decode#encodingdecoding-using-apache-commons-code), also [this](https://stackoverflow.com/questions/11544568/decoding-a-base64-string-in-java) and e.t.c.) - and, as I already mentioned, Base64 delegates to BaseNCodec, which does not tolerate null.

@markt-asf I really do not understand why it is so hard to apply this minor, non-breaking change :) As I mentioned above, people are using this API in thier code (it is public b.t.w), so why not make their life easier?)

markt-asf:
All those references you provided are for `org.apache.commons.codec.binary.BaseNCodec` not `org.apache.tomcat.util.codec.binary.BaseNCodec`

That the class is public is irrelevant. It has to be public as it is a utility class used by various Tomcat packages. Tomcat's public API is documented in the RELEASE-NOTES.

The change won't be applied because it isn't necessary and there is no desire to add unnecessary code bloat to Tomcat.

It your application wants to use Commons Codec then it should package it in WEB-INF/lib.

Mikhail2048:
Ok, that make sence. I will use the public API

gmshake:
> Ok, I am sorry, I will reference you directly :) I disagree with you, @markt-asf. The class Base64 is used **_directly_** by the clients. Just a couple of examples: [this](https://stackoverflow.com/questions/10744715/how-to-base64-encode-a-java-object-using-org-apache-commons-codec-binary-base64/10747284), and [this](https://www.baeldung.com/java-base64-encode-and-decode#encodingdecoding-using-apache-commons-code), also [this](https://stackoverflow.com/questions/11544568/decoding-a-base64-string-in-java) and e.t.c.) - and, as I already mentioned, Base64 delegates to BaseNCodec, which does not tolerate null.
> 
> @markt-asf I really do not understand why it is so hard to apply this minor, non-breaking change :) As I mentioned above, people are using this API in thier code (it is public b.t.w), so why not make their life easier?)

@Mikhail2048 , as @markt-asf has emphasised:
> This code is not part of a framework. It is Tomcat's internal API. Applications should not be calling this code.

The examples you referenced is `org.apache.commons.codec.binary.Base64`, that is not this one `org.apache.tomcat.util.codec.binary.BaseNCodec`. Be aware the two packages are different.

ChristopherSchultz:
@Mikhail2048 Your patch also doesn't even compile, so it wouldn't have been accepted, anyway.

PR #462, opened 2021-12-25
Add support for password-based encryption scheme 2 params (PBES2)

xiezhaokun:
Add support for password-based encryption scheme 2 params (PBES2)
https://bz.apache.org/bugzilla/show_bug.cgi?id=65767

### DIFF
--- a/java/org/apache/tomcat/util/net/jsse/PEMFile.java
+++ b/java/org/apache/tomcat/util/net/jsse/PEMFile.java
@@ -23,6 +23,7 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.math.BigInteger;
 import java.nio.charset.StandardCharsets;
+import java.security.AlgorithmParameters;
 import java.security.GeneralSecurityException;
 import java.security.InvalidKeyException;
 import java.security.KeyFactory;
@@ -59,6 +60,8 @@ public class PEMFile {
 
     private static final byte[] OID_EC_PUBLIC_KEY =
             new byte[] { 0x06, 0x07, 0x2A, (byte) 0x86, 0x48, (byte) 0xCE, 0x3D, 0x02, 0x01 };
+
+    private static final String OID_PKCS5_PBES2 = "1.2.840.113549.1.5.13";
 
     public static String toPEM(X509Certificate certificate) throws CertificateEncodingException {
         StringBuilder result = new StringBuilder();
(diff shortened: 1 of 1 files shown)


xiezhaokun:
https://bz.apache.org/bugzilla/show_bug.cgi?id=65767

java/org/apache/tomcat/util/net/jsse/PEMFile.java
```diff
@@ -210,6 +214,14 @@ public PrivateKey toPrivateKey(String password, String keyAlgorithm, Format form
             throw exception;
         }
 
+        private String getPBEAlgorithm(EncryptedPrivateKeyInfo privateKeyInfo) {
+            AlgorithmParameters parameters = privateKeyInfo.getAlgParameters();
+            if(parameters != null && OID_PKCS5_PBES2.equals(privateKeyInfo.getAlgName())) {
+                return parameters.toString();
```
ChristopherSchultz, line 220: This seems fragile. The javadoc for `toString` doesn't say anything about the format of the resulting String. Will it always be acceptable to `Cipher.getInstance`? Is it acceptable in every case we can currently test?

xiezhaokun: If use PBES2 , the parameters is  PBES2Parameters and engineToString return pbes2AlgorithmName .

```

 @SuppressWarnings("deprecation")
    protected void engineInit(byte[] encoded)
        throws IOException
    {
        String kdfAlgo = null;
        String cipherAlgo = null;

        DerValue pBES2_params = new DerValue(encoded);
        if (pBES2_params.tag != DerValue.tag_Sequence) {
            throw new IOException("PBE parameter parsing error: "
                + "not an ASN.1 SEQUENCE tag");
        }
        DerValue kdf = pBES2_params.data.getDerValue();

        // Before JDK-8202837, PBES2-params was mistakenly encoded like
        // an AlgorithmId which is a sequence of its own OID and the real
        // PBES2-params. If the first DerValue is an OID instead of a
        // PBES2-KDFs (which should be a SEQUENCE), we are likely to be
        // dealing with this buggy encoding. Skip the OID and treat the
        // next DerValue as the real PBES2-params.
        if (kdf.getTag() == DerValue.tag_ObjectId) {
            pBES2_params = pBES2_params.data.getDerValue();
            kdf = pBES2_params.data.getDerValue();
        }

        kdfAlgo = parseKDF(kdf);

        if (pBES2_params.tag != DerValue.tag_Sequence) {
            throw new IOException("PBE parameter parsing error: "
                + "not an ASN.1 SEQUENCE tag");
        }
        cipherAlgo = parseES(pBES2_params.data.getDerValue());

        pbes2AlgorithmName = new StringBuilder().append("PBEWith")
            .append(kdfAlgo).append("And").append(cipherAlgo).toString();
    }

    /*
     * Returns a formatted string describing the parameters.
     *
     * The algorithn name pattern is: "PBEWith<prf>And<encryption>"
     * where <prf> is one of: HmacSHA1, HmacSHA224, HmacSHA256, HmacSHA384,
     * or HmacSHA512, and <encryption> is AES with a keysize suffix.
     */
    protected String engineToString() {
        return pbes2AlgorithmName;
    }
```

xiezhaokun: You can also refer to the mapPBEParamsToAlgorithm method of PKCS12KeyStore in JRE.

markt-asf: Having looked at the complexity of extracting the algorithm name from from encoded parameters, and given that the JRE depends on the toString() implementation as well in places, I think it is reasonable to use toString() for now. We'll always have the option of parsing the parameters directly if we discover problems with the toString() approach in the future.
Given PBES2 support is currently broken, the worst this change will do is maintain the status quo. The more likely result is the PBES2 will be supported for most (all?) use cases.

markt-asf:
I've applied this PR manually so I could add an extended comment next to the use of toString()

xiezhaokun:
https://github.com/apache/tomcat/commit/d871a229da1408520614c87fd1f50af5c0d639ca

PR #468, opened 2022-01-27
add test sub targets depend on setup-jacoco

chenyonghui:
if run test-* targets,eg ant test-nio, get errors
build.xml:1896: Problem: failed to create task or type antlib:org.jacoco.ant:coverage


markt-asf:
Thanks.

Applied manually so I could start with main and then backport. I also added a changelog entry that included crediting you.

PR #470, opened 2022-01-28
Load "javax.annotation.*" classes from common classloader

ppkarwasz:
This is the `javax` version of #469.

Some users mistakenly add `javax.annotation-api` to their applications, which prevents Tomcat from finding `@Resource` and similar annotations on Java 9+.

To prevent this "javax.annotation" should also be always loaded from the _common_ classloader.

On Java 8 and earlier classes from `javax.annotation` where served from the _system_ classloader, so the change should be a no-op.


markt-asf:
Back-ported from 10.0.x (with appropriate changes to package name)

PR #476, opened 2022-02-18
EL FunctionMapperImpl: skip Method lookup if not required

tandraschko:
we currently have a weird case, where it seems that our session-manager seems to use the wrong CL (memcached-session-manager), and therefore a ClassNotFoundException is logged.
Nevertheless, it only happens on the second request, where the method infos are already loaded.

So this PR improves the performance a bit as the lookup isnt done anymore for cases, where the infos are already loaded and it fixes our special case.

### DIFF
--- a/java/org/apache/el/lang/FunctionMapperImpl.java
+++ b/java/org/apache/el/lang/FunctionMapperImpl.java
@@ -122,18 +122,27 @@ public class FunctionMapperImpl extends FunctionMapper implements
         public void writeExternal(ObjectOutput out) throws IOException {
             out.writeUTF((this.prefix != null) ? this.prefix : "");
             out.writeUTF(this.localName);
-            // make sure m isn't null
-            getMethod();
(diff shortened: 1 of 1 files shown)


tandraschko:
i would also be happy if you could backport it to 9.0.x

markt-asf:
I do think you should try and find the root cause of the issue you are observing as it might be causing other issues.

I'm not convinced the proposed change is the right one. There should be no need to call `getMethod()` at all. If one of `owner`, `name` or `types` is `null` and `m` is `null`, calling `getMethod()` is just going to trigger an error. And `m` should not be `null` in that case anyway.

Looking at the history, it appears two separate fixes were applied for the same issue:
- https://bz.apache.org/bugzilla/show_bug.cgi?id=41797
- https://bz.apache.org/bugzilla/show_bug.cgi?id=44428

I'm going to refactor this and implement a different fix.

tandraschko:
@markt-asf thanks, even better!

PR #478, opened 2022-02-24
ignore output directory target after building

Welmanhotheart:
considering that many developer may use maven to rebuild their local project and then produce the output directory 'target', so add it to .gitignore


markt-asf:
This won't be applied as a) Tomcat doesn't use Maven to build and b) the comment in the file

PR #480, opened 2022-03-03
Fix typo

hakanbakacak:


ChristopherSchultz:
IMO not worth it on its own. Removing a single space? Not worth a change to the file history.

PR #481, opened 2022-03-09
changing HashMap to LinkedHashMap to maintain execution order of deployment

pavandv9:
currently deployment order is not maintain, in order to overcome this we can change HashMap to LinkedHashMap.
So that the deployment order in <context> maintained.

### DIFF
--- a/java/org/apache/catalina/core/ContainerBase.java
+++ b/java/org/apache/catalina/core/ContainerBase.java
@@ -159,7 +159,7 @@ public abstract class ContainerBase extends LifecycleMBeanBase implements Contai
     /**
      * The child Containers belonging to this Container, keyed by name.
      */
-    protected final HashMap<String, Container> children = new HashMap<>();
+    protected final LinkedHashMap<String, Container> children = new LinkedHashMap<>();
 
 
     /**


pavandv9:
Can someone review this.

markt-asf:
That is a backwards incompatible change so, if made, it will only be made to the 10.1.x branch.

Web applications are meant to be independent. If there are dependencies between two or more web applications then those applications should be coded so that they correctly handle starting in different orders.

Generally, the Tomcat project does not implement features that enforce this sort of ordering as it undermines portability.

I am close to -1 on this proposal but will leave the PR open for others to comment.

pavandv9:
I agree to your point, but again since web applications are independent many applications may be running in order.

So I hope by doing this it doesn't affect the existing feature and it will add more advantage to it.

ChristopherSchultz:
What exactly does this accomplish? LinkedHashMap just produces iterators which maintain insertion-order. It doesn't impose any order on when they were inserted.

Is this intended to maintain _redeployment_-order when auto-deployment is enabled and all applications are updated simultaneously? If so, it's entirely possible that your application updates would be processed across multiple background-processing executions and therefore wouldn't necessarily be redeployed in the order you expected, anyway.

I must not understand the use-case, here.

markt-asf:
No response to Chris's questions so closing this as WONTFIX. Maintaining deployment order of applications is something probably best discussed on the users list in the first instance.

PR #482, opened 2022-03-16
replace deprecated getSubjectDN()

sim-lus:
According to java spec, getSubjectDN() should not be used:
"denigrated, replaced by getSubjectX500Principal(). This method returns the subject as an implementation specific Principal object, which should not be relied upon by portable code."

Not sure if my proposed change is correct, at least it is probably not compatible for all use cases.


markt-asf:
There is an issue here. `getSubjectDN()` returns a DN that includes spaces between the attribute name/value pairs. `getSubjectX500Principal().getName()` does not include spaces. That is going to create backwards compatibility issues. This is going to need a re-think. I'll follow-up on the dev@ list

markt-asf:
Ah. Problem solved. Just need to specify RFC 1779 format.

PR #484, opened 2022-03-17
Change the logic for judging whether a jsp file is outdated

F4ded:
Bug fix: after a jsp file is accessed, modify the jsp file to make it have a syntax error, when you accessing the jsp file again, the response will report a synatx error for the first time, but if it is accessed again within 4s(tomcat default config `modificationTestInterval`), you will get the reponse of the jsp file before modification.

if a jsp file have a syntax error, its `JspServletWrapper` will carry a `compileException` which will be throw in `jspCompiler.compile()`, making `theServlet` can not reload. if you access the jsp file modified, tomcat will reset the `lastModificationTest` to the current system time. within 4s, tomcat will judge the jsp file with syntax errors as not outdated, then the request will be handled by the old jsp file's `theServlet`.

before fixing: [before](https://raw.githubusercontent.com/F4ded/blog-pic/master/iloveimg-compressed/before.gif)
afiter fixing: [after](https://raw.githubusercontent.com/F4ded/blog-pic/master/iloveimg-compressed/after.gif)

### DIFF
--- a/java/org/apache/jasper/compiler/Compiler.java
+++ b/java/org/apache/jasper/compiler/Compiler.java
@@ -448,7 +448,7 @@ public abstract class Compiler {
 
             if (jsw.getLastModificationTest()
                     + (ctxt.getOptions().getModificationTestInterval() * 1000) > System
-                    .currentTimeMillis()) {
+                    .currentTimeMillis() && jsw.getCompileException() == null) {
                 return false;
             }
             jsw.setLastModificationTest(System.currentTimeMillis());
--- a/java/org/apache/jasper/servlet/JspServletWrapper.java
+++ b/java/org/apache/jasper/servlet/JspServletWrapper.java
@@ -215,6 +215,10 @@ public class JspServletWrapper {
         return ctxt.getServletContext();
     }
 
+    public JasperException getCompileException() {
+        return compileException;
+    }
+
     /**
      * Sets the compilation exception for this JspServletWrapper.
      *


ChristopherSchultz:
I've noticed this behavior in the past and never bothered to track it down to the source.

I'm having trouble loading the images, though I suspect they don't have much important information in them.

ChristopherSchultz:
I would appreciate it if someone who understands the Jasper components of Tomcat better than I do could please evaluate this patch for correctness. It looks good to me, but I haven't reproduced the problem and tested the solution. It seems simple enough that a code-inspection would suffice as a review and inclusion... if the code is well-understood.

F4ded:
> 
Here is the details.

org.apache.jasper.compiler.Compiler#isOutDated(boolean):
```java
public boolean isOutDated(boolean checkClass) {

        if (jsw != null
                && (ctxt.getOptions().getModificationTestInterval() > 0)) {

            if (jsw.getLastModificationTest()
                    + (ctxt.getOptions().getModificationTestInterval() * 1000) > System
                    .currentTimeMillis()) {
                return false;
            }
            jsw.setLastModificationTest(System.currentTimeMillis());
        }
    //...
}
```
- `jsw.getLastModificationTest()` default value is 0 
- `ctxt.getOptions().getModificationTestInterval()` default value is 4

After modifying the jsp file, condition `if(jsw.getLast ...... > System.currentTimeMillis())` is false when the jsp file is accessed for the first time, then `jsw.setLastModificationTest(System.currentTimeMillis())` will be executed.

So within 4 seconds we access this jsp page again, the condition `if(jsw.getLast ...... > System.currentTimeMillis())` will be true and the function `isOutdated` will return false:
image

The statement `try{..}` will not be executed, when the function `compile()` returns to `JspServletWrapper#service`, the `
compileException` will not be thrown, then function `getServlet()` will return a servlet which corresponds to the old jsp file, request will be handled by it, so the we get the response of old jsp file:
image

F4ded:
> I've noticed this behavior in the past and never bothered to track it down to the source.
> 
> I'm having trouble loading the images, though I suspect they don't have much important information in them.

I have updated the links to these images, they are available now.

kkolinko:
I think that the feature that this PR tries to remove may protect Tomcat from a DOS attack.

Consider, that an instance of Apache Tomcat is running in production and has a JSP that fails to compile. Does every request trigger a recompilation, or only one in 4 seconds? How much load does it cause?

The docs suggest to turn off recompilation of JSPs when running in production, but many people do not bother to change their configuration. So a DOS is a concern.

If this PR is closed in anticipation of a different fix, I think it may be good to file this as an issue (enhancement request) into Bugzilla.

Regarding the patch itself,  line 451 in Compiler.java:
I think that the "jsw.getCompileException() == null" check is cheaper than a "System.currentTimeMillis()" call and thus should come first.

kkolinko:
I think that it is much better to discuss bugs in Bugzilla. So that an issue is discussed in one place. Regardless of how may PRs are there.

F4ded:
> I think that the feature that this PR tries to remove may protect Tomcat from a DOS attack.
> 
> Consider, that an instance of Apache Tomcat is running in production and has a JSP that fails to compile. Does every request trigger a recompilation, or only one in 4 seconds? How much load does it cause?
> 
> The docs suggest to turn off recompilation of JSPs when running in production, but many people do not bother to change their configuration. So a DOS is a concern.
> 
> If this PR is closed in anticipation of a different fix, I think it may be good to file this as an issue (enhancement request) into Bugzilla.
> 
> Regarding the patch itself, line 451 in Compiler.java: I think that the "jsw.getCompileException() == null" check is cheaper than a "System.currentTimeMillis()" call and thus should come first.

I seem to realize the problem with this commit, if a jsp file is modified to have a syntax error, after adding `jsw.getCompileException() == null`, the func `isOutdated` will always return true, so every request will trigger a recompilation, which will take up some resources.

There might be a better solution, I will try it later.

markt-asf:
I am closing this as WONTFIX.
The current default behaviour is by design for DoS protection.
If you want the behaviour this PR is intended to provide set the `recompileOnFail` option to `true`.

PR #487, opened 2022-03-23
Improved logging of unknown http2 settings

HoffmannTom:
The logging of unknown http2 settings should contain the key and the value of the setting.
The key is currently converted into a setting object. Unknown keys are mapped to Integer.MAX_VALUE
Thus the log currently contains the MAX_VALUE as key and the value of the settings. The original key got lost during the conversion to the setting object.

Therefore the logging was moved to methods, which have access to the original key and the value.


rmaucher:
Closing since the patch is wrong and I'm already about to commit it.

rmaucher:
Oh, actually, this has the key -> id change, but it should still do Integer.toString(id).

HoffmannTom:
Thank you for the quick response and help!
My second pull request in future will hopefully improve.

PR #488, opened 2022-03-24
Donating EL Translations

volosied:
Open Liberty uses Tomcat's EL API, but, since some of the property file translations were missing,  IBM completed them and also created new files for other languages. We'd like to contribute these back to Tomcat. 

See https://github.com/OpenLiberty/open-liberty/tree/integration/dev/io.openliberty.jakarta.expressionLanguage.4.0/resources/jakarta/el

Thanks.


markt-asf:
The files use unicode escapes rather than UTF-8. They should use UTF-8 as this makes them easier for native speakers to review.

Tomcat uses POEditor to [manage translations](https://cwiki.apache.org/confluence/x/vIPzBQ). POEditor is our preferred route for receiving translation updates. Given the volume of updates here, we can jump through the additional hoops necessary to accept these contribution as a PR but would ask that any future updates go via POEditor.

markt-asf:
Please rebase the PR. The "writeable" -> "writable" change was made last year.

volosied:
Hey Mark,  these translations where done with files prior to the `writeable" -> "writable` change (which I had updated manually for this PR)  I squashed the two commits. 

I went ahead an made a new commit which converted the unicode escapes characters to UTF8.

markt-asf:
Thanks. I ended up applying this manually as it was easier to sync with POEditor that way.

volosied:
@markt-asf  Quick question -- I see some translations are missing or am I not looking at this right? 
For example, I see an Italian translation is in the POEditor: 
image

But I don't see the Italian `LocalStrings_it.properties` file here: https://github.com/apache/tomcat/tree/main/java/jakarta/el


Basically, I don't see the following EL files in Tomcat: 
LocalStrings_ro.properties (Romanian)
LocalStrings_pl.properties (Polish)
LocalStrings_hu.properties (Hungarian)
LocalStrings_it.properties (Italian)
LocalStrings_zh_TW.properties (Traditional Chinese )
LocalStrings_zh.properties (Chinese)

markt-asf:
There are a couple of historical exceptions but we generally only pull in a new language once ~10% of the terms are translated.
The Chinese translations are in the `LocalStrings_sh_CN.properties` file.
We could potentially pull in the translations so they are in the source but not build a JAR for that language. Would that help?
We could also think about shipping some/all i18n JARs separately so users would download the core Tomcat distribution and then add any additional language(s) they wanted. That would be a discussion for the dev list.

PR #489, opened 2022-03-29
Suppress null socket channel warning for closed sockets

rhernandez35:
Null socket channels are only a problem if the socket isn't closed. Otherwise, they're just indicative of a stale event being processed.


markt-asf:
This is not correct. Null socket channels are usually an indication that something has gone wrong. The warning is deliberate.
If you have a use case where a correctly handled request triggers this warning then please raise this on the Tomcat users mailing list in the first instance.

PR #490, opened 2022-04-01
Replace non-existent com.h2database version.

dsoumis:
com.h2database version 2.0.210 does not exist in maven central repository. It should be replaced with version 2.1.210 which is flawless.


markt-asf:
Thanke for spotting this.

PR #493, opened 2022-04-06
Do not use legacy character encoding(Shift_JIS) as default

be-hase:
## Motivation

With this commit, `Shift_JIS` is now used as the default for `jp`.
eda77cd88b84d6002d9efb818e6d3fbc241e2284

`Shift_JIS` is a legacy character code, and this change causes a bug in the behavior of the application at Japan.

For example, Spring Boot has the following support due to this commit.
https://github.com/spring-projects/spring-boot/pull/30535
However, I think it is better to revert the commit on the tomcat side as well.

## What

Revert eda77cd88b84d6002d9efb818e6d3fbc241e2284

### DIFF
--- a/java/org/apache/catalina/util/CharsetMapperDefault.properties
+++ b/java/org/apache/catalina/util/CharsetMapperDefault.properties
@@ -15,4 +15,3 @@
 
 en=ISO-8859-1
 fr=ISO-8859-1
-ja=Shift_JIS
--- a/webapps/docs/changelog.xml
+++ b/webapps/docs/changelog.xml
@@ -339,11 +339,6 @@
         list of JARs to skip when scanning for TLDs, web fragments and
         annotations. (michaelo)
       </add>
-      <add>
-        Expand the default mappings used by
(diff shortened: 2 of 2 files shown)


markt-asf:
This needs careful thought as the mapping was added to address a TCK failure. We need to look into the full history to figure out the right fix.

be-hase:
Sorry, I'm not familiar with full history. I would be grateful if you could consider this issue.

be-hase:
[off topic]
For [Spring4Shell](https://spring.io/blog/2022/03/31/spring-framework-rce-early-announcement), I did a version up of Spring Boot. Then I met this `Shift_JIS` problem. 😭

markt-asf:
I am going to revert this change but simply reverting the commit also reverts the changelog which I don't want. I'll do a manual revert. The TCK issue can be addressed via configuration in the global web.xml. I'm closing this PR as I need to do a manual revert.

be-hase:
Thanks your handling :)

PR #496, opened 2022-04-08
Fix Tomcat Versions wiki url

leggiero:
If approved, I'll open PR's with the same change for the 10.0.x, 9.0.x and 8.5.x that have the same issue.


markt-asf:
This should be changed to the canonical URL which is:
https://cwiki.apache.org/confluence/display/TOMCAT/Tomcat+Versions

In the meantime, I have added a redirect so users don't see a 404.

markt-asf:
FYI, no need for one PR per branch unless the patch doesn't apply cleanly for some reason. Generally, we merge PRs ot main and then cherry-pick them to the earlier branches.

markt-asf:
Thanks for spotting this. I did a more global update of wiki URLs to use the new wiki location.

PR #501, opened 2022-04-11
Support max http request header size and max http header size configuration separation

CrazyHZM:
Mail link：https://bz.apache.org/bugzilla/show_bug.cgi?id=65866

### DIFF
--- a/java/org/apache/coyote/http11/AbstractHttp11Protocol.java
+++ b/java/org/apache/coyote/http11/AbstractHttp11Protocol.java
@@ -244,6 +244,34 @@ public abstract class AbstractHttp11Protocol<S> extends AbstractProtocol<S> {
     public void setMaxHttpHeaderSize(int valueI) { maxHttpHeaderSize = valueI; }
 
 
+    /**
+     * Maximum size of the HTTP request message header.
+     */
+    private int maxHttpRequestHeaderSize = -1;
+
+    public int getMaxHttpRequestHeaderSize() {
+        return maxHttpRequestHeaderSize == -1 ? getMaxHttpHeaderSize() : maxHttpRequestHeaderSize;
+    }
+
+    public void setMaxHttpRequestHeaderSize(int valueI) {
+        maxHttpRequestHeaderSize = valueI;
+    }
+
+
+    /**
+     * Maximum size of the HTTP response message header.
+     */
+    private int maxHttpResponseHeaderSize = -1;
+
+    public int getMaxHttpResponseHeaderSize() {
+        return maxHttpResponseHeaderSize == -1 ? getMaxHttpHeaderSize() : maxHttpResponseHeaderSize;
+    }
+
+    public void setMaxHttpResponseHeaderSize(int valueI) {
+        maxHttpResponseHeaderSize = valueI;
+    }
+
+
     private int connectionUploadTimeout = 300000;
     /**
      * Specifies a different (usually longer) connection timeout during data
--- a/java/org/apache/coyote/http11/Http11Processor.java
+++ b/java/org/apache/coyote/http11/Http11Processor.java
@@ -158,11 +158,11 @@ public class Http11Processor extends AbstractProcessor {
         httpParser = new HttpParser(protocol.getRelaxedPathChars(),
                 protocol.getRelaxedQueryChars());
 
-        inputBuffer = new Http11InputBuffer(request, protocol.getMaxHttpHeaderSize(),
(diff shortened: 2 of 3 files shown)


CrazyHZM:
Please review the code first, if there is no problem, I will support this on other branches as well.

java/org/apache/coyote/http11/AbstractHttp11Protocol.java
```diff
@@ -244,6 +244,34 @@ public void setMaxSavePostSize(int maxSavePostSize) {
     public void setMaxHttpHeaderSize(int valueI) { maxHttpHeaderSize = valueI; }
 
 
+    /**
+     * Maximum size of the HTTP request message header.
+     */
+    private int maxHttpRequestHeaderSize = -1;
```
gmshake, line 250: After the change, `maxHttpHeaderSize` becomes default maximum size of the HTTP message header for both request and response.

If we no longer have consumer of `getMaxHttpHeaderSize()`, then I'd suggest to remove `maxHttpHeaderSize`, and change `setMaxHttpHeaderSize()` like this 

```java
@Deprecated
public void setMaxHttpHeaderSize(int valueI) {
    maxHttpRequestHeaderSize = valueI;
    maxHttpResponseHeaderSize = valueI;
}
```

rmaucher:
That seems pretty straightforward. I'm slightly skeptical about the real practical use since trying to optimize the sizes too much is risky, but no problem, this is the user's responsibility.

java/org/apache/coyote/http11/AbstractHttp11Protocol.java
CrazyHZM: I'm worried about the compatibility risk of removing getMaxHttpHeaderSize(), because this is a public API after all, I wonder if this change can be done in 10.1.x. of which I agree with the annotation Deprecated.

CrazyHZM:
> That seems pretty straightforward. I'm slightly skeptical about the real practical use since trying to optimize the sizes too much is risky, but no problem, this is the user's responsibility.

@rmaucher 
I saw some discussion about this, you can refer to it: 
https://github.com/spring-projects/spring-boot/issues/29382

rmannibucau:
+1 I see two advantages:

1. enables cases where "front" tomcat has big incoming headers but "normal" forwarded headers
2. enables better web server abstractions which often have this configuration

Open question: shouldn't `setMaxHttpHeaderSize` set both request and response ones and `maxHttpHeaderSize` field be dropped for backward compat?

java/org/apache/coyote/http11/AbstractHttp11Protocol.java
gmshake: Keep `getMaxHttpHeaderSize()` only preserves API level compatibility, the semantic of `getMaxHttpHeaderSize()` has change once we introduce dedicated header settings for request and response.

To illustrate the semantic change:
```java
setMaxHttpHeaderSize(4096);
setMaxHttpRequestHeaderSize(8192);
setMaxHttpResponseHeaderSize(16384);
```
Then `getMaxHttpHeaderSize()` make no sense except the only meaning "default maxHttpHeaderSize"

Maybe it is  deprecate `getMaxHttpHeaderSize()` and return a size that is not too bad:
```java
public int getMaxHttpHeaderSize() {
    return Math.min(maxHttpRequestHeaderSize, maxHttpResponseHeaderSize)
}
```

I still insist that semantic level compatibility is more important than API level compatibility. For `setMaxHttpHeaderSize()` it originally means set maximum size of the HTTP message header for both request and response.

rmaucher:
Overall, I prefer the PR as it is right now, it's simpler. If the new setters are used, then the behavior could be slightly different, but since this is new API this should be fine.

gmshake:
> Overall, I prefer the PR as it is right now, it's simpler. If the new setters are used, then the behavior could be slightly different, but since this is new API this should be fine.

It is not complicated. I forked and edit online to demonstrate the part of change: https://github.com/gmshake/tomcat/commit/75289cf9e7a3920cc6d3ea0463db53a6ab0d090c?diff=split

rmaucher:
The proposed patch is still simpler and should work well, so I will merge that one.

rmaucher (approved):

CrazyHZM:
@rmaucher 
I can patch several other branches, right?

rmaucher:
Merged with changed. I didn't understand the changes to TestHttp2Limits, HTTP/2 should use the new request header size and that should be it. Also updated docs and changelog.

CrazyHZM:
@rmaucher 
Do i need to resubmit a clean pr?

PR #504, opened 2022-04-14
disable jsp and jspx by default

k4n5ha0:
jsp and jspx is dangerous. likes spring4shell and others hacker,they use uplaod jsp or write a webshell to disk.
If project need jsp or jspx, they pack web.xml in war with jsp mappings by themself.
secure by default.
thx!

### DIFF
--- a/conf/web.xml
+++ b/conf/web.xml
@@ -442,13 +442,13 @@
         <url-pattern>/</url-pattern>
     </servlet-mapping>
 
-    <!-- The mappings for the JSP servlet -->
+<!-- The mappings for the JSP servlet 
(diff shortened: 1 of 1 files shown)


markt-asf:
This is a bad idea for so many different reasons. To name a few:

- "Spring4Shell" allows arbitrary file uploads. All an attacker has to do to bypass this change is to upload a web.xml file that re-enables the mapping
- It does nothing to help users that want/need to use JSPs.
- Users that had followed the documented security recommendations and set OS file permissions appropriately would have been protected not only against "Spring4Shell"but against any similar vulnerability as well

PR #505, opened 2022-04-25
JasperInitializer not correctly initialized when a Default JspFactory is already defined

jeanouii:
An issue as been introduced in JasperInitializer with commit https://github.com/apache/tomcat/commit/5e8eb5533f551c3dbc3003e4c2f4f0d2958a8eb3

When a default JspFactory is already defined the static field `defaultFactory` in JasperInitializer isn't initialized. It produces a NullPointerException line 108 when setting the poolSize.

This fix enforces a proper initialization even when the default JspFactory is already set.
Adding a small test checking the PoolSize can be set using the ServletContext and therefor the bug as been fixed.


rmaucher:
Adding this field was a bad idea. I'm removing it now instead.

PR #508, opened 2022-05-06
Cloudevents

jgutie2r:
I have made some changes to ease de development of event-driven serverless functions using embed Tomcat. I have used the API provided by CloudEvents.
With these changes you can develop a serverless function extending the class CloudEventListener and implementing the method consumeEvent(CloudEvent ev, HttpServletRequest, HttpServletResponse). An instance of this class can be registered in a Tomcat instance (as is done with a HttpServlet subclass).

A new method has been added to HttpServletResponse to send a CloudEvent as response to the request.

These changes do not break the API (provide backward compatibility), if I'm not wrong.

Besides I have added three filters (HealthFilter, MetricsFilter, TokenAuthFilter) and a class to register these filters. This do not affect the other changes (can be removed).

A script is included to install the jar in the local repository for maven.

I hope these changes are not out of scope of the project.
---
The code in this project has been used to develop functions that have been deployed on Knative.


markt-asf:
The Jakarta APIs are defined by the Jakarta specification projects. It is not permitted to change them in any way.

This PR has reverted multiple recent changes. That is not acceptable.

This PR contains a large number of unnecessary cosmetic changes that make it harder to review the significant changes.

This PR adds three additional JARS to the Tomcat distribution.

This PR will not be merged. If there are changes you would like to see that would make it easier to integrate the CouldEvents API these will be considered but the CouldEvents API isn't going to be shipped with Tomcat at this time.

PR #511, opened 2022-05-09
Allow to decrypt PEM keys.

jfclere:
Allow to decrypt PEM keys starting like:
 Proc-Type: 4,ENCRYPTED
DEK-Info: DES-EDE3-CBC,3EEF0D821F09467F

CRuSQwLsLRukfxzyTvl15rP4q+DGV3zTdD+vpZgpR/OiJvN2FAb/hLlMjo4RBE7m
...

### DIFF
--- a/java/org/apache/tomcat/util/net/jsse/PEMFile.java
+++ b/java/org/apache/tomcat/util/net/jsse/PEMFile.java
@@ -27,6 +27,7 @@ import java.security.AlgorithmParameters;
 import java.security.GeneralSecurityException;
 import java.security.InvalidKeyException;
 import java.security.KeyFactory;
+import java.security.MessageDigest;
 import java.security.PrivateKey;
 import java.security.cert.CertificateEncodingException;
 import java.security.cert.CertificateException;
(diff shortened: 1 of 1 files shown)


jfclere:
Note openssl can used to convert the key like:
-----BEGIN RSA PRIVATE KEY-----
Proc-Type: 4,ENCRYPTED
DEK-Info: AES-256-CBC,A337ABA05A83C84E225C79ACD675C6D2
to the PKCS#8 format something like:
openssl pkcs8 -topk8 -in server.key -out server.pkcs8.key

markt-asf:
+1 to broadening the support of our PEM file conversion.

I think the algorithm and the digest should be saved with the part for consistency with the rest of the code. I'm happy to do the refactoring to do that.

Refactoring made me think about tests to make sure i didn't break anything. I noticed we don't have any tests for `PEMFile` so I'm going to add some at the same time. I should have something to commit shortly.

markt-asf:
Mostly done but I'm looking at trying to make the support for encrypted PKCS1 files more generic rather than just supporting the one cipher.

jfclere:
@markt-asf OK Please close this PR when yours is ready (or when you have committed).

markt-asf:
Done. Not quite as generic as I would like but I couldn't figure out how to get from DEK-Info to JSSE standard names without a manual lookup.
Thanks for doing all the hard work on this. I just did a little refactoring

PR #515, opened 2022-05-18
Bug:Deadlocks occur when both engine and host startStopThreads values are set and the latter value is larger than the former

aooohan:
**_apache-tomcat-10.0.21 version has been reproduced, I have not tried other versions_**
# Reproduce
```xml
<Engine name="Catalina" defaultHost="localhost"  startStopThreads="3">
            unpackWARs="true" autoDeploy="true" startStopThreads="4">
      </Host>
</Engine>
```
**When starting tomcat as configured above, the situation shown in the image below will occur.**

image

**been stuck here**⬆️⬆️⬆️

**Then  i used the Arthas tool to look at the threads and the results are shown below.**

image
image
image

 When starting tomcat, the current StandardServer instance object has been locked in LifecycleBase#start. When we do not set the startStopThreads value of StandardEngine, startStopThreads defaults to 1. 
At this time, startStopExecutor in StandardEngine is actually an instance of InlineExecutorService (in fact, the current thread), ⬇️
https://github.com/apache/tomcat/blob/22ceccc5a3e1361e82bb4f10e2a9f4117714183d/java/org/apache/catalina/core/ContainerBase.java#L882
so when StandardEngine starts its subcontainer StandardHost through the startStopExecutor thread pool, because it is the current thread, there is no problem when executing the StandardServer#reconfigureUtilityExecutor method.
https://github.com/apache/tomcat/blob/22ceccc5a3e1361e82bb4f10e2a9f4117714183d/java/org/apache/catalina/core/StandardServer.java#L434
However, when the startStopExecutor in StandardEngine is not InlineExecutorService (ie Server#getUtilityExecutor()),
https://github.com/apache/tomcat/blob/22ceccc5a3e1361e82bb4f10e2a9f4117714183d/java/org/apache/catalina/core/ContainerBase.java#L888
 the StandardServer#reconfigureUtilityExecutor method is executed again（The host's startStopThreads value is greater than the engine's startStopThreads ）, because the lock of the StandardServer instance has not been released at this time, resulting in a deadlock.

> My English is limited, I don't know if I have explained it clearly, or I am thinking in the wrong direction

### DIFF
--- a/java/org/apache/catalina/core/StandardServer.java
+++ b/java/org/apache/catalina/core/StandardServer.java
@@ -187,6 +187,7 @@ public final class StandardServer extends LifecycleMBeanBase implements Server {
      * Utility executor with scheduling capabilities.
      */
     private ScheduledThreadPoolExecutor utilityExecutor = null;
+    private final Object utilityExecutorObject = new Object();
 
     /**
      * Utility executor wrapper.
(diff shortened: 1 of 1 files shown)


rmaucher:
Good idea, the sync is overloaded in that case so this is risky, and other operations already use a servicesLock item for this reason.

aooohan:
@rmaucher Thank u for reviewing my PR, it's an honour to contribute PR to Tomcat.

rmaucher:
Thanks a lot for your PR. Sorry, I forgot to close it after merging it manually. The fix will be in Tomcat 10.1.0-M16, 10.0.22, 9.0.64. Tomcat 8.5 is not affected since it doesn't have the utility executor feature.

PR #516, opened 2022-05-24
Bugfix for 66084 - bytesWritten in writeBytes() miscalculation for OutputBuffer

aooohan:


markt-asf:
Thanks for the PR. I'll add a change log entry and then back-port it.

PR #519, opened 2022-06-02
Polish. NioEndpoint

aooohan:
As I was learning tomcat, I discovered some logic that could be extracted into separate methods. Hope my pr is useful.


rmaucher:
Ok why not. interestOpsEquals is a bad name though, maybe interestOpsHas or something like it ?

aooohan:
> Ok why not. interestOpsEquals is a bad name though, maybe interestOpsHas or something like it ?
Thanks for the comment，interestOpsHas is better name than interestOpsEquals and i will update it.

aooohan:
> Ok why not. interestOpsEquals is a bad name though, maybe interestOpsHas or something like it ?

Thanks for your comment, interestOpsHas a better name than interestOpsEquals and I will update it.

markt-asf:
Applied manually go I could merge the polish commits and make a couple of minor changes.

PR #521, opened 2022-06-05
Bugfix for 66104 - Check for the existence of the log directory before executing the cleanup task at startup

aooohan:
fix:[https://bz.apache.org/bugzilla/show_bug.cgi?id=66104](https://bz.apache.org/bugzilla/show_bug.cgi?id=66104)

### DIFF
--- a/java/org/apache/juli/FileHandler.java
+++ b/java/org/apache/juli/FileHandler.java
@@ -538,7 +538,7 @@ public class FileHandler extends Handler {
     }
 
     private void clean() {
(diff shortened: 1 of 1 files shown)


markt-asf:
Thanks for the bug report and the PR.

I applied this manually so I could:
- refactor this code and the other location that obtains a Path from the directory
- used a Java 7+ compatible approach that could be back-ported and then changed this to the Java 11+ approach
- add a changelog entry

PR #523, opened 2022-06-11
Polished.

aooohan:


java/org/apache/el/parser/AstMethodParameters.java
```diff
@@ -39,15 +40,12 @@ public Object[] getParameters(EvaluationContext ctx) {
...
         // Purely for debug purposes. May not be complete or correct. Certainly
         // is not efficient. Be sure not to call this from 'real' code.
-        StringBuilder result = new StringBuilder();
-        result.append('(');
+        StringJoiner result = new StringJoiner(",","(",")");
         if (children != null) {
             for (Node n : children) {
-                result.append(n.toString());
-                result.append(',');
+                result.add(n.toString());
```
ChristopherSchultz, line 46: I'm not sure how much it matters, but the old implementation would have ended up with a trailing `,` after the final element, where the new implementation doesn't. Perhaps that's preferable; I didn't check out the use-case for the code.

aooohan: Hi Christopher,thank you for reviewing my pr. I think it's just for debug presentation purposes.It doesn't affect anything, whether it end up with a trailing `,` after the final element or not.

java/org/apache/catalina/core/ApplicationPushBuilder.java
```diff
@@ -26,6 +26,7 @@
 import java.util.Locale;
 import java.util.Map;
 import java.util.Set;
+import java.util.StringJoiner;
```
ChristopherSchultz, line 29: Just a note: this cannot be back-ported to Tomcat 8.5.x because it must be buildable and runnable on Java 7.

aooohan: Thanks, Christopher.

aooohan:
Merge manually.

PR #524, opened 2022-06-15
Dir redirect use301

funkman:
For the SEO folks (seemed to be a common ask which hasn't gone away)

When a directory redirect is requested. Allow a configuration setting to use a 301 HTTP response status instead of 302.


funkman:
yeah - I'll redo this as an int option instead.

funkman:
Merge latest apache/main in. Then additionally renamed parameter so it is transformed to an int and restricted it to 302 302 307 308

Did find one gap that I'm not sure if its worth fixing. There is a sendRedirect in CoyoteAdapter  (which defaults to 302) when the mapper detects a context. For example `/docs` will 302 to `/docs/`

markt-asf:
I like the look of the current diff.

My only concern is by not calling `sendRedirect()` we bypass a bunch of checks. My plan is to see if there is an easy way around that. If there is, I'll merge this with whatever that fix is. If there isn't, I think we'll need to postpone merging to the next release round while we think about it.

funkman:
Looking at `org.apache.catalina.connector.Response` - Yes `sendRedirect()` does quite a few safety checks first. So yeah - this might have bad side effects with those missing checks. (such as isCommitted )

I do notice a `sendRedirect(String location, int status)`  in `org.apache.catalina.connector.Response` which would be convenient to call instead and I just (re) checked https://javaee.github.io/javaee-spec/javadocs/javax/servlet/http/HttpServletResponse.html optimistically hoping `status` was a new option I missed.

markt-asf:
I was thinking of calling `sendRedirect()` and then explicitly setting the status if a different one was required. Possibly checking the location header to determine if the call to `sendRedirect()` was successful or not. Haven't tested it though. Which reminds me, we might want to add some test cases.

markt-asf:
I've added a test case. My initial idea isn't going to work because `sendRedirect()` commits the response. Adding a method to the Servlet API looks like the cleanest solution. But that only helps Tomcat 11 onwards. I have some ideas how this could work but they are all fairly ugly hacks. More thought required.

funkman:
closing due to open issues (don't want to it to hang) --- will rethink options as time permits

markt-asf:
A variation of this has been applied to 11.0.x as Servlet 6.1 has added additional methods to allow the status code to be specified in a sendRedirect() call.

PR #526, opened 2022-06-20
not report stack error information

mjs8002:
When a user enters a backslash (\) or other special characters in the URL of the account suspension page, the page returns error code 400 and stack error information；


markt-asf:
WONTFIX.

PR #527, opened 2022-06-26
easymock version change from 4.2 to 4.3

lakeoffaith:
when i run idea with the main branch of the tomcat,i found the version of  easymock  in the tomcat/res/ide-support/idea/tomcat.iml should be sync update.


markt-asf:
Thanks for the PR. Back-ported to 10.0.x, 9.0.x and 8.5.x as well.

lakeoffaith:
ok.i have already finished it.

PR #528, opened 2022-06-30
easymock version change from 4.2 to 4.3

lakeoffaith:


markt-asf:
This change has already been back-ported.

You can ignore Travis CI. It is unreliable and will be going away soon.

lakeoffaith:
so the upgrade version of easymock   on apache:10.0.x , what can i do to fix it , or just let it away.

lakeoffaith:
> This change has already been back-ported.
> 
> You can ignore Travis CI. It is unreliable and will be going away soon.




so the upgrade version of easymock on apache:10.0.x , what can i do to fix it , or just let it away.

PR #532, opened 2022-07-18
Don't perform protection checks in Unix Domain Socket mode

aooohan:
See [stackoverflow](https://stackoverflow.com/questions/72975059/tomcat-crashes-while-using-unixdomainsocket-as-a-connector) and #402 

8.5.x, 9.0.x, 10.0.x all have this problem.

### DIFF
--- a/java/org/apache/tomcat/util/net/AprEndpoint.java
+++ b/java/org/apache/tomcat/util/net/AprEndpoint.java
@@ -813,11 +813,11 @@ public class AprEndpoint extends AbstractEndpoint<Long,Long> implements SNICallB
             // Do the duplicate accept check here rather than in serverSocketaccept()
             // so we can cache the results in the SocketWrapper
             AprSocketWrapper wrapper = new AprSocketWrapper(socket, this);
-            // Bug does not affect Windows. Skip the check on that platform.
-            if (!JrePlatform.IS_WINDOWS) {
+            // Bug does not affect Windows platform and Unix Domain Socket. Skip the check.
+            if (!JrePlatform.IS_WINDOWS && getUnixDomainSocketPath() == null) {
                 long currentNanoTime = System.nanoTime();
                 if (wrapper.getRemotePort() == previousAcceptedPort) {
-                    if (previousAcceptedAddress != null && wrapper.getRemoteAddr().equals(previousAcceptedAddress)) {
+                    if (wrapper.getRemoteAddr().equals(previousAcceptedAddress)) {
                         if (currentNanoTime - previousAcceptedSocketNanoTime < 1000) {
                             throw new IOException(sm.getString("endpoint.err.duplicateAccept"));
                         }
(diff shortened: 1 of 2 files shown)


exabrial:
Just want to say thanks! Our initial testing shows that bypassing the TCP layer and using Unix sockets is _very significant_ performance increase.... like _a lot_ faster. Pretty danged cool! We'd probably terminate TCP/TLS with Haproxy then load balance to Tomcat listening on Unix sockets.

We still have a few more issues to overcome, but those are separate items or perhaps bugs, for another discussion:

* We noticed the the socket file doesn't seem to get cleaned up, despite the documentation indicating it should. As a workaround, we have the systemd unit remove the file after Tomcat stops. We are trying to root cause this in Tomcat code and see if we can figure out whats wrong.
* We noticed `request.getRemoteAddr()` and `request.getRemoteHost()` are broken (expected). We improvised a Valve to hardcode `127.0.0.1` for each and that seems to do the trick, then use the RemoteAddrValve to set the real values. A better option might be: Haproxy has something akin to AJP called "Proxy Protocol" that can pass a lot of information from the original request. There's a open bug about it that we might look into helping get across the finish line: https://bz.apache.org/bugzilla/show_bug.cgi?id=57830 

Anyway thank you! Very much appreciated!

aooohan:
> We noticed the the socket file doesn't seem to get cleaned up, despite the documentation indicating it should. As a workaround, we have the systemd unit remove the file after Tomcat stops. We are trying to root cause this in Tomcat code and see if we can figure out whats wrong.

@exabrial I also noticed this problem. I found that the automatic cleaning of UDS file was implemented in NioEndpoint, and I didn't find any logic about it in AprEndpoint.

rmaucher:
The cleanup in NIO is still limited (on purpose): it is only cleaned up if the socket was created successfully by the connector, and it anything fails really badly it will stay there.

aooohan:
> > We noticed the the socket file doesn't seem to get cleaned up, despite the documentation indicating it should. As a workaround, we have the systemd unit remove the file after Tomcat stops. We are trying to root cause this in Tomcat code and see if we can figure out whats wrong.
> 
> @exabrial I also noticed this problem. I found that the automatic cleaning of UDS file was implemented in NioEndpoint, and I didn't find any logic about it in AprEndpoint.

@exabrial  Aha, i found it! The cleanup function is implemented in tc-native, you can take a look at this
https://github.com/apache/tomcat-native/blob/a3498fa0992ac37c7358e00d1555395b52762e9b/xdocs/index.xml#L180
and https://github.com/apache/tomcat-native/commit/a3498fa0992ac37c7358e00d1555395b52762e9b

aooohan:
> The cleanup in NIO is still limited (on purpose): it is only cleaned up if the socket was created successfully by the connector, and it anything fails really badly it will stay there.

Thanks for your reply. Indeed, the cleanup does have limitations.

markt-asf:
Merged manually. Thanks for the PR.

markt-asf:
Note that 8.5.x is not affected by this issue.

PR #533, opened 2022-07-28
[Bug 66184] Ensure that all root loggers have a level or parent

ppkarwasz:
Since Java 8 the default level of a root logger is no longer set in the
constructor. This PR sets the level to INFO (LogManager's default) if it
is absent.

This patch should work on all the branches.


markt-asf:
Thanks for the bug report and PR. I just want to clean up a few minor points (e.g. add a change log entry) and then I'll back-port this.

PR #534, opened 2022-08-02
Issue 66195 observe mr jar system property

nobeh:
Observe if system property for MR JAR is available
    
    - Check if the JVM system property for MR-JAR is set
    - If set, parse the value as a boolean
    - If set, and when `#isMultiRelease()` is invoked, return the configured
      value
    
Refs #66195

### DIFF
--- a/java/org/apache/tomcat/util/compat/Jre9Compat.java
+++ b/java/org/apache/tomcat/util/compat/Jre9Compat.java
@@ -39,6 +39,15 @@ class Jre9Compat extends JreCompat {
 
     private static final Log log = LogFactory.getLog(Jre9Compat.class);
     private static final StringManager sm = StringManager.getManager(Jre9Compat.class);
+    /**
+     * @see <a href="https://docs.oracle.com/javase/9/docs/api/java/util/jar/JarFile.html">Multi-Release JAR API</a>
+     */
+    private static String JDK_MULTI_RELEASE_JAR_PROPERTY =
+        System.getProperty("jdk.util.jar.enableMultiRelease", null);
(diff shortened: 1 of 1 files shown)


nobeh:
https://bz.apache.org/bugzilla/show_bug.cgi?id=66195

aooohan:
Hi @nobeh ,

The `JarFile#isMultiRelease` method itself already implements the relevant logic, so we don't need to do anything extra with this property. In addition, we know from the documentation that the value of the `jdk.util.jar.enableMultiRelease` property is not just of the boolean type, but may also be `force`. So your implementation is also faulty.

nobeh:
> already implements the relevant logic, so we don't need to do anything extra with this property

Yes, that's correct. However, Tomcat uses a "MethodHandle" invocation to call that method. When a system is under load, through profiling, we've observed that invoking the method handle to evaluate to `false` could actually be an expensive operation. This is one of the main motivations for this change. 

> but may also be force

Yes. Thanks for the catch. I will update this.

nobeh:
> to evaluate to `false` could actually be an expensive operation. This is one of the main motivations for this change.

And the root cause for that is because of a `synchronized` block:

https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/jar/JarFile.java#L999

markt-asf:
The sync block won't be reached:
https://github.com/openjdk/jdk/blob/jdk-11%2B28/src/java.base/share/classes/java/util/jar/JarFile.java#L382

The case hasn't been made that this change is necessary nor has any evidence been provided of a genuine performance improvement.

PR #535, opened 2022-08-06
List duplicate locations when failing a deploy due to duplicated fragment names

rolsdorph:
For complex deployments with many dependencies, it's easy to lose track of exactly which ones are causing trouble. Currently, you need to look inside each of your dependency jars in order to figure out what's wrong - I think this change is a harmless quality-of-life improvement, and that it should apply cleanly to all three release branches.

```
Caused by: java.lang.IllegalArgumentException: More than one fragment with the name [some-popular-library] was found. 
This is not legal with relative ordering. See section 8.2.2 2c of the Servlet specification for details. 
Consider using absolute ordering. 
(Duplicates found in: [[file:/home/mads/dev/tomcat/output/build/webapps/madswar/WEB-INF/lib/popular-lib-duplicate.jar, file:/home/mads/dev/tomcat/output/build/webapps/madswar/WEB-INF/lib/popular-lib.jar]]).
```

### DIFF
--- a/java/org/apache/tomcat/util/descriptor/web/FragmentJarScannerCallback.java
+++ b/java/org/apache/tomcat/util/descriptor/web/FragmentJarScannerCallback.java
@@ -136,7 +136,7 @@ public class FragmentJarScannerCallback implements JarScannerCallback {
             // this name as having a duplicate so Tomcat can handle it
             // correctly when the fragments are being ordered.
             String duplicateName = fragment.getName();
-            fragments.get(duplicateName).setDuplicated(true);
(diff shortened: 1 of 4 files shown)


markt-asf:
Thanks for the PR. I like it. I'm probably being over cautious but I'd like to retain backwards compatibility for the existing methods on WebXml - at least up to 10.0.x anyway. Therefore, I'm going to apply this manually with a few tweaks.

markt-asf:
AS per previous comment, merged manually. Thanks very much for the PR.

PR #538, opened 2022-08-13
Fix typo

aooohan:


markt-asf:
Thanks for the PR. Back-ported to 10.0.x. (9.0.x and earlier does not have the affected code.)

PR #540, opened 2022-08-15
Fix typo

tianshuang:


markt-asf:
Thanks for the PR. Back-ported to all other supported branches.

PR #543, opened 2022-08-18
Optimize comment

tianshuang:
HTTP2 also supports relative redirects.


markt-asf:
Thanks. I'll back-port this to earlier Tomcat versions as well.

PR #544, opened 2022-08-18
Avoid errors from shutting down `LoggerThread`

ppkarwasz:
A Java `Error` in the `LoggerThread` can effectively cripple Tomcat JULI, preventing all `AsyncFileHandler` from writing to their files.

We experienced such a logging shutdown on a real Gerrit installation, when a `toString()` method threw a `NoClassDefFoundError` and stopped the logger thread used by all the applications on the server.

Should I file a BZ bug?

### DIFF
--- a/java/org/apache/juli/AsyncFileHandler.java
+++ b/java/org/apache/juli/AsyncFileHandler.java
@@ -205,6 +205,12 @@ public class AsyncFileHandler extends FileHandler {
                 } catch (InterruptedException x) {
                     // Ignore the attempt to interrupt the thread.
                 } catch (Throwable x) {
+                    if (x instanceof VirtualMachineError) {
+                        throw (VirtualMachineError) x;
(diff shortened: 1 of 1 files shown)


ppkarwasz:
@markt-asf,

As far as I can see, the `LoggerThread` is the only part of JULI that needs to be concerned with `Throwable`s. I inline the relevant part of `ExceptionUtils#handleException`. A stack overflow is not a problem in this case, since we are in a thread's `run` method.

 On the other hand I would be tempted to replace `LoggerThread` entirely with a single threaded `ExecutorService` as in [new-async-file-handler](https://github.com/ppkarwasz/tomcat/tree/new-async-file-handler). That solution will require much more testing, but will solve another potential problem: the `LoggerThread` can have as context classloader the classloader of a web application.

markt-asf:
Fixed via #545

PR #545, opened 2022-08-20
Replace `LoggerThread` with single-threaded executor

ppkarwasz:
This provides an alternative solution to the `LoggerThread` problems described in #544.

It replaces `LoggerThread` with a single-threaded `ExecutorService` almost identical to the one used by `FileHandler`. This improves `AsyncFileHandler` in two aspects:

 * the thread created by the executor service always uses Tomcat JULI's classloader as TCCL, preventing a memory leak,
 * if an `Error` kills the working thread, it is regenerated by the executor service.

A test is provided to ensure that the previous overflow policies work as documented.

### DIFF
--- a/java/org/apache/juli/AsyncFileHandler.java
+++ b/java/org/apache/juli/AsyncFileHandler.java
@@ -61,7 +61,7 @@ public class AsyncFileHandler extends FileHandler {
                                Integer.toString(DEFAULT_MAX_RECORDS)));
 
     private static final ExecutorService LOGGER_SERVICE = new LoggerExecutorService(OVERFLOW_DROP_TYPE,
-            DEFAULT_MAX_RECORDS);
+            MAX_RECORDS);
 
     private final Object closeLock = new Object();
     protected volatile boolean closed = false;
@@ -180,7 +180,6 @@ public class AsyncFileHandler extends FileHandler {
                     if (executor.getQueue().offer(r, 1000, TimeUnit.MILLISECONDS)) {
                         break;
                     }
-                    ;
                 } catch (InterruptedException e) {
                     Thread.currentThread().interrupt();
                     throw new RejectedExecutionException("Interrupted", e);
--- a/java/org/apache/juli/AsyncFileHandler.java
+++ b/java/org/apache/juli/AsyncFileHandler.java
@@ -16,10 +16,14 @@
  */
 package org.apache.juli;
 
+import java.util.concurrent.ExecutorService;
 import java.util.concurrent.LinkedBlockingDeque;
+import java.util.concurrent.RejectedExecutionException;
+import java.util.concurrent.RejectedExecutionHandler;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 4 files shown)


java/org/apache/juli/AsyncFileHandler.java
```diff
@@ -54,17 +60,12 @@ public class AsyncFileHandler extends FileHandler {
...
-    protected static final LinkedBlockingDeque<LogEntry> queue =
-            new LinkedBlockingDeque<>(MAX_RECORDS);
-
-    protected static final LoggerThread logger = new LoggerThread();
-
-    static {
-        logger.start();
-    }
+    private static final ExecutorService LOGGER_SERVICE = new LoggerExecutorService(OVERFLOW_DROP_TYPE,
+            DEFAULT_MAX_RECORDS);
```
aooohan, line 64: There should use `MAX_RECORDS`,not `DEFAULT_MAX_RECORDS`.

java/org/apache/juli/AsyncFileHandler.java
```diff
@@ -117,115 +120,82 @@ public void publish(LogRecord record) {
...
-                    LogEntry entry = queue.take();
-                    entry.flush();
-                } catch (InterruptedException x) {
-                    // Ignore the attempt to interrupt the thread.
-                } catch (Exception x) {
-                    x.printStackTrace();
+                    if (executor.getQueue().offer(r, 1000, TimeUnit.MILLISECONDS)) {
+                        break;
+                    }
+                    ;
```
aooohan, line 183: This line needs to be removed.

test/org/apache/juli/TestAsyncFileHandlerOverflow.java
```diff
@@ -0,0 +1,142 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.juli;
+
+import static org.junit.Assert.assertEquals;
```
aooohan, line 19: Please do not use static import for the assertion tool as it is not very clear.

java/org/apache/juli/AsyncFileHandler.java
ppkarwasz: Corrected, thanks.

markt-asf:
I think this is a better approach than the current code that uses a Thread. Therefore, I think we can close #544.

I have been reviewing the PR and overall I like it. I have noticed that it is possible to lose log messages during shutdown (add a 100ms delay just before the call to publish internal to simulate a large backlog). The threading approach originally suffered from this problem as well and I think the same solution will work (register uses of the AsyncHandler and explicitly shutdown the executor during JULI shutdown). I plan on working on this today.

markt-asf:
Applied manually so I could address the lost messages on shutdown issue.
Thanks for the PR.

ppkarwasz:
@markt-asf, thanks for taking care of this.

PR #548, opened 2022-08-26
Avoid int overflow when parsing octet

mayrstefan:
When parsing an octet stop when value exceeds 255. Should fix [bug 66240](https://bz.apache.org/bugzilla/show_bug.cgi?id=66240)


aooohan:
Thanks for the PR.

I applied this manually so I could:
- add some test cases
- add changelog entry

PR #552, opened 2022-09-11
UTF-8 makes Non-English Windows Console garbled

xth:
Windows console dose not use UTF-8 by default.
Unset the ConsoleHandler.encoding ,the JVM would choose the platform default.

### DIFF
--- a/conf/logging.properties
+++ b/conf/logging.properties
@@ -48,7 +48,7 @@ handlers = 1catalina.org.apache.juli.AsyncFileHandler, 2localhost.org.apache.jul
 
 java.util.logging.ConsoleHandler.level = FINE
 java.util.logging.ConsoleHandler.formatter = org.apache.juli.OneLineFormatter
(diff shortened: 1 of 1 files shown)


aooohan:
I have no objection as the ConsoleHandler shouldn't be used on production system, so this change will allow for better monitoring at development time.

kkolinko:
-1
In the majority of use cases the console output (the stdout & stderr streams) are redirected, to a file or to a process (such as being displayed in an IDE). This includes the case when Tomcat is installed as a service on Windows.
Use of UTF-8 encoding prevents loss of data.

aooohan:
> -1 In the majority of use cases the console output (the stdout & stderr streams) are redirected, to a file or to a process (such as being displayed in an IDE). This includes the case when Tomcat is installed as a service on Windows. Use of UTF-8 encoding prevents loss of data.

Thanks @kkolinko  for your reply, I did not consider it carefully enough.:stuck_out_tongue_winking_eye:

PR #555, opened 2022-10-12
Update OpenSSLEngine.java  changed 'toArray(new String[size])'  to  'toArray(new String[0])'

DigitalFatCat:
It should use 'toArray(new String[0])' instead of 'toArray(new String[size])'

### DIFF
--- a/java/org/apache/tomcat/util/net/openssl/OpenSSLEngine.java
+++ b/java/org/apache/tomcat/util/net/openssl/OpenSSLEngine.java
@@ -830,7 +830,7 @@ public final class OpenSSLEngine extends SSLEngine implements SSLUtil.ProtocolIn
         if (size == 0) {
             return new String[0];
         } else {
-            return enabled.toArray(new String[size]);
+            return enabled.toArray(new String[0]);
         }
     }
 


aooohan:
It doesn't make sense to do this.

markt-asf:
It usually does make sense. See https://www.baeldung.com/java-collection-toarray-methods
I thought we'd been through the code base and fixed all these previously. A quick search indicates that isn't the case - or we missed quite a few. I have patch locally to address all of the remaining toArray() calls that I just need to review before committing.

aooohan:
> It usually does make sense. See https://www.baeldung.com/java-collection-toarray-methods

Wow, interesting, it turns out that there is a difference in performance between these. I am ashamed of my ignorance.
> I have patch locally to address all of the remaining toArray() calls that I just need to review before committing.

Ok

aooohan:
@DigitalFatCat  Thanks for the PR.  : )

ChristopherSchultz:
If we _really_ want to improve performance, we should be using static zero-sized arrays instead of instantiating them each time. A zero-item array is immutable, and thus infinitely sharable everywhere.

ChristopherSchultz:
> > It usually does make sense. See https://www.baeldung.com/java-collection-toarray-methods
> 
> Wow, interesting, it turns out that there is a difference in performance between these. I am ashamed of my ignorance.

You should not feel bad at all. It makes absolutely zero sense that this would be faster than anything else, and yet it is measurably so.

PR #556, opened 2022-10-12
Note that maxHttpHeaderSize always allocates the max size

xordspar0:


webapps/docs/config/http.xml
```diff
@@ -525,6 +525,11 @@
       the request line, header names and header values. If not specified, this
       attribute is set to the value of the <code>maxHttpHeaderSize</code>
       attribute.</p>
+      <p>If you see "Request header is too large" errors you can increase this,
+      but be aware that Tomcat will allocate the full amount you specify for
+      every request. For example, if you specify a maxHttpRequestHeaderSize of
+      1 MB and your application handles 100 concurrent requests, you will see
+      100 MB of heap consumed by request headers.</p>
```
aooohan, line 532: It doesn't affect the understanding, but it's not quite accurate, the buffer will actually be larger than 100MB due to some additional margins being taken into account

aooohan (approved):

PR #558, opened 2022-10-20
It is time to stop using absolute redirects

eperez:


aooohan:
Why? Is there a reason or a source you can refer to?

eperez:
> Why? Is there a reason or a source you can refer to?

It has passed some years now since this design mistake was fixed: https://www.rfc-editor.org/rfc/rfc7231#section-7.1.2
Keeping the old default just creates problems.

markt-asf:
`useRelativeRedirects` is already `true` by default. However, the Servlet specification *requires* (see section 5.5) that Tomcat uses absolute redirects. Therefore, if `STRICT_SERVLET_COMPLIANCE` is `true`, absolute redirects are used.
Note that, generally, the only reason to enable `STRICT_SERVLET_COMPLIANCE` is to pass the TCK. In normal production usage I'd expect this to be `false`.

PR #559, opened 2022-10-21
fix Bug 66318 - Set multiple domain names for one address，In Windows 10 OS. AbstractProcessor.populateRequestAttributeRemoteHost aways get wrong host

snifferhu:


aooohan:
Your commit has nothing to do with your PR title.
Also your commit is redundant: https://bz.apache.org/bugzilla/show_bug.cgi?id=66233

PR #560, opened 2022-10-22
Avoid the unnecessary parsing for web xml as it equals NoDefaultWebXml.

m0ver:
If the defaultWebXml is not null and equals NoDefaultWebXml, then no need to proceed the web xml configuration.


m0ver (commented):
Correct the format of the code.

markt-asf:
This is not correct. No default web.xml does not mean no web.xml. The correct check is already in place in `getGlobalWebXmlSource()`

PR #561, opened 2022-10-23
Introduce a method to skip the web.xml files scan.

m0ver:


m0ver:
Thanks Mark! 

I'm using tomcat embed server as one of web servers in my application. Just notice that `webConfig()` needs to process a lot of things related web.xml which seems not to be required for an embed web server. so to skip this process can accelerate the start up actually. any ideas?

Do you think is it better to use `skipWebConfig()` to instead of `skipWebXmlFileScan()`?

markt-asf:
The discussion belongs on the mailing list.

PR #565, opened 2022-11-09
updated link to download cvs

Naturbelassen:
The documentation suggest to get cvs by following this link: http://www.cvshome.org/ 
However the linked page has nothing to do with csv.


aooohan:
Thanks for bringing it to my attention. But CSV is too old and outdated and I think it's time to use Git instead. So I will try to change some docs related to this.

markt-asf:
I've updated the docs to use more generic language.

PR #569, opened 2022-11-16
Refine native resources metadata

sdeleuze:
This commit updates the native resources hints in order to avoid classpath wide inclusion of resources.

@markt-asf @mhalbritter Please check I did not make any mistake.

### DIFF
--- a/modules/stuffed/tomcat-resource.json
+++ b/modules/stuffed/tomcat-resource.json
@@ -73,9 +73,13 @@
     {"name":"org.apache.tomcat.websocket.server.LocalStrings"}
   ],
   "resources":[
-    {"pattern":".*/mbeans-descriptors.xml$"},
-    {"pattern":".*/*.properties$"},
-    {"pattern":".*/*.dtd$"},
-    {"pattern":".*/*.xsd$"}
+    {"pattern":"^org/apache/tomcat/.*mbeans-descriptors\\.xml$"},
+    {"pattern":"^org/apache/catalina/.*mbeans-descriptors\\.xml$"},
+    {"pattern":"^org/apache/coyote/.*mbeans-descriptors\\.xml$"},
+    {"pattern":"^org/apache/catalina/.*\\.properties$"},
+    {"pattern":"^jakarta/servlet/resources/.*"},
+    {"pattern":"^org/apache/tomcat/.*\\.dtd$"},
+    {"pattern":"^org/apache/jasper/.*mbeans-descriptors\\.xml$"},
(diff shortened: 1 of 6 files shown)


modules/stuffed/tomcat-resource.json
```diff
@@ -73,9 +73,13 @@
     {"name":"org.apache.tomcat.websocket.server.LocalStrings"}
   ],
   "resources":[
-    {"pattern":".*/mbeans-descriptors.xml$"},
-    {"pattern":".*/*.properties$"},
-    {"pattern":".*/*.dtd$"},
-    {"pattern":".*/*.xsd$"}
```
ChristopherSchultz, line 79: Is a pattern no longer (or was never?) necessary for `.xsd`?

sdeleuze: They are now included via `{"pattern":"^jakarta/servlet/resources/.*"}`.

modules/stuffed/tomcat-resource.json
```diff
@@ -73,9 +73,13 @@
...
+    {"pattern":"^org/apache/tomcat/.*mbeans-descriptors\\.xml$"},
+    {"pattern":"^org/apache/catalina/.*mbeans-descriptors\\.xml$"},
+    {"pattern":"^org/apache/coyote/.*mbeans-descriptors\\.xml$"},
+    {"pattern":"^org/apache/catalina/.*\\.properties$"},
+    {"pattern":"^jakarta/servlet/resources/.*"},
```
mhalbritter, line 80: So we really need to include all of the XSDs and the DTDs?

sdeleuze: I am not sure, if for typical use case we could keep only the latest version for example, would be nice, but I would need Tomcat team guidelines for that cc @markt-asf.

markt-asf: They are only required if xmlValidation (or tldValidation) is enabled. I think you'll get warnings if they are missing.

sdeleuze: Interesting, let me have another look to see if this should be shipped by default or not.

sdeleuze: After a deeper look, I have chosen to remove `jakarta/servlet/resources/*` resources from the default configuration of `tomcat-embed-core`, since those optional resources weight `1.5M`! Users leveraging them will just have to configure something like `{"pattern":"^jakarta/servlet/resources/.*"}`, not that different from what people need to do for JNI or extra reflection configuration.

PR #572, opened 2022-12-01
66370-Fix for Boolean System property lookup with Security Manager enabled

isaacrivriv:
https://bz.apache.org/bugzilla/show_bug.cgi?id=66370

### DIFF
--- a/java/jakarta/el/Util.java
+++ b/java/jakarta/el/Util.java
@@ -43,8 +43,8 @@ class Util {
     private static final Class<?>[] EMPTY_CLASS_ARRAY = new Class<?>[0];
     private static final Object[] EMPTY_OBJECT_ARRAY = new Object[0];
 
-    private static final boolean GET_CLASSLOADER_USE_PRIVILEGED =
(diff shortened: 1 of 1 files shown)


lalo-mx:
See https://bz.apache.org/bugzilla/show_bug.cgi?id=66294

isaacrivriv:
Yes I saw that, my question was more as to why by default the property was made to disable the security manager. There are other areas of code where the security manager is used but there are no properties to disable them. This is why I suggest in this PR to update the property in order to keep the same behavior as the other areas of code that use the security manager if enabled by default. We could use `org.apache.el.GET_CLASSLOADER_IGNORE_PRIVILEGED` instead of `org.apache.el.GET_CLASSLOADER_USE_PRIVILEGED` to disable or enable the security manager in this case which would still have the benefits listed in that issue to not use the security manager but would need to be manually set.

ChristopherSchultz:
You should read both https://lists.apache.org/thread/66djc4j4ybz45ponly0x5g94oyt844cm and https://lists.apache.org/thread/7w4x90zwp3qhn9qopbhmg5wponcc468n in their entirety to see the thought-process(es) here.

markt-asf:
The PR will not be applied.

PR #576, opened 2023-01-24
Enable to disable request/error counters and cache stats

rmannibucau:


ChristopherSchultz:
This seems like a lot of code to avoid two addition operations.

What's the use-case, here?

rmannibucau:
@ChristopherSchultz mainly reduce the associated overhead (cpu - cas - and mem - in no luck case multiple cells are allocated per context) since most of the time these counters are useless as such - it is not rare to either not use them at all or duplicate them in a monitoring solution. In terms of code it is only ~30LoC, rest is the enablement of the counters. Agree it is not critical - why I didnt push it much more, but it is overall weird to enable by default such things and has a cost and is not easy to drop since it is in this particular valve - it would be another it could be dropped, wich can be a better option maybe?

rmannibucau:
[auto] outdated PR, guess there is no will to get it so closing as a housekeeping cycle

PR #578, opened 2023-02-01
Remove redundant modifiers for interface members

dsoumis:
I have parsed the whole project and removed redundant modifiers for interface members resulting in vanishing of ~2000 warnings.

Although it's a big commit, I strongly believe that it should be reviewed and if ok, merged, as it provides an overall more consistent project.


markt-asf:
I've applied this change via a combination of IDE tools and manual review as the CheckStyle test for redundant modifiers covers more than just interface methods.

I was initially in two minds whether to apply this but as I worked through it, I could see the benefits it offers in making the visibility - particularly of constructors - clearer. Prior to this patch, many were marked as public even though there were effectively package private.

PR #579, opened 2023-02-02
Add activateDropConnection config to choose drop connection by http status code

gksxodnd007:
when response http status code is `400, 408, 411, 413, 414, 500, 501, 503`, the connection is closed by tomcat even though that is keep-alive. undertow and jetty and netty don't close keep-alive connection unlike tomcat. i believe that it can be optional feature. 

+ our service run on kubernetes and we use istio as service mesh. so tomcat only connects to sidecar. we don't need to frequently close the connection for easily load balancing or other reason.

the first image is before remove statusDropsConnection method, and the second image is after that.
(upstream connection count means creating new connection count)
image

image

`activateDropConnection` field name can be change, if someone give more good thing. thank you. (--)(__)

### DIFF
--- a/java/org/apache/coyote/http11/AbstractHttp11Protocol.java
+++ b/java/org/apache/coyote/http11/AbstractHttp11Protocol.java
@@ -661,6 +661,20 @@ public abstract class AbstractHttp11Protocol<S> extends AbstractProtocol<S> {
         getEndpoint().setMaxKeepAliveRequests(mkar);
     }
 
+    public boolean getActivateDropConnection() {
+        return getEndpoint().getActivateDropConnection();
+    }
+    /**
+     * If drop connection activate,
+     * then connection doesn't close even though response http status code is abnormal.
+     * The default is false
+     *
+     * @see Http11Processor#statusDropsConnection(int)
+     */
+    public void setActivateDropConnection(boolean activateDropConnection) {
+        getEndpoint().setActivateDropConnection(activateDropConnection);
+    }
+
 
     // ----------------------------------------------- HTTPS specific properties
     // ------------------------------------------ passed through to the EndPoint
--- a/java/org/apache/coyote/http11/Http11Processor.java
+++ b/java/org/apache/coyote/http11/Http11Processor.java
@@ -195,7 +195,11 @@ public class Http11Processor extends AbstractProcessor {
      * Determine if we must drop the connection because of the HTTP status
      * code.  Use the same list of codes as Apache/httpd.
      */
-    private static boolean statusDropsConnection(int status) {
+    private boolean statusDropsConnection(int status) {
+        if (!protocol.getActivateDropConnection()) {
+            return false;
+        }
+
         return status == 400 /* SC_BAD_REQUEST */ ||
                status == 408 /* SC_REQUEST_TIMEOUT */ ||
                status == 411 /* SC_LENGTH_REQUIRED */ ||
--- a/java/org/apache/tomcat/util/net/AbstractEndpoint.java
+++ b/java/org/apache/tomcat/util/net/AbstractEndpoint.java
@@ -938,6 +938,17 @@ public abstract class AbstractEndpoint<S,U> {
         this.maxKeepAliveRequests = maxKeepAliveRequests;
     }
 
+    /**
+     * activate drop connection when response http status code is abnormal.
+     */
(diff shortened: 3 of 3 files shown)


rmaucher:
Ok why not, but "activateDropConnection" does not sound very good.

gksxodnd007:
@markt-asf 
thank you for your good opinion. but I think, it is no need to force drop connections. also, the default action is not changing by setting false the value. and some situations will be safe like just connecting to the only sidecar because the risk is preserved by the sidecar or other proxy server.

dongseokk:
> -1. veto. I have security concerns. The connection is dropped in error conditions to avoid request smuggling risks. Of the status codes lists, the only one we might be able to handle safely is 503 and that would require changes to ensure that the current request was fully swallowed before processing the next request.

The security concerns make sense, but I think users may want to opt out.
The concerns would be mitigated fairly if the default behavior is dropping.
Users will explicitly disable it only when they know what they are doing.

markt-asf:
- 400 Always unsafe to continue since if you have a badly formatted request you can't tell where that request ends and the next one starts.
- 408 Always unsafe to continue since the current request was not fully read
- 411 Likely unsafe since the only reason for asking for a length is if you can't work it out. However, Tomcat never uses this status code so the point is moot.
- 413 Unsafe since it is likely that the request will not be read fully
- 414 Unsafe since it is likely that the request will not be read fully
- 500 Unsafe since the request processing is in an unknown state
- 503 Safe if, and only if, the request is fully read
- 501 Safe if, and only if, the request is fully read

If the change was limited to 501 and 503 and if it is confirmed that Tomcat reads the request fully in those cases (or there are changes to ensure that this is the case) then I'd be willing to reconsider my veto.

I'd suggest, at this point, the OP's time would be better spent determining which status codes are triggering the observed connection closures and addressing the associated root cause(s).

hojongs:
@markt-asf 
I get what you mean. But some people (including me) usually use the status codes with additional cases. For example, the status code 400 doesn't occur only in the badly formatted request. When requests are formatted well but the request value is incorrect, I use the status code 400 to respond to the request. (because I think the status code 400 is the most suitable in this case) and another use case: Sometimes our service can respond to clients with status code 500 during connections because of some reasons.

In these cases, I don't want decreased performance by dropping connections due to responses of the status code 500 even though I need to endure the security concerns you mentioned. So I want this option to choose control over the default behavior.

I think it'll be fine if users can have the opportunity to choose the behavior while the default behavior is dropping connection for the security concern you mentioned. It'll be safer for servers to drop connections as the default behavior.

gksxodnd007:
@markt-asf 
I see the reactor netty code to check how they close the connection when the current request was not fully read.
they also close the connection even though that is keep-alive, when the decode processing of the request failed.
how about adding defensive code to prevent the request smuggling risks? if I add the code, would you reconsider about dropping the connection by HTTP status code?

ref:
- https://github.com/reactor/reactor-netty/blob/c71ebe4372f35496eb04471355ea84739bc6381a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpTrafficHandler.java#L193-L197
- https://github.com/reactor/reactor-netty/blob/c71ebe4372f35496eb04471355ea84739bc6381a/reactor-netty-http/src/main/java/reactor/netty/http/server/HttpTrafficHandler.java#L280-L283

rmaucher:
Of course, some instances of these status codes would be "ok" to recover from, but Mark is actually right: it is not going to be deterministic.
Also it is not possible to add "defensive" code since starting to decode random bytes in a random location will allow request smuggling (a proxy and Tomcat have to both "see" the same HTTP requests coming in and process them the same way). If you are using Tomcat without proxying of any kind (that happens very often, right ?), then it is less dangerous although there is some DoS potential on HTTP header parsing errors.

gksxodnd007:
@rmaucher 
thank you for your explanation :) I understand. is there any better way that reuses the connection for utilizing computing resources?

markt-asf:
My suggestion at this point would be to use custom 4xx and 5xx status codes in the application rather than re-using known codes.

gksxodnd007:
@markt-asf 
I'm sorry, but I'll follow your opinion. and I learn about the request smuggling risks while communication with you. thank you very much :)

markt-asf:
No need for an apology. This has been a useful discussion.

PR #581, opened 2023-02-02
Speedup by removing non pattern replaceAll with constant arg

tbw777:
replaceAll("ABC", "") is non Pattern method and therefore must be replaced to simple fast replace()
A proofs of changes: https://gist.github.com/tbw777/8a6ef60af21487c5faec67037099fd0b

### DIFF
--- a/java/org/apache/tomcat/buildutil/translate/Utils.java
+++ b/java/org/apache/tomcat/buildutil/translate/Utils.java
@@ -123,7 +123,7 @@ public class Utils {
 
         result = ESCAPE_LEADING_SPACE.matcher(result).replaceAll("\\\\$1");
 
-        result = result.replaceAll("\t", "\\t");
+        result = result.replace("\t", "\\t");
 
         return result;
     }
--- a/java/org/apache/catalina/startup/ClassLoaderFactory.java
+++ b/java/org/apache/catalina/startup/ClassLoaderFactory.java
@@ -287,7 +287,7 @@ public final class ClassLoaderFactory {
         // JARs. If these URLs are used to construct URLs for resources in a JAR
         // the URL will be used as is. It is therefore necessary to ensure that
         // the sequence "!/" is not present in a class loader URL.
-        String result = urlString.replaceAll("!/", "%21/");
+        String result = urlString.replace("!/", "%21/");
         return new URI(result).toURL();
     }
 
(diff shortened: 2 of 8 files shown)


java/org/apache/tomcat/buildutil/translate/Utils.java
```diff
@@ -123,7 +123,7 @@ static String formatValueCommon(String in) {
 
         result = ESCAPE_LEADING_SPACE.matcher(result).replaceAll("\\\\$1");
 
-        result = result.replaceAll("\t", "\\t");
+        result = result.replace("\t", "t");
```
markt-asf, line 127: This is not correct. The literal (unescaped) replacement text should be `\t`

markt-asf (changes requested):
Thanks for this PR - these are good changes to have.

java/org/apache/tomcat/buildutil/translate/Utils.java
tbw777: The second argument in replaceAll means escape t, not \t.
You can check proofs tests. equals returning true if replace maked to t, not \t

markt-asf: It is a bug in the original code. It should be replacing all tabs with the literal string `\t`.

markt-asf: For the record, none of the current i18n strings use tabs. I did find some messages using them in early 7.0.x versions, but those messages were never referenced in the Java code.

tbw777:
@markt-asf 
Branch was updated and https://gist.github.com/tbw777/8a6ef60af21487c5faec67037099fd0b also
Check please.

java/org/apache/tomcat/buildutil/translate/Utils.java
markt-asf: Tx.

markt-asf (approved):

ChristopherSchultz:
There isn't much of a difference between `String.replace` and `String.replaceAll`. They both use regular expressions under the hood. Java's `Pattern` class has a special case for "this search-string doesn't include any metacharacters" (it's a flag called `LITERAL`) and has optimized code-paths for those cases.

IMO `replaceAll` is more clear than `replace` because it's absolutely clear that all instances will be replaced in the string.

If performance is the primary factor, here, we should use an implementation of string-replacement that does not involve creating `Pattern` and `Matcher` objects and re-encoding the replacement String _every single time_.

ChristopherSchultz:
I think this "performance improvement" is not an improvement and doesn't affect performance in any meaningful way.

ChristopherSchultz:
Apologies: my comments are rubbish after Java 9. Please forgive the noise.

happyhua:
A replaceAll -> replace change in for example `java/org/apache/tomcat/buildutil/translate/Utils.java` can produce different output.


```
        String source = "1\t2\t3";

        System.out.println("source: " + source);
        System.out.println("replaceAll: " + source.replaceAll("\t", "\\t"));
        System.out.println("replace: " + source.replace("\t", "\\t"));

```

```
source: 1	2	3
replaceAll: 1t2t3
replace: 1\t2\t3
```

PR #582, opened 2023-02-06
Enabled headless build configuration

tbw777:


markt-asf:
What problem is this PR trying to solve?

tbw777:
no problem, just selecting less jvm api to work

ChristopherSchultz:
If no code requests anything AWT-related, then the headless config of the JVM makes no difference. I see no improvement here at all, just added lines to an already complex build script.

tbw777:
Also this is guarantee that no awt components used

ChristopherSchultz:
It does not guarantee that no AWT components are used. It only guarantees that anything which requires an actual screen will fail (such as trying to open a Frame, etc.). AWT is still perfectly usable when "headless". In fact, that's the whole point of putting the JVM into "headless" mode: to allow AWT to be used even when there isn't a "head" (screen).

markt-asf:
I don't see any benefit to adding these extra settings.

PR #583, opened 2023-02-06
Replaced old time API

tbw777:
SimpleDateFormatter isnt thread safe and deprecated with new jvm8 time API


java/org/apache/catalina/util/Strftime.java
```diff
@@ -113,36 +113,52 @@ public class Strftime {
...
     /**
      * Format the date according to the strftime-style string given in the constructor.
      *
-     * @param date the date to format
+     * @param instant the date to format
      * @return the formatted date
      */
-    public String format( Date date ) {
-        return simpleDateFormat.format( date );
+    public String format(Instant instant) {
```
markt-asf, line 125: Changes to public APIs need more work. Generally, the old API is deprecated in the latest stable release (and possible older releases) and the new API is added to at least the current development release. The new API may be back-ported to older releases where it makes sense to do so.

PR #586, opened 2023-02-10
Fixed closing serverSocket in doCloseServerSocket()

tbw777:


markt-asf:
Thanks for the report. Fixed slightly differently (effectively the same code) to align with the existing NIO2 code.

PR #587, opened 2023-02-10
Removed unused assignments

tbw777:


aooohan:
-1
I don't think this change makes sense to do that, the current code is a bit more intuitive.
As a reminder, when you propose a PR, we hope you will add some description to avoid wasting your time and ours.

PR #588, opened 2023-02-10
Removed object creation

tbw777:


ChristopherSchultz (commented):
Seems like `setLength(1)` would be a better choice for `case ' '`.

PR #591, opened 2023-02-16
Implement more special maps for rewrite valve

dsoumis:
Implemented more special maps from https://httpd.apache.org/docs/2.4/rewrite/rewritemap.html as implied by the FIXME tag.

Added relevant tests and added ServletMapping to test int: MapType.


rmaucher:
I merged it with a lot of changes and cleanups. Thanks.

PR #592, opened 2023-02-17
Improved regexp performance: "a-zA-Z0-9_" -> "\w"

tbw777:
https://gist.github.com/tbw777/8ce56d2cc3a0216012362d18b7262eb2

### DIFF
--- a/conf/web.xml
+++ b/conf/web.xml
@@ -359,7 +359,7 @@
   <!--                        the arguments to the OS. See the CGI How-To   -->
   <!--                        for more details. The default varies by       -->
   <!--                        platform.                                     -->
-  <!--                        Windows: [[a-zA-Z0-9\Q-_.\\/:\E]+]            -->
+  <!--                        Windows: [[\w\Q-.\\/:\E]+]            -->
   <!--                        Others:  [.*]                                 -->
   <!--                        Note that internally the CGI Servlet treats   -->
   <!--                        [.*] as a special case to improve performance -->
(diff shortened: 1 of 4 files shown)


java/jakarta/servlet/jsp/resources/jspxml.xsd
```diff
@@ -25,7 +25,7 @@
 <!ENTITY SetProp    "(&Identifier;|\*)">
 <!ENTITY RelativeURL  "[^:#/\?]*(:{0,0}|[#/\?].*)">
 <!ENTITY Length     "[0-9]*&#x25;?">
-<!ENTITY AsciiName    "[A-Za-z0-9_-]*">
+<!ENTITY AsciiName    "[\w-]*">
```
markt-asf, line 28: The Tomcat project is unable to change these schema. We have to use the ones provided by the specification. Please direct this part of the PR to the Jakarta Schema project:
https://github.com/eclipse-ee4j/jakartaee-schemas

markt-asf (changes requested):
Thanks for the performance data. That is really useful.
Once the PR is updated to remove the schema changes, I think this will be ready to merge.

ChristopherSchultz:
I'm curious about performance data when the `\w` has other things added to it, which is the case for all examples in Tomcat. This microbenchmark only compares `[A-Za-z0-9_]` as the whole character class against `\w` and `[\w]` without any other characters added to the character class.

Also, most expressions used in Tomcat have a trailing `+` which in the attached performance data show that performance is again terrible. It still appears measurably and consistently better than `[A-Za-z0-9_]` but I think it's important to benchmark what Tomcat _actually_ uses and not something _very close_ to what Tomcat uses.

tbw777:
https://gist.github.com/tbw777/cd394ec67a01f9e7e8fe4d0c66d74637

ChristopherSchultz:
Thanks @tbw777 for the updated micro-benchmarks. I agree that the performance improvement is significant enough to warrant a change to Tomcat.

tbw777:
Compile speed
https://gist.github.com/tbw777/6a6303f64894e65a160d3f6321685d27

aooohan:
Merge manually, thanks.

PR #593, opened 2023-02-21
Update CONTRIBUTING.md

zengwei2000:
fix typo


markt-asf:
That spelling is correct for US English.
Given the range of contributors to Tomcat, there is a mix of UK and US English used in the Tomcat docs.

PR #594, opened 2023-03-03
Update CoyoteAdapter.java

pangxianhai:
we have a problem at function parseSessionCookiesId. If the request has two JSESSIONID on the Cookie. The session will init twice. So request.isRequestedSessionIdValid() will  be invoked. But at request.isRequestedSessionIdValid() , it will call manager.findSession(requestedSessionId). This time request is not bind ClassLoader. We customized the manager. On the manager, will used some class at War. So it's will throw a ClassNotFound exception. So I think the session only need init once. Because the second init is the same as the first time. If request.isRequestedSessionIdValid return true,the second is not need.

### DIFF
--- a/java/org/apache/catalina/connector/CoyoteAdapter.java
+++ b/java/org/apache/catalina/connector/CoyoteAdapter.java
@@ -1021,6 +1021,7 @@ public class CoyoteAdapter implements Adapter {
                     if (log.isDebugEnabled()) {
                         log.debug(" Requested cookie session id is " + request.getRequestedSessionId());
                     }
+                    break;
                 } else {
                     if (!request.isRequestedSessionIdValid()) {
                         // Replace the session id until one is valid


pangxianhai:
we have a problem at function parseSessionCookiesId. If the request has two JSESSIONID on the Cookie. The session will init twice. So request.isRequestedSessionIdValid() will be invoked. But at request.isRequestedSessionIdValid() , it will call manager.findSession(requestedSessionId). This time request is not bind ClassLoader. We customized the manager. On the manager, will used some class at War. So it's will throw a ClassNotFound exception. So I think the session only need init once. Because the second init is the same as the first time. If request.isRequestedSessionIdValid return true,the second is not need.

markt-asf:
Thanks for the PR. I understand the problem but this isn't the right fix. It will break apps when there are two session cookies and the second one is the only valid one.

markt-asf:
I've applied what I believe to be the 'right' fix for this. We now use the web application class loader when looking up the session and checking the validity.

PR #595, opened 2023-03-07
Exception message error

Window-Yan:


aooohan:
Thanks for reporting this. However, I found differences between this file and spec, so I will update it to align with spec.

PR #598, opened 2023-03-13
Use available constants

isapir:
We seem to have some duplication, e.g. `o.a.c.Globals.CATALINA_HOME_PROP` which is set from `o.a.c.startup.Constants.CATALINA_HOME_PROP`.

Do we need `o.a.c.startup.Constants` to be separate for some reason, e.g. for packaging different JARs, or can we consolidate the code and use just `o.a.c.Globals` instead?

The current PR does not modify that.  It only replaces the literal strings `"catalina.home"` and `"catalina.base"` with constants from `Globals`, or in the `o.a.c.startup` package with `Constants`.

I'm opening the PR so there are more eyes on it before I make the change, and will merge it after some reasonable time passes if no one raises any issues with it.

### DIFF
--- a/java/org/apache/catalina/storeconfig/ConnectorStoreAppender.java
+++ b/java/org/apache/catalina/storeconfig/ConnectorStoreAppender.java
@@ -29,6 +29,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
+import org.apache.catalina.Globals;
 import org.apache.catalina.connector.Connector;
 import org.apache.coyote.ProtocolHandler;
 import org.apache.tomcat.util.IntrospectionUtils;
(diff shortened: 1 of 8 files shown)


java/org/apache/catalina/storeconfig/ConnectorStoreAppender.java
```diff
@@ -29,6 +29,7 @@
 import java.util.List;
 import java.util.Set;
 
+import org.apache.catalina.Globals;
```
isapir, line 32: I see that we also have a file named `java/org/apache/catalina/storeconfig/Constants.java` but it does not set `CATALINA_BASE_PROP`.  I am guessing there is a reasoning behind the different `Constants` classes so trying to figure it out.

rmaucher:
The thing to look for is to not add any dependencies to the bootstrap, but it looks ok here (I think).

isapir:
Thanks @rmaucher!  I will watch out for that.

PR #599, opened 2023-03-15
Use available constants

isapir:
Created a branch and PR because my local git shows some new class files and I'm not sure where they came from so doing it this way so that I can safely review the patch before merging it


rmaucher:
Personally, I hate these "work" branches. Please avoid them, and at the very least delete them afterwards. Thanks.

isapir:
@rmaucher Sounds good.  I added the branch because my local git repo had a bunch of class files and I wanted to make sure that I only push the changes that I meant to push.  I will do better about deleting such branches after merging them in the future.  Thank you.

kingjay66:
Thanks for the update

PR #600, opened 2023-03-16
echo the value of %JAVA_HOME% when it is invalid.

VirtualTim:
This is useful for diagnosing %JAVA_HOME% issues.

### DIFF
--- a/bin/service.bat
+++ b/bin/service.bat
@@ -118,7 +118,7 @@ if not "%JRE_HOME%" == "" goto okJavaHome
 set "JRE_HOME=%JAVA_HOME%"
 goto okJavaHome
 :noJavaHome
-echo The JAVA_HOME environment variable is not defined correctly
+echo The JAVA_HOME environment variable "%JAVA_HOME%" is not defined correctly
 echo This environment variable is needed to run this program
 echo NB: JAVA_HOME should point to a JDK not a JRE
 exit /b 1


ChristopherSchultz (changes requested):
If JAVA_HOME is not defined at all, then this yields no visible change.

I think it's important to make it clear if JAVA_HOME is not defined and that's why we couldn't find the JVM. Maybe put it on a separate line like:

```
echo  The JAVA_HOME environment variable is not defined correctly.
echo JAVA_HOME="%JAVA_HOME%"
```

Also, would you care to make a similar change to the other scripts that check for JAVA_HOME -- and for both .bat an .sh files?

PasunuriSrinidhi:
if not "%JAVA_HOME%" == "" (
  set "JAVA="%JAVA_HOME%\bin\java.exe""
) else (
  echo The JAVA_HOME environment variable is not defined correctly.
  echo JAVA_HOME="%JAVA_HOME%"
)
This way, if JAVA_HOME is not defined, the user will see a clear message stating that the variable is not defined correctly, and the value of the variable (which will be empty in this case) will be echoed back for reference.

markt-asf:
If neither `JRE_HOME` nor `JAVA_HOME` is defined, a message is already shown.

I have undertaken a wider review of setclasspath.bat and service.bat and added additional messages when values aren't valid and aligned the handling of `JRE_HOME` and `JAVA_HOME` for both scripts. There is no functional change but there are additional messages and the code is consistent.

VirtualTim:
Oh cool. I was planning on getting around to it, but didn't get time.

markt-asf:
np - thanks for the prompt to improve things.

PR #601, opened 2023-03-18
8.5.x

GHMOBILEMARCH:
Hello, Developer:

I'd like to know if a header for client requests is useful, or how to enable a header that supports client requests. eg: accept
eg: The client sets accept: image/jpeg and requests a txt file, but the server replies HTTP/1.1 200 Accept-Ranges: bytes;
Why didn't the server reply HTTP/1.1 406 instead of HTTP/1.1 200

thank you


VirtualTim:
Yeah PR's aren't the place to ask questions.
I believe that the only place for this is the tomcat mailing list - https://tomcat.apache.org/lists.html#tomcat-users.

PR #604, opened 2023-03-25
Fix Bug 66543 -  Wrong log message used in StandardContext#fireRequestDestroyEvent

FSchumacher:
Give it its own log message instead of re-using the one from fireRequestInitEvent.

### DIFF
--- a/webapps/docs/changelog.xml
+++ b/webapps/docs/changelog.xml
@@ -174,6 +174,10 @@
         with some OSGi custom URL schemes that can trigger potentially slow DNS
         lookups in some configurations. (markt)
       </fix>
+      <fix>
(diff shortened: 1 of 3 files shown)


webapps/docs/changelog.xml
```diff
@@ -174,6 +174,10 @@
         with some OSGi custom URL schemes that can trigger potentially slow DNS
         lookups in some configurations. (markt)
       </fix>
+      <fix>
+        <bug>66543</bug>: Give `StandardContext#fireRequestDestroyEvent` its own
+        log message. (fschumacher)
```
markt-asf, line 179: That formatting won't work here. You'll need to use `<code>...</code>`

markt-asf (approved):

aooohan (approved):

PR #605, opened 2023-03-28
Further fix for BZ 66508

markt-asf:
https://bz.apache.org/bugzilla/show_bug.cgi?id=66508

Avoid deadlock for close messages when
WsRemoteEndpointImplServer.endMessage() for a previous message is processed on a container thread

The fix for this involves the thread currently processing a socket effectively yielding the socket lock and processor to another thread. I think that is worthy of closer review before merge.

There are a couple of potential alternatives.

Plan B would be to make handing of the close message async. That is potentially better but will be more complex to implement and may uncover other complexities.

Plan C would be to require useAsyncIO="true" to enable WebSocket.


markt-asf:
Sorry. Branch in wrong repo. I'll resubmit.

PR #606, opened 2023-03-28
Further fix for BZ 66508

markt-asf:
https://bz.apache.org/bugzilla/show_bug.cgi?id=66508

Avoid deadlock for close messages when
WsRemoteEndpointImplServer.endMessage() for a previous message is processed on a container thread

The fix for this involves the thread currently processing a socket effectively yielding the socket lock and processor to another thread. I think that is worthy of closer review before merge.

There are a couple of potential alternatives.

Plan B would be to make handing of the close message async. That is potentially better but will be more complex to implement and may uncover other complexities.

Plan C would be to require useAsyncIO="true" to enable WebSocket.

### DIFF
--- a/java/org/apache/tomcat/websocket/WsRemoteEndpointImplBase.java
+++ b/java/org/apache/tomcat/websocket/WsRemoteEndpointImplBase.java
@@ -66,7 +66,7 @@ public abstract class WsRemoteEndpointImplBase implements RemoteEndpoint {
     private final IntermediateMessageHandler intermediateMessageHandler = new IntermediateMessageHandler(this);
 
     private Transformation transformation = null;
-    private final Semaphore messagePartInProgress = new Semaphore(1);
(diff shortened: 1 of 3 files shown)


rmaucher:
Adding workarounds for this should be good (I suppose you have the means to test it on hand ;) ). I like the original goal of websockets to be built on the Servlet API, but there are some known difficulties for this.

Of course, using async IO should always be recommended now that it works, since it is going to be more efficient in the "I'm doing real IO" scenario (more lightweight and straightforward).

markt-asf:
I have a test in that I can manually trigger the issue with a debugger at the moment and can step through the code with the fix to confirm it works. My desire is to create a unit test the recreates the issue. I have an idea for that but haven't implemented it yet.

markt-asf:
Updated and applied directly.

PR #607, opened 2023-03-29
Added RateLimitFilter

isapir:
This patch adds a RateLimitFilter based on the discussion in the Dev mailing list at [1] with more features to be added later

[1] https://lists.apache.org/thread/0gt1kyjs86g9oqxofdgm0zbrb14lzgj6

### DIFF
--- a/java/org/apache/catalina/filters/RateLimitFilter.java
+++ b/java/org/apache/catalina/filters/RateLimitFilter.java
@@ -1,0 +1,230 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.catalina.filters;
+
+import jakarta.servlet.FilterChain;
+import jakarta.servlet.FilterConfig;
+import jakarta.servlet.GenericFilter;
+import jakarta.servlet.ServletException;
+import jakarta.servlet.ServletRequest;
+import jakarta.servlet.ServletResponse;
+import jakarta.servlet.http.HttpServletResponse;
+import org.apache.catalina.util.TimeBucketCounter;
+import org.apache.juli.logging.Log;
+import org.apache.juli.logging.LogFactory;
+import org.apache.tomcat.util.res.StringManager;
+
+import java.io.IOException;
+
+public class RateLimitFilter extends GenericFilter {
+
+    /**
+     * default duration in seconds
+     */
+    public static final int DEFAULT_BUCKET_DURATION = 60;
+
+    /**
+     * default number of requests per duration
+     */
+    public static final int DEFAULT_BUCKET_REQUESTS = 300;
+
+    /**
+     * default value for enforce
+     */
+    public static final boolean DEFAULT_ENFORCE = true;
+
+    /**
+     * default status code to return if requests per duration exceeded
+     */
+    public static final int DEFAULT_STATUS_CODE = 429;
+
+    /**
+     * default status message to return if requests per duration exceeded
+     */
+    public static final String DEFAULT_STATUS_MESSAGE = "Too many requests";
+
+    /**
+     * request attribute that will contain the number of requests per duration
+     */
+    public static final String RATE_LIMIT_ATTRIBUTE_COUNT = "org.apache.catalina.filters.RateLimitFilter.Count";
+
+    /**
+     * filter init-param to set the bucket duration in seconds
+     */
+    public static final String PARAM_BUCKET_DURATION = "ratelimit.bucket.duration";
+
+    /**
+     * filter init-param to set the bucket number of requests
+     */
+    public static final String PARAM_BUCKET_REQUESTS = "ratelimit.bucket.requests";
+
+    /**
+     * filter init-param to set the enforce flag
+     */
+    public static final String PARAM_ENFORCE = "ratelimit.enforce";
+
+    /**
+     * filter init-param to set a custom status code if requests per duration exceeded
+     */
+    public static final String PARAM_STATUS_CODE = "ratelimit.status.code";
+
+    /**
+     * filter init-param to set a custom status message if requests per duration exceeded
+     */
+    public static final String PARAM_STATUS_MESSAGE = "ratelimit.status.message";
+
+    TimeBucketCounter bucketCounter;
+
+    private int actualRequests;
+
+    private int bucketRequests = DEFAULT_BUCKET_REQUESTS;
+
+    private int bucketDuration = DEFAULT_BUCKET_DURATION;
+
+    private boolean enforce = DEFAULT_ENFORCE;
+    private int statusCode = DEFAULT_STATUS_CODE;
+
+    private String statusMessage = DEFAULT_STATUS_MESSAGE;
+
+    private transient Log log = LogFactory.getLog(RateLimitFilter.class);
+
+    private static final StringManager sm = StringManager.getManager(RateLimitFilter.class);
+
+    /**
+     * @return the actual maximum allowed requests per time bucket
+     */
+    public int getActualRequests() {
+        return actualRequests;
+    }
+
+    /**
+     * @return the actual duration of a time bucket in milliseconds
+     */
+    public int getActualDurationInSeconds() {
+        return bucketCounter.getActualDuration() / 1000;
+    }
+
+    @Override
+    public void init() throws ServletException {
+
+        FilterConfig config = getFilterConfig();
+
+        String param;
+        param = config.getInitParameter(PARAM_BUCKET_DURATION);
+        if (param != null)
+            bucketDuration = Integer.parseInt(param);
+
+        param = config.getInitParameter(PARAM_BUCKET_REQUESTS);
+        if (param != null)
+            bucketRequests = Integer.parseInt(param);
+
+        param = config.getInitParameter(PARAM_ENFORCE);
+        if (param != null)
+            enforce = Boolean.parseBoolean(param);
+
+        param = config.getInitParameter(PARAM_STATUS_CODE);
+        if (param != null)
+            statusCode = Integer.parseInt(param);
+
+        param = config.getInitParameter(PARAM_STATUS_MESSAGE);
+        if (param != null)
+            statusMessage = param;
+
+        bucketCounter = new TimeBucketCounter(bucketDuration);
+
+        actualRequests = (int) Math.round(bucketCounter.getRatio() * bucketRequests);
+
+        log.info(sm.getString("rateLimitFilter.initialized",
+            super.getFilterName(), bucketRequests, bucketDuration, getActualRequests(),
+            getActualDurationInSeconds(), (!enforce ? "Not " : "") + "enforcing")
+        );
+    }
+
+    /**
+     * The <code>doFilter</code> method of the Filter is called by the container
+     * each time a request/response pair is passed through the chain due to a
+     * client request for a resource at the end of the chain. The FilterChain
+     * passed in to this method allows the Filter to pass on the request and
+     * response to the next entity in the chain.
+     * <p>
+     * A typical implementation of this method would follow the following
+     * pattern:- <br>
+     * 1. Examine the request<br>
+     * 2. Optionally wrap the request object with a custom implementation to
+     * filter content or headers for input filtering <br>
+     * 3. Optionally wrap the response object with a custom implementation to
+     * filter content or headers for output filtering <br>
+     * 4. a) <strong>Either</strong> invoke the next entity in the chain using
+     * the FilterChain object (<code>chain.doFilter()</code>), <br>
+     * 4. b) <strong>or</strong> not pass on the request/response pair to the
+     * next entity in the filter chain to block the request processing<br>
+     * 5. Directly set headers on the response after invocation of the next
+     * entity in the filter chain.
+     *
+     * @param request  The request to process
+     * @param response The response associated with the request
+     * @param chain    Provides access to the next filter in the chain for this
+     *                 filter to pass the request and response to for further
+     *                 processing
+     * @throws IOException      if an I/O error occurs during this filter's
+     *                          processing of the request
+     * @throws ServletException if the processing fails for any other reason
+     */
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+                            throws IOException, ServletException {
+
+        String ipAddr = request.getRemoteAddr();
+        int reqCount = bucketCounter.increment(ipAddr);
+
+        request.setAttribute(RATE_LIMIT_ATTRIBUTE_COUNT, reqCount);
+
+        if (enforce && (reqCount > actualRequests)) {
+
+            ((HttpServletResponse) response).sendError(statusCode, statusMessage);
+            log.warn(sm.getString("rateLimitFilter.maxRequestsExceeded",
+                super.getFilterName(), reqCount, ipAddr, getActualRequests(), getActualDurationInSeconds())
+            );
+
+            return;
+        }
+
+        chain.doFilter(request, response);
+    }
+
+    /**
+     * Called by the web container to indicate to a filter that it is being
+     * taken out of service. This method is only called once all threads within
+     * the filter's doFilter method have exited or after a timeout period has
+     * passed. After the web container calls this method, it will not call the
+     * doFilter method again on this instance of the filter. <br>
+     * <br>
+     * <p>
+     * This method gives the filter an opportunity to clean up any resources
+     * that are being held (for example, memory, file handles, threads) and make
+     * sure that any persistent state is synchronized with the filter's current
+     * state in memory.
+     * <p>
+     * The default implementation is a NO-OP.
+     */
+    @Override
+    public void destroy() {
+        this.bucketCounter.destroy();
+        super.destroy();
+    }
+}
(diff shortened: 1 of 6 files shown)


java/org/apache/catalina/filters/RateLimitFilter.java
```diff
@@ -0,0 +1,230 @@
...
+     *                 processing
+     * @throws IOException      if an I/O error occurs during this filter's
+     *                          processing of the request
+     * @throws ServletException if the processing fails for any other reason
+     */
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+                            throws IOException, ServletException {
+
+        String ipAddr = request.getRemoteAddr();
```
aooohan, line 192: For tomcat, which behind a proxy (e.g. nginx), this filter is meaningless. I think we can write common method (check whether `x-forwarded-for`  header exists) that handle that case or allow user to specific a way that how to get really IP (specific header or others).

java/org/apache/catalina/util/TimeBucketCounter.java
```diff
@@ -0,0 +1,217 @@
...
+            while (isRunning) {
+                String currentBucketPrefix = String.valueOf(getCurrentBucketPrefix());
+                ConcurrentHashMap.KeySetView<String, AtomicInteger> keys = map.keySet();
+
+                for (String k : keys) {
+                    if (!k.startsWith(currentBucketPrefix)) {
+                        // the key is obsolete, remove it
+                        keys.remove((k));
+                    }
+                }
```
aooohan, line 209: Here can be replaced with `keys.removeIf(k -> !k.startsWith(currentBucketPrefix));`, which is more streamlined.

java/org/apache/catalina/filters/RateLimitFilter.java
funkman: Somewhat agree. I don't recall seeing a Connector attribute (or filter) that will provide the client remoteAddr. Which can also come i handy for CDN's which send the client IP via header too. It feels like that might need to be a new feature.

If x-forwarded-for is used, we need to be wary and document its ability to be insecure too. Such as bad actors sending this header randomized. In which case, documenting that we need to document the assumption that the header is value is trusted.

So by itself `String ipAddr = request.getRemoteAddr();` (I think) is OK ... if a way to implement request.getRemoteAddr() can be adjusted accordingly. 

The alternative is an option called ipHeader (horrible name, just a suggestion) which is the name of the header to use to get client IP address.

rmaucher: Personally I was seeing this as a niche feature for Tomcat standalone. If there's a proxy, then it should be that component doing any rate limiting. Or did I miss something ?

funkman: If expectation is standalone, then its just a documentation update to clarifying limitations. That should (optimistically)  eliminate security practitioners claiming new CVE's

isapir: Tomcat already ships with an excellent RemoteIpFilter / RemoteIpValve and this filter does not attempt to replace them, but to work with them. The Filter documentation comes with a **WARNING** [1]:

> if Tomcat is behind a reverse proxy then you must
    make sure that the Rate Limit Filter sees the client IP address, so if for
    example you are using the Remote IP Filter, 
    then the filter mapping for the Rate Limit Filter must come <em>after</em>
    the mapping of the Remote IP Filter to ensure that each request has its IP
    address resolved before the Rate Limit Filter is applied

Did I get that wrong or am I correct that by having the RemoteIpFilter earlier in the Filter Chain then subsequent filters will see the "translated" address when calling `request.getRemoteAddr()`?  cc @rmaucher @ChristopherSchultz @markt-asf 

This Filter can work either in a standalone configuration, or behind a reverse proxy as long as the Remote Address is resolved using the methods mentioned above.  Even with reverse proxy it is useful to have this filter in place as it adds:

1. Configuration via web.xml which might be more convenient for those who are more familiar with Servlet containers than with web servers
2. A lenient mode in a way of `enforce=false`, allowing the user to configure the filter to only add a Request Attribute with the request count.  That allows the application to inspect the value and make a more intelligent decision since it knows more than the web server or the filter. 

[1] https://github.com/apache/tomcat/pull/607/files#diff-e79a16176f7d3d059a2bf4c43d5f50e9c107e27a2511379056ceea48b578c3e5R1001

aooohan: >  The Filter documentation comes with a WARNING [1]:

Sorry, I didn't look the doc carefully. 
> Did I get that wrong or am I correct that by having the RemoteIpFilter earlier in the Filter Chain then subsequent filters will see the "translated" address when calling request.getRemoteAddr()

It's ok to use with `RemoteIpFilter`.

BTW, I think we should also copy WARNING to Javadoc of `RateLimitFilter`. :)

rmaucher: Great doc ! (of course, nobody reads it)

java/org/apache/catalina/filters/RateLimitFilter.java
```diff
@@ -0,0 +1,230 @@
...
+
+        actualRequests = (int) Math.round(bucketCounter.getRatio() * bucketRequests);
+
+        log.info(sm.getString("rateLimitFilter.initialized",
+            super.getFilterName(), bucketRequests, bucketDuration, getActualRequests(),
+            getActualDurationInSeconds(), (!enforce ? "Not " : "") + "enforcing")
+        );
+    }
+
+    /**
```
ChristopherSchultz, line 158: This is useless javadoc. It replicates the generic documentation for `Filter.doFilter` and adds nothing.

If you want to describe how *this* filter works, then by all means, do so. But replicating the superclass/interface documentation is not useful, here.

java/org/apache/catalina/util/TimeBucketCounter.java
ChristopherSchultz: It depends upon how far back you want to back-port. Tomcat 8.5.x runs on Java 7. Though we build using Java 11, I think this should be buildable on Java 7 as well. Feel free to use lambdas in 9.0+ but please use the long-form in 8.5.x.

java/org/apache/catalina/filters/RateLimitFilter.java
isapir: Yeah, well, the IDE graciously offered to autocomplete the @Override method definition with the Javadoc and I accepted :(

I will remove that, but is that the right place to describe how this filter works?  I will look at other filter implementations for guidance.

java/org/apache/catalina/util/TimeBucketCounter.java
isapir: Looks like Collection.removeIf() has been added in Java 8 [1].  The code also makes use of Map.computeIfAbsent() which was also add in Java 8.

Since Tomcat 8.5 EOL is exactly a year from now [2], we need to decide whether we should port this back to 8.5.  The main question that I have to Tomcat 8.5 users is: what's holding you back from upgrading to Tomcat 9.0?

[1] https://docs.oracle.com/javase/8/docs/api/java/util/Collection.html#removeIf-java.util.function.Predicate-
[2] https://tomcat.apache.org/tomcat-85-eol.html

ChristopherSchultz: Inertia.

Feel free to just _not_ back-port to 8.5.x.

java/org/apache/catalina/filters/RateLimitFilter.java
ChristopherSchultz: I'd leave the javadoc missing. Document the behavior in the class-level documentation and just ignore the doFilter method. The entire purpose of the Filter class is to host the doFilter method, so ... they are kind of the same thing. There's no reason to document the individual method IMO.

java/org/apache/catalina/util/TimeBucketCounter.java
isapir: Chris - Let's see if the Filter gets any traction.  If we notice from mailing list conversations etc. that it is popular enough then we can port it back to 8.5.x with the small required changes.

java/org/apache/catalina/filters/RateLimitFilter.java
isapir: Chris - sounds good.  Will do.

java/org/apache/catalina/util/TimeBucketCounter.java
```diff
@@ -0,0 +1,217 @@
...
+        while (bitCheck > 1) {
+            bitCheck = pof2 >> ++bits;
+        }
+
+        this.numBits = bits;
+
+        this.ratio = ratioToPowerOf2(durationMillis);
+
+        int cleanupsPerBucketDuration = (durationMillis >= 60_000) ? 6 : 3;
+        Thread mt = new MaintenanceThread(durationMillis / cleanupsPerBucketDuration);
```
rmannibucau, line 75: guess this is not needed to 1. use a dedicated thread for that which prevents an app to have it, 2. not await it properly (can leak only sleep time but still) so I would use tomcat `Server` utility `ScheduledExecutorService` to handle it properly and in `destroy` just cancel the related task.

java/org/apache/catalina/util/TimeBucketCounter.java
```diff
@@ -0,0 +1,217 @@
...
+/**
+ * this class maintains a thread safe hash map that has timestamp-based buckets
+ * followed by a string for a key, and a counter for a value. each time the
+ * increment() method is called it adds the key if it does not exist, increments
+ * its value and returns it.
+ *
+ * a maintenance thread cleans up keys that are prefixed by previous timestamp
+ * buckets.
+ */
+public class TimeBucketCounter {
```
rmannibucau, line 32: can it be worth using an interface there? This is one impl but not sure it fits all cases

java/org/apache/catalina/util/TimeBucketCounter.java
```diff
@@ -0,0 +1,217 @@
...
+     * increments the counter for the passed identifier in the current time
+     * bucket and returns the new value
+     *
+     * @param identifier an identifier for which we want to maintain count, e.g. IP Address
+     * @return the count within the current time bucket
+     */
+    public final int increment(String identifier) {
+        String key = getCurrentBucketPrefix() + "-" + identifier;
+        AtomicInteger ai = map.computeIfAbsent(key, v -> new AtomicInteger());
+        return ai.incrementAndGet();
```
rmannibucau, line 89: is it worth using `updateAndGet` or accumulate flavors? here you can end up enabling requests if you do a full rotation having enough requests in the configured window

java/org/apache/catalina/filters/RateLimitFilter.java
```diff
@@ -0,0 +1,230 @@
...
+     * @throws IOException      if an I/O error occurs during this filter's
+     *                          processing of the request
+     * @throws ServletException if the processing fails for any other reason
+     */
+    @Override
+    public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain)
+                            throws IOException, ServletException {
+
+        String ipAddr = request.getRemoteAddr();
+        int reqCount = bucketCounter.increment(ipAddr);
```
rmannibucau, line 193: shouldnt the "key" be configurable? thinking to common cases where it often chains (order can vary):

1. global (key=constant=app)
2. user (key=req.principal.name)
3. ip

So having it configurable enables to get this behavior without overriding it, wdyt?

isapir: I considered something similar at first, but I didn't want to over complicate the first commit.  More features can be added later, for sure.

java/org/apache/catalina/util/TimeBucketCounter.java
isapir: I also considered that one and that's the reason I added it to the utils package instead of an inner class of the RateLimitFilter, but I think that it would make more sense to extract an interface once we have a good idea for other implementations.

The problem with APIs is that once they are published it's very hard to change or get rid of them, so I think that after some time in the wild we will have a better idea of what it can or should be.

java/org/apache/catalina/util/TimeBucketCounter.java
isapir: I will look into that, thanks!

java/org/apache/catalina/util/TimeBucketCounter.java
isapir: One idea that I had was to allow to configure certain paths with a different "weight", rather than configure a different Filter for different path, e.g. while requests to the homepage are counted as 1, a request for a login page might count as 10.

That way, you can configure the Filter for 300 requests per minute, for example, which will allow for 300 requests to most of the site, but only 30 requests to the login page, mitigating brute force attacks that try to guess a password.

If we do that then `updateAndGet()` might be more applicable, but again, I want to introduce this filter and get some feedback before adding more features to it.

Also, we have another thread on the PR about back porting this feature, and if we decide to go all the way back to Tomcat 8.5 then this would be another one of the things that needs to be rewritten to run on Java 7.

rmannibucau: this can easily be mimic-ed for java 7 but point is right now you can disable the rate limit if you get enough requests. If you take the extreme case for a rate limiter - DDoS - then you can just not reach this goal so think it is worth ensuring you don't go > max+1.

isapir: @rmannibucau Perhaps I'm missing something.  Can you please explain how `updateAndGet()` is better than `incrementAndGet()`?  

Looking at the implementation of both they look very similar other than the fact that `updateAndGet()` would have to call `updateFunction.applyAsInt()` which would add unnecessary overhead.  Both implementations call `weakCompareAndSetInt()`.

rmannibucau: overhead will be 0 since it will be inlined at some point but the function enables to set the max instead of incrementing and getting the issue i mention. Alternatively you can workaround it by doing `if (incrAndGet > max) v.set(max+1)`, it is no more atomic but likely saner than nothing.

java/org/apache/catalina/util/TimeBucketCounter.java
rmaucher: Using the background events from Tomcat would be nice, but this is not a valve. A regular filter in a regular webapp cannot use it.
In EE there is https://jakarta.ee/specifications/concurrency/3.0/ but Tomcat does not provide it.

rmannibucau: https://jakarta.ee/specifications/concurrency/3.0/ does not bring anything to java se except some useless listeners for that case so it is fine to stay away from it but goal is to reuse some banalised pool from tomcat instead of creating multiple leaky active threads.
I'm not sure why this filter wouldn't use it since the filter is an internal of tomcat as the background pool, in terms of classloading there is no strong blocker - and worse case you can still get the scheduled executor service from a servlet context attribute if so which would enable to keep the filter standard and enable tomcat to inject the background pool there if the filter is present but dont think it is needed - and the filter will not be reused in another server so guess it should be fine.

rmaucher: Sorry for liking that spec ... I like it because it provides a standard way to pass scheduling services to the app. Also I had to reimplement "robust" scheduling (logging and reschedule when something doesn't work). As a result, if given the choice, I would prefer having it than not having it.
Right now, there's no way to pass Tomcat's utility executor to the filter, so this PR should keep the dedicated thread.

rmannibucau: > I like it because it provides a standard way to pass scheduling services to the app.

But no way to configure it, to define if it leaks between apps or not (here we want it leaks) and other common pitfalls :(.

That said the way to link it to tomcat is JNDI and I guess the background scheduler could be made boundable in JNDI too quite easily so the solution can still converge.

rmaucher: Yes, JNDI is supposed to be used, but it's not very good here (users can disable naming, also still need to add an object factory for that, plus configuration).
Tomcat sets some custom Servlet context attributes in ServletContext.startInternal, this can be used for some scheduling service. I think using a thread for this filter is "ok" until it is removed once/if the feature is added.

rmannibucau: what about this compromise: in init store a scheduledexecutorservice (one thread) in the servletcontext if it does not already exists else reuse, will at least enable to chain filters without recreating it, wdyt? acceptable compromise?

java/org/apache/catalina/util/TimeBucketCounter.java
isapir: @rmannibucau But you still didn't explain how that would be better than the current code and didn't post any references.  We should always push better code when possible, but can you explain what problem that change would solve and how?  Thank you.

rmannibucau: @isapir well I tried to explain in your case if you have a long window (for business reason, it is a common case) you can overflow "int.max" and start allowing requests which should be forbidden whereas using an atomic "update" instead of blind increment logic enables to not increment when you already are higher than the max (therefore my proposal to reset the counter to max+1 when increment > max+1). This is easy to get in an unit test even if "slow".

java/org/apache/catalina/util/TimeBucketCounter.java
rmaucher: Exposing Tomcat's utility executor is only doable when the security manager is not enabled. This is now done with a new Servlet context attribute in Tomcat trunk (11) since no security manager at all.
In all other cases, the filter should use a thread as it is doing right now, so there's nothing to fix in the PR in that area.

rmaucher: I backported to 10.1 and 9.0, enabled when there's no security manager. This should be good enough and this filter can now use it instead of its dedicated thread (after merging it in is fine by me, this is still a minor improvement overall), falling back to creating an executor if not available.

java/org/apache/catalina/util/TimeBucketCounter.java
isapir: To exceed Integer.MAX_VALUE (2,147,483,647) in a day, for example, one would need to hit the server at the rate of almost 25,000 Requests per Second.  Handling cases like that are not a use case for this filter.

rmannibucau: Ok-ish but so should maybe be written as a limitation in the filter class javadoc cause rate limiting is way more generic than current impl (which is ok but reading the doc it is not obvious IMHO).

PR #608, opened 2023-04-11
Remove uses of classes in deprecated package javax.security.cert

eirbjo:
Efforts are underway in the OpenJDK project to remove the long deprecated-for-removal classes in the package `javax.security.cert`. These classes were introduced for backwards compatibility concerns with the unbundled JSSE release for JDK 1.2/1.3, but their use have been discouraged since they were introduced.

It would be good to update Tomcat to not depend on / use these archaic APIs.

See https://bugs.openjdk.org/browse/JDK-8227024 and the corresponding CSR https://bugs.openjdk.org/browse/JDK-8227395


markt-asf:
Thanks for the PR.

Note for reviewers: There is a default method in Java 17 (but not Java 11) so this change is safe in Tomcat 11 but can't be back-ported to earlier versions.

PR #609, opened 2023-04-12
Unify Windows/Linux package filesets

VirtualTim:
This looks like a bigger change than it is, simply because I sorted the list.

**Changes**:
* Sorted the Windows/Linux include/exclude lists, to make it easy to compare the two
* **Windows**
  * Removed exclusion of `service.bat`.
  * Added exclusion of `bin/*.sh`
* **Linux**
  * Removed the exclusion of `conf/**`
    * This now matches Windows, and I think this was intended.
  * Changed the exclusion `bin/service.bat` to `bin/*.bat`
  * Removed the exclusion of `src/**`, as it's not excluded in Windows, and I don't think it's needed.

The "big" change is that now `service.bat` is included in the windows zip. I don't know why it was ever excluded, and when I asked about it in the mailing list I was told to create a PR.

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -2808,24 +2808,24 @@ asf.ldap.username=${release.asfusername}
          encoding="UTF8">
       <zipfileset file="${tomcat.dist}" fullpath="${final.name}"/>
       <zipfileset dir="${tomcat.dist}" prefix="${final.name}">
-        <include name="bin/**"/>
-        <include name="conf/**"/>
-        <include name="logs/**"/>
-        <include name="lib/**"/>
-        <include name="webapps/**"/>
-        <include name="work/**"/>
-        <include name="temp/**"/>
-        <include name="LICENSE"/>
-        <include name="NOTICE"/>
-        <include name="README.md"/>
-        <include name="RELEASE-NOTES"/>
-        <include name="RUNNING.txt"/>
+        <include name="bin/**" />
         <include name="BUILDING.txt" />
-        <include name="CONTRIBUTING.md"/>
-        <exclude name="bin/service.bat"/>
-        <exclude name="bin/x64/"/>
-        <exclude name="bin/*.exe"/>
-        <exclude name="bin/*.dll"/>
+        <include name="conf/**" />
+        <include name="CONTRIBUTING.md" />
+        <include name="lib/**" />
+        <include name="LICENSE" />
+        <include name="logs/**" />
+        <include name="NOTICE" />
+        <include name="README.md" />
+        <include name="RELEASE-NOTES" />
+        <include name="RUNNING.txt" />
+        <include name="temp/**" />
+        <include name="webapps/**" />
+        <include name="work/**" />
+        <exclude name="bin/*.dll" />
+        <exclude name="bin/*.exe" />
+        <exclude name="bin/*.sh" />
+        <exclude name="bin/x64/" />
       </zipfileset>
     </zip>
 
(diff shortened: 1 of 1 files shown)


markt-asf:
The build script would benefit from some comments here. The core .zip and .tar.gz distros are not intended to be Windows/Linux distros. They should both be platform neutral (i.e. work on both). The only difference should be the line endings. Both should have .bat and .sh files.

The Windows service files (service.bat,.exe and .dll files) are only meant to be in the Windows specific distributions.

Historically, service.bat wasn't included in the Windows installer (built via a separate script) but it is now included. There isn't much point including service.bat if the associated binaries are not present.

The conf/** files are not excluded from the tar.gz distribution, they are added separately using a different permission set.

There are essentially three sets of files being added:
 - the standard directories (bin, conf, lib etc)
 - the top-level files (LICENSE, README, etc)
 - and the Windows binaries

I think it may be clearer if, rather than sorting all these files as one group, they are kept as three groups (perhaps another opportunity for some brief comments) but sorted within the groups.

markt-asf:
tomcat10.exe is a renamed Commons Daemon ProcRun executable. It is not a wrapper around the JAR (at least there is no Java code embedded in that exe). service.bat is useless without tomcat10.exe and tomcat10w.exe. It would make more sense to deploy the Windows specific binaries to Maven.

The conf/** files are included before the main set of files for the .tar.gz distro.

VirtualTim:
OK, so what do you think the end goal should look like?
org.apache.tomcat/tomcat is currently described as "Binary distribution of Apache Tomcat".
Currently that contains a tar.gz that works on Windows/Linux, a zip that does not work on Linux (this is why I assume they were mean to be platform-specific), and no way (out of the box) to run it as a Windows service.

So here are my ideas:

1. Split org.apache.tomcat/tomcat in two, with a Windows and a Linux classifier. Ie:
1.1. tomcat-11.0.0-M4-linux.tar.gz
1.1.1. contains no bat files
1.2. tomcat-11.0.0-M4-windows.zip
1.2.1. contains no sh files
1.2.2. contains tomcat10.exe, tomcat10w.exe, service.bat

2. From all platform-specific stuff (bat, sh, exe) from org.apache.tomcat/tomcat.
2.1. Create two new platform-specific artefacts that depend on org.apache.tomcat/tomcat, and contain the required bat, sh, etc.

3. Just put everything in one tar.gz. I'm pretty sure Windows will be able to deal with UNIX line endings and permissions.

4. Add the service.bat to org.apache.tomcat/tomcat.
4.1. This it the only option that does not break compatibility for anyone already using this artefact.

Personally I'd go with 4., as preserving backwards compatibility is important, and consider 1. or 2. for tomcat 12/13.

markt-asf:
What end goal are you looking for?  Going back to users@ thread from a month ago, it looks like you want to be able to obtain a version of Tomcat from Maven Central that can be installed as a Windows service.

Adding service.bat to the platform neutral .zip and .tar.gz archives doesn't help - they don't contain the necessary binaries.

Adding the Windows specific binaries to Maven Central (as I suggested in the users@ thread) still looks to be the cleanest solution. It also gets you Tomcat Native.

What makes you say the zip distribution doesn't work on Linux. I've just tested it and it works as expected.

ChristopherSchultz:
Why not have a separate Maven package for *just* the binaries / scripts required for the platform, and a dependency on the core package?

I am -1 on labeling the distribution containing the `.sh` scripts as `-linux`. It is much more general than *Linux*. Feel free to call it `-unix` if you have to, even though you can run it perfectly fine on Windows as well...

markt-asf:
My initial reasoning for not having Maven packages (one for each architecture) with just the binaries in it is that we don't have that anywhere else at the moment whereas we do have the full Windows binaries. The later would be easier to setup. Thinking about it, I think a Windows packages with everything will also be easier for users.

PR #610, opened 2023-04-18
Enhance PEMFile to load from an inputstream too

rmannibucau:
The file abstraction is sometimes bothering (tomcat embedded to mention it) so enable to use PEMFile logic without having to dump a temp file.


ChristopherSchultz (approved):
LGTM

ChristopherSchultz:
I've merged this separately, including a changelog entry.

Thanks for the PR.

PR #611, opened 2023-04-24
Ensure brackets are closed in PoolProperties.toString

reftel:
Fixes minor annoyance when analyzing log files


aooohan (approved):
Unit test may be not be needed. But it doesn't matter. ; )

aooohan:
Thanks for the PR.

PR #613, opened 2023-04-28
transferTo can transfer only part of the file, so need to loop

jackshirazi:
We (Elastic APM) have a customer (using tomcat 9.0.72) that is hitting this bug. The file copy is getting interrupted (likely from CPU time slicing signals, which is happening because of additional profiling executing in the background) and so only partly completes. As the current implementation doesn't loop, this results in a corrupt war/zip file that can't be loaded. To complete the file copy, transferTo needs to loop


ChristopherSchultz:
LGTM

ChristopherSchultz:
@rmaucher applied this PR earlier today.

PR #614, opened 2023-05-08
Remove unnecessary reset

kang-hl:


markt-asf:
The method is necessary. It releases the reference to the SocketWrapper, allowing it ti be GC'd.

kang-hl:
NioSocketWrapper is referenced by NioChannel, and it may need to release the reference.

kang-hl:
> The method is necessary. It releases the reference to the SocketWrapper, allowing it ti be GC'd.

NioSocketWrapper is referenced by NioChannel, and it may need to release the reference.

PR #615, opened 2023-05-11
Add a reset method to release reference

kang-hl:


aooohan:
I've look at this, I think this change is a bit complicated. Resetting the channel while closing it is ok.

aooohan:
I've done this and also cleared up reference elsewhere.

PR #616, opened 2023-05-12
Fix backport

neillee95:


markt-asf:
No explanation was provided for this PR. It is tempting to close it with a similar lack of explanation. However...

The purpose of the PR is not obvious from the context (the pros/cons of `trySetAccessible()` and `setAccessible(true)` are not clear cut) and the current code is as intended.

PR #617, opened 2023-05-18
#MessageBytes toString rollback

neo-ytx:
1. I think toString method should not change the object's status. Because this method is used for return a String. Most people think they can get object's stringValue from this method, and will not change the object's status.
2. The issue I found: After I integrate openTelemetry with my tomcat web services, the service always return 404. The openTelemetry use the toString method, then tomcat think it is a str not byte and return 404 error.

### DIFF
--- a/java/org/apache/tomcat/util/buf/MessageBytes.java
+++ b/java/org/apache/tomcat/util/buf/MessageBytes.java
@@ -75,6 +75,8 @@ public final class MessageBytes implements Cloneable, Serializable {
     // String
     private String strValue;
 
+    private boolean hasStrValue=false;
+
     /**
      * Creates a new, uninitialized MessageBytes object.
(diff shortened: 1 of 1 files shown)


aooohan:
Hmm. A similar issue has been raised by other user before, look this https://lists.apache.org/list?users@tomcat.apache.org:2023-3:openTelemetry

Hope this helps.

neo-ytx:
Thanks a lot~. And I still have one question: Is the toString method changing object status a good behavior?

aooohan:
> Thanks a lot~. And I still have one question: Is the toString method changing object status a good behavior?

No, not a good behavior. This Class has been refactored before to solve some issues, and I'm not familiar with this, you can post your question to user mailing list to discuss. ; )

PR #620, opened 2023-05-21
Update ci.yml

DanielR-spec:
update description file


markt-asf:
Proposed change not correct.

PR #621, opened 2023-05-24
Bug 66609 - invalid XML in directory listing with file names containing "&" and "'"

alexkachanov:
* fix and junit test


java/org/apache/catalina/util/URLEncoder.java
```diff
@@ -40,6 +40,7 @@ public final class URLEncoder implements Cloneable {
 
     public static final URLEncoder DEFAULT = new URLEncoder();
     public static final URLEncoder QUERY = new URLEncoder();
+    public static final URLEncoder XML = new URLEncoder();
```
alexkachanov, line 43: added new XML encoder that encodes strings that are valid in XML

java/org/apache/catalina/util/URLEncoder.java
```diff
@@ -95,6 +96,8 @@ public final class URLEncoder implements Cloneable {
         QUERY.addSafeCharacter('_');
         QUERY.addSafeCharacter('=');
         QUERY.addSafeCharacter('&');
+        
+        XML.addSafeCharacter('/');
```
alexkachanov, line 100: "/" is a valid symbol

java/org/apache/catalina/servlets/DefaultServlet.java
```diff
@@ -1650,7 +1654,7 @@ protected InputStream renderXml(HttpServletRequest request, String contextPath,
               .append('\'');
             sb.append(" urlPath='")
               .append(rewrittenContextPath)
-              .append(rewriteUrl(directoryWebappPath + entry))
+              .append(rewriteUrl(directoryWebappPath + entry, URLEncoder.XML))
```
alexkachanov, line 1657: using XML encoder to encode path attribute which may contain "'" and "&" symbols

markt-asf:
(Ab)using the URLEncoder for this seems wrong. Tomcat has a class specifically for escaping values in content - `org.apache.tomcat.util.security.Escape`. The fix should use this.

Further, it appears that there are still multiple implementations of XML escaping in the Tomcat code base. These need to be rationalised to use `Escape` class.

I agree having a test case for this is an improvement. I'd like to see that test case cover all the XML characters that need to be escaped and also test the WebDAV case as well as the directory listing case.

alexkachanov:
I was using URLEncoder class because it was used in original rewriteUrl method. So I wanted to keep it consistent and keep the changes it to minimum. I see that Escape.xml(String content) makes sense here. Let me change the patch

java/org/apache/catalina/servlets/DefaultServlet.java
```diff
@@ -1650,7 +1650,7 @@ protected InputStream renderXml(HttpServletRequest request, String contextPath,
               .append('\'');
             sb.append(" urlPath='")
               .append(rewrittenContextPath)
-              .append(rewriteUrl(directoryWebappPath + entry))
+              .append(Escape.xml(directoryWebappPath + entry))
```
alexkachanov, line 1653: using Escape.xml instead of rewriteUrl

markt-asf:
Thanks for the PR. Applied manually with a few tweaks (e.g. CheckStyle config).

PR #623, opened 2023-05-30
BZ 66513: Enforce one concurrent request per session requirement

markt-asf:
This is untested.

If accepted, this will need some documentation updates.

The purpose of this draft is to obtain early feedback on the approach.


rmaucher:
That seems like the better way to do it.

rmannibucau:
* Is there a plan for the map eviction? background thread?
* How does it work in a cluster since counter is local, should it be a cluster event?

markt-asf:
Map eviction is handled in the finally block.

That we can only make this work on a single node is why Rémy is proposing that we remove/deprecate this Valve.

aooohan (approved):
LGTM.

markt-asf:
I haven't deprecated this yet. I'm waiting on input from the OP of BZ 66513 and/or @jfclere as the original author of the Valve in case there is a use case I have missed.

rmaucher:
Very clean overall, esp the javadoc update.

jfclere:
@markt-asf that looks better that what was there ;-)

markt-asf:
> @markt-asf that looks better that what was there ;-)

Thanks.

Can you recall if there was a use case for this Valve on a single node? Or was it always intended for use behind a non-sticky load-balancer and, therefore, it was up to the client to enforce the "no more than one concurrent request per session" rule?

jfclere:
trying to remember...  I guess I wrote to save and restore session information for a SINGLE tomcat node.

PR #625, opened 2023-06-12
Remove transport.bio from catalina-tribes.jar.tmp.bnd

bekionel:
Since package org.apache.catalina.tribes.transport.bio was deprecated in tomcat 10, https://tomcat.apache.org/tomcat-9.0-doc/api/org/apache/catalina/tribes/transport/bio/package-summary.html , references to it in catalina-tribes.jar should be removed.


markt-asf:
Back-ported to 10.1.x as the package was removed in 10.1.x. Not back-ported further as the package is still present in 9.0.x and earlier.

PR #629, opened 2023-06-21
Documentation change for Digest Authenticator Valve - spell corrected for attribute algorithm

gohilmca:
this change is with respect to issue #66662


markt-asf:
Many thanks. I'll do the back-port now.

PR #634, opened 2023-07-11
Bug 66684:  set the revocationEnabled as FALSE when crl file is empty

DigitalFatCat:
https://bz.apache.org/bugzilla/show_bug.cgi?id=66684

Recently, we tested the use of the crLFile configuration （in server.xml） in the scenario where two-way certificate authentication is enabled. When the file pointed to by the crlFile configuration item is an empty file (the file exists but the file content is blank), Tomcat cannot provide services and SSL HandShake reports an error（certificate_unknown）. When crlFile is not configured or the content in crlFile is correct, no error is reported.

I'm thinking that don't need to set the revocationEnabled parameter to true (In org.apache.tomcat.util.net.SSLUtilBase#getParameters 498L) when the file content is blank.


markt-asf:
This PR will not be merged. See BZ 66684 for details of why.

PR #639, opened 2023-07-17
Custom ssl engine parameters

fmarco76:
Deploying tomcat with a custom ssl engine could require to perform some extra configuration before its usage. As an example, [tomcatjss](https://github.com/dogtagpki/tomcatjss/) project needs to provide the actual IPs of client and server for audit purpose.
To provide the IPs we are working on a patch which require to copy several tomcat methods from _NIO_ endpoint and channel because the current implementation does not allow to access the ssl engine before it is used (the method to be replaced are in [this PR](https://github.com/dogtagpki/tomcatjss/pull/73)). 

Changes in this PR allow to extend tomcat _NIO_ classes and modify the ssl engine without reimplementing the existing methods.

### DIFF
--- a/java/org/apache/tomcat/util/net/AbstractNetworkChannelEndpoint.java
+++ b/java/org/apache/tomcat/util/net/AbstractNetworkChannelEndpoint.java
@@ -26,6 +26,7 @@ public abstract class AbstractNetworkChannelEndpoint<S extends Channel, U extend
         extends AbstractEndpoint<S, U> {
 
     protected abstract NetworkChannel getServerSocket();
+    protected abstract S createChannel(SocketBufferHandler buffer);
 
 
     @Override
--- a/java/org/apache/tomcat/util/net/Nio2Endpoint.java
+++ b/java/org/apache/tomcat/util/net/Nio2Endpoint.java
@@ -307,11 +307,7 @@ public class Nio2Endpoint extends AbstractNetworkChannelEndpoint<Nio2Channel,Asy
                         socketProperties.getAppReadBufSize(),
                         socketProperties.getAppWriteBufSize(),
                         socketProperties.getDirectBuffer());
-                if (isSSLEnabled()) {
-                    channel = new SecureNio2Channel(bufhandler, this);
-                } else {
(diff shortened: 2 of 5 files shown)


fmarco76:
> The only thing that caught my eye was that the signature of `AbstractNetworkChannelEndpoint.createChannel` has changed by widening the return value from `NetworkChannel` to `Channel`. This change is not backward compatible and we might want to introduce a new method instead of changing the signature of the current one.

Hi @ChristopherSchultz, the `AbstractNetworkChannelEndpoint.createChannel` is introduced by this PR so I am not sure this is widening the return value. Currently, the channel is created inside the method `NioEndpoint.setSocketOptions` and does not invokes other methods. I have introduced the `createChannel` so I can override  it in a subclass  and associate a different channel keeping the endpoint logic.

The only problem with back-porting I can think of it is that classes extending the `AbstractNetworkChannelEndpoint` will need to implement this method. To avoid this I can convert to a concrete method which throw an exception when invoked. In this way classes not using it will still work and the other have to override it.  
Would this change make it  good for back-porting or should I look at different approaches. Thanks!

rmaucher:
This proposed change looks good to me.

rmaucher:
I merged this change to Tomcat main branch (11). Backporting is not straightforward however as this is an API change stacked on top of a previous API change for Tomcat 11, so the code that uses it would be tied to 11 anyway.

fmarco76:
@rmaucher thanks for the merge. I have tested cherry-picking to tomcat 9 and it required a couple of minor fixes to work. Is there something I can do to facilitate the backport?

rmaucher:
> @rmaucher thanks for the merge. I have tested cherry-picking to tomcat 9 and it required a couple of minor fixes to work. Is there something I can do to facilitate the backport?

This seems difficult, AbstractJsseEndpoint would be the spot (same for 10.1), but it would be an incompatible API change (so probably not ok), and code written for 11 would not be compatible anyway.

fmarco76:
> but it would be an incompatible API change

Yes,  the API would be slightly different between version 9 and 11.

PR #644, opened 2023-08-08
Warn when non default ciphers are removed because the engine rejects them.

jfclere:
When using nmap --script ssl-enum-ciphers -p port host
the ciphers proposed are often a small subset of the expect ones.
A warning helps to see what is rejected.

### DIFF
--- a/java/org/apache/tomcat/util/net/SSLUtilBase.java
+++ b/java/org/apache/tomcat/util/net/SSLUtilBase.java
@@ -127,10 +127,11 @@ public abstract class SSLUtilBase implements SSLUtil {
             // TODO: sslHostConfig can query that with Panama, but skip for now
             this.enabledCiphers = new String[0];
         } else {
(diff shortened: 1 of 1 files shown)


java/org/apache/tomcat/util/net/SSLUtilBase.java
```diff
@@ -122,6 +122,10 @@ protected SSLUtilBase(SSLHostConfigCertificate certificate, boolean warnTls13) {
         sslHostConfig.setTls13RenegotiationAvailable(isTls13RenegAuthAvailable());
 
         // Calculate the enabled ciphers
+        boolean warnOnSkip = true;
+        if (sslHostConfig.getCiphers().equals(sslHostConfig.DEFAULT_TLS_CIPHERS)) {
+            warnOnSkip = false;
+        }
```
markt-asf, line 128: `boolean warnOnSkip = !sslHostConfig.getCiphers().equals(sslHostConfig.DEFAULT_TLS_CIPHERS);`

And move this inside the block where it is used.

markt-asf:
LGTM

PR #647, opened 2023-08-22
Replaced synchronized with StampedLock

sergmain:
Replaced synchronized with StampedLock as preparation for virtual threads


markt-asf:
What is the justification for this change? Synchronization is not automatically an issue for virtual threads. I'm not seeing anything in this code that would be unfriendly to virtual threads.

sergmain:
From my point of view there are 2 approaches - analyze every synchronized or remove every synchronized without analyzing. My pull-request is with second approach.

markt-asf:
Synchronized blocks only need to be replaced if they contain blocking operations. This one clearly doesn't.

PR #648, opened 2023-09-05
Refactor 'getSession' Method in 'Request' Class to Eliminate Code Redundancy

xxeol2:
This Pull Request aims to reduce code duplication in the `getSession` method of the `Request` class.

The refactoring closely follows the structure and principles used in the `reset` method of `java.org.apache.catalina.connector.OutputBuffer`:
```java
public void reset() {
	reset(false);
}

public void reset(boolean resetWriterStreamFlags) {
	clear(bb);
	clear(cb);
	bytesWritten = 0;
	charsWritten = 0;
	if (resetWriterStreamFlags) {
		if (conv != null) {
			conv.recycle();
		}
		conv = null;
	}
	initial = true;
}
```

### DIFF
--- a/java/org/apache/catalina/connector/Request.java
+++ b/java/org/apache/catalina/connector/Request.java
@@ -2288,12 +2288,7 @@ public class Request implements HttpServletRequest {
      */
     @Override
     public HttpSession getSession() {
-        Session session = doGetSession(true);
-        if (session == null) {
-            return null;
-        }
-
-        return session.getSession();
+        getSession(true);
     }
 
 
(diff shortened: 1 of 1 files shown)


java/org/apache/catalina/connector/Request.java
```diff
@@ -2288,12 +2288,7 @@ public String getServletPath() {
      */
     @Override
     public HttpSession getSession() {
-        Session session = doGetSession(true);
-        if (session == null) {
-            return null;
-        }
-
-        return session.getSession();
+        getSession(true);
```
aooohan, line 2291: No results returned.
```suggestion
        return getSession(true);
```

aooohan (changes requested):

java/org/apache/catalina/connector/Request.java
xxeol2: Oh, I made a mistake!

aooohan (approved):

aooohan:
Thanks for the PR. ; )

PR #649, opened 2023-09-05
Eliminating duplicate execution of checkFacade logic

xxeol2:
In the current implementation of the `getSession` method in the `RequestFacade` class, the checkFacade logic was redundantly executed twice.
I improved this by eliminating the redundancy.


aooohan:
This change doesn't make any sense.

markt-asf:
The change makes sense to me. It removes a duplicate call to `checkFacade()`.

PR #651, opened 2023-09-05
Refactor SSL certificate population method

wonyongChoi05:
Simplified the SSL certificate population logic by removing the unnecessary null check and using an ArrayList for dynamic array management. The code is now more concise and readable, and it efficiently processes X.509 certificates.


aooohan:
This change doesn't make any sense.

markt-asf:
The switch to using a wildcard import would need to be reverted as would the changes to blank lines between methods.
The null check is necessary in some form and it is more concise than the alternative proposed in this PR.
The switch to using an array list has merit. I'll implement that change shortly.

wonyongChoi05:
> The switch to using a wildcard import would need to be reverted as would the changes to blank lines between methods. The null check is necessary in some form and it is more concise than the alternative proposed in this PR. The switch to using an array list has merit. I'll implement that change shortly.

Thank you for your answer.

PR #656, opened 2023-09-05
Optimization when call getSessionCookieName & getSessionUriParamName

greeng00se:
Existing code calls getConfiguredSessionCookieName even if the context is empty.

In getConfiguredSessionCookieName, only act to get the session cookie if the context is non-null.

I think it's possible to avoid calling getConfiguredSessionCookieName and use Default defined by spec if context is null.

It's a small optimisation, but I think it saves us one less function call.

The one thing that worries me is about the Priority comment

```
// Priority is:
// 1. Cookie name defined in context
// 2. Cookie name configured for app
// 3. Default defined by spec
```

For now, I didn't modify that part.

### DIFF
--- a/java/org/apache/catalina/util/SessionConfig.java
+++ b/java/org/apache/catalina/util/SessionConfig.java
@@ -32,14 +32,10 @@ public class SessionConfig {
      * @return the cookie name for the context
      */
     public static String getSessionCookieName(Context context) {
-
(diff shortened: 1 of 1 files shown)


greeng00se:
I didn't think about when neither priority 1 nor 2 is present in pr, sorry.

greeng00se:
sorry. Case 3 is not considered.

PR #658, opened 2023-09-05
Unify conditional statement format for some constants in http1 package

Jaeyoung22:
Although I totally agree that there could be some unnecessary parentheses to aid the readability of the code, I think that the unity of the code is still important for readability.

### Conditional statements of CR & LF in Http11InputBuffer
```java
// Line 445
if (prevChr == Constants.CR && chr != Constants.LF) {
    // ...
}

// Line 544
    // ...
}
``` 
Except for Line 372 that I changed, all conditional statements that have two conditions about CR and LF don't have parentheses for each condition.

### Conditional statements in ChunkedInputFilter
```java
// Line 344
if (chr == Constants.CR || chr == Constants.LF) {
    // ...
}

// Line 462
if (chr == Constants.CR || chr == Constants.LF) {
    // ...
}

// Line 551
if (chr == Constants.CR || chr == Constants.LF) {
    // ...
}
``` 
Except for Line 487 and 525 that I changed, all conditional statements that have two conditions about constants don't have parentheses for each condition.

### Logic about space and tab in Http11InputBuffer
```java
// Line 395, 453, 974(else if)
if (chr == Constants.SP || chr == Constants.HT) {
    // …
}

// Line 419, 515, 935
if (!(chr == Constants.SP || chr == Constants.HT)) {
    // …
}
``` 
Because of the above codes, I tried to change Line 1001 of Http11InputBuffer, and Line 577 of ChunkedInputFilter.

### DIFF
--- a/java/org/apache/catalina/ant/AbstractCatalinaTask.java
+++ b/java/org/apache/catalina/ant/AbstractCatalinaTask.java
@@ -233,7 +233,7 @@ public abstract class AbstractCatalinaTask extends BaseRedirectorHelperTask {
                 int ch = reader.read();
                 if (ch < 0) {
                     break;
-                } else if ((ch == '\r') || (ch == '\n')) {
+                } else if (ch == '\r' || ch == '\n') {
                     // in Win \r\n would cause handleOutput() to be called
                     // twice, the second time with an empty string,
                     // producing blank lines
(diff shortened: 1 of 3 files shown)


Jaeyoung22:
This Pull Request UNINTENTIONALLY sended because of github incident :( So I closed this PR and resend.

PR #659, opened 2023-09-05
Unify constant delimiters and Refactoring for better readability

parkmuhyeun:
## Description

This PR made the following two modifications
- Unify constant delimiters
- Refactoring for better readability

### Unify constant delimiters

Constants managed in the Constants class in the coyote package are separated by _ for better readability.

The Constants class in the ajp and http2 packages is fine, but the Constants class in the coyote and http11 packages has constants that break uniformity, so I fixed them.
- TRANSFERENCODING -> TRASNFER_ENCODING
- STAGE_ENDINPUT -> STAGE_END_INPUT
- STAGE_ENDOUTPUT -> STAGE_END_OUTPUT
- STAGE_KEEPALIVE -> STAGE_KEEP_ALIVE

### Refactoring for better readability

I found a switch statement in the StatusTransformer class with a magic number. I planned to refactor it, thinking it would be more readable if it was separated into constants.

I could have used the constants from coyote's Constants class as the comment says, but statusTransformer is now in the catalina package, which is a different package. So I added the relevant constants to the Constants class in the catalina package and used them.

+Also, I noticed that coyote's Constants constants are only used within the coyote package.

## Comments

This is my first PR, so please let me know if there's anything wrong!

### DIFF
--- a/java/org/apache/catalina/manager/StatusTransformer.java
+++ b/java/org/apache/catalina/manager/StatusTransformer.java
@@ -375,33 +375,33 @@ public class StatusTransformer {
 
         switch (stage) {
 
-        case (Constants.STAGE_PARSE):
+        case (org.apache.coyote.Constants.STAGE_PARSE):
             stageStr = "P";
             fullStatus = false;
             break;
-        case (Constants.STAGE_PREPARE):
+        case (org.apache.coyote.Constants.STAGE_PREPARE):
             stageStr = "P";
             fullStatus = false;
             break;
-        case (Constants.STAGE_SERVICE):
+        case (org.apache.coyote.Constants.STAGE_SERVICE):
             stageStr = "S";
             break;
-        case (Constants.STAGE_END_INPUT):
+        case (org.apache.coyote.Constants.STAGE_ENDINPUT):
             stageStr = "F";
             break;
-        case (Constants.STAGE_END_OUTPUT):
+        case (org.apache.coyote.Constants.STAGE_ENDOUTPUT):
             stageStr = "F";
             break;
-        case (Constants.STAGE_ENDED):
+        case (org.apache.coyote.Constants.STAGE_ENDED):
             stageStr = "R";
             fullStatus = false;
             break;
-        case (Constants.STAGE_KEEP_ALIVE):
+        case (org.apache.coyote.Constants.STAGE_KEEPALIVE):
             stageStr = "K";
             fullStatus = true;
             showRequest = false;
             break;
-        case (Constants.STAGE_NEW):
+        case (org.apache.coyote.Constants.STAGE_NEW):
             stageStr = "R";
             fullStatus = false;
             break;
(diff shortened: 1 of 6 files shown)


java/org/apache/catalina/manager/StatusTransformer.java
```diff
@@ -375,33 +375,33 @@ protected static void writeProcessorState(PrintWriter writer,
 
         switch (stage) {
 
-        case (1/*org.apache.coyote.Constants.STAGE_PARSE*/):
+        case (Constants.STAGE_PARSE):
```
markt-asf, line 378: Why not use the existing constants (the ones in the comments)?

java/org/apache/coyote/Constants.java
```diff
@@ -37,9 +37,9 @@ public final class Constants {
     public static final int STAGE_PARSE = 1;
     public static final int STAGE_PREPARE = 2;
     public static final int STAGE_SERVICE = 3;
-    public static final int STAGE_ENDINPUT = 4;
-    public static final int STAGE_ENDOUTPUT = 5;
-    public static final int STAGE_KEEPALIVE = 6;
+    public static final int STAGE_END_INPUT = 4;
+    public static final int STAGE_END_OUTPUT = 5;
+    public static final int STAGE_KEEP_ALIVE = 6;
```
markt-asf, line 42: We don't want to change the names of public constants unless we have a really good reason. The risk of breaking a third-party integration is too high.

markt-asf (changes requested):

parkmuhyeun:
I reverted the stage constants back and modified the switch statements to use the constants that were originally there!

markt-asf (changes requested):
I'm fairly sure this won't compile in its current state. There are more changes to constant names that need to be reverted.

parkmuhyeun:
Oh, sorry. I missed that one! I've fixed it.

java/org/apache/coyote/http11/Constants.java
```diff
@@ -106,7 +106,7 @@ public final class Constants {
     public static final String KEEP_ALIVE_HEADER_VALUE_TOKEN = "keep-alive";
     public static final String CHUNKED = "chunked";
     public static final byte[] ACK_BYTES = ByteChunk.convertToBytes("HTTP/1.1 100 " + CRLF + CRLF);
-    public static final String TRANSFERENCODING = "Transfer-Encoding";
+    public static final String TRANSFER_ENCODING = "Transfer-Encoding";
```
markt-asf, line 109: This change still needs to be reverted. It is a public constant and my previous comments in this PR about not changing public constants apply equally to this one.

markt-asf (changes requested):

parkmuhyeun:
TransferENCODING wasn't mentioned, so I didn't recognize it. I reverted all the stuff about constant names!

markt-asf:
Tx for the PR. I added a commit to remove an unnecessary line of code I spotted while reviewing the PR.

PR #660, opened 2023-09-05
Optimize when call getSessionCookieName & getSessionUriParamName

greeng00se:
### Description

Optimization when call getSessionCookieName & getSessionUriParamName

### Explanation

Existing code calls getConfiguredSessionCookieName even if the context is empty.
In getConfiguredSessionCookieName, only act to get the session cookie if the context is non-null.
I think it's possible to avoid calling getConfiguredSessionCookieName and use Default defined by spec if context is null.
It's a small optimisation, but I think it saves us one less function call.

Additionally, modified getConfiguredSessionCookieName to not require a null check on its result.

### Comment

One fewer comparison operation if context is non-null. and I think it makes the code a little easier to understand for the remaining priority comments.

### DIFF
--- a/java/org/apache/catalina/util/SessionConfig.java
+++ b/java/org/apache/catalina/util/SessionConfig.java
@@ -32,14 +32,10 @@ public class SessionConfig {
      * @return the cookie name for the context
      */
     public static String getSessionCookieName(Context context) {
-
-        String result = getConfiguredSessionCookieName(context);
-
-        if (result == null) {
-            result = DEFAULT_SESSION_COOKIE_NAME;
+        if (context == null) {
+            return DEFAULT_SESSION_COOKIE_NAME;
         }
-
-        return result;
+        return getConfiguredSessionCookieName(context, DEFAULT_SESSION_COOKIE_NAME);
     }
 
     /**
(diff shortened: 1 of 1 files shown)


markt-asf:
Saving a single function call is unlikely to make any difference. I'd expect the compiler to optimise that.

Changes that reduce duplication and/or improve the readability of the code are the ones that are most likely to be applied. If there is a trade-off between duplication and readability, the priority will nearly always be readability.

In this instance I think you'd be better keeping the null check for context in the static method. That further reduces duplication and has the benefit that all of the logic is then in a single place and the code would more clearly follow the priority order documented in the code.

greeng00se:
@markt-asf 
Thanks for the nice comment.
I hadn't considered that partial compiler optimisation.
I agree, readable code should be a priority and I moved the null-check logic to a static method...

java/org/apache/catalina/util/SessionConfig.java
```diff
@@ -49,38 +42,30 @@ public static String getSessionCookieName(Context context) {
...
+    private static String getConfiguredSessionCookieName(Context context, String defaultName) {
+        if (context == null) {
+            return defaultName;
+        }
         // Priority is:
         // 1. Cookie name defined in context
         // 2. Cookie name configured for app
         // 3. Default defined by spec
-        if (context != null) {
-            String cookieName = context.getSessionCookieName();
```
markt-asf, line 70: The code will be cleaner if you keep the original null check. That will also keep the code better aligned with the documented priority order.

markt-asf (changes requested):

java/org/apache/catalina/util/SessionConfig.java
greeng00se: Thank you for your attention to detail in the code. 
Could you please check the modified code?

markt-asf: LGTM

markt-asf (approved):

markt-asf:
Tx for the PR.

greeng00se:
@markt-asf thx for good comment XD

PR #661, opened 2023-09-05
Unify conditional statement format for some constants in http1 package

Jaeyoung22:
Although I totally agree that there could be some unnecessary parentheses to aid the readability of the code, I think that the unity of the code is still important for readability.

### Conditional statements of CR & LF in Http11InputBuffer
```java
// Line 445
if (prevChr == Constants.CR && chr != Constants.LF) {
    // ...
}

// Line 544
    // ...
}
``` 
Except for Line 372 that I changed, all conditional statements about CR and LF don't have parentheses for each condition.

### Conditional statements in ChunkedInputFilter
```java
// Line 344
if (chr == Constants.CR || chr == Constants.LF) {
    // ...
}

// Line 462
if (chr == Constants.CR || chr == Constants.LF) {
    // ...
}

// Line 551
if (chr == Constants.CR || chr == Constants.LF) {
    // ...
}
``` 
Except for Line 487 and 525 that I changed, all conditional statements about constants don't have parentheses for each condition.

### Logic about space and tab in Http11InputBuffer
```java
// Line 395, 453, 974(else if)
if (chr == Constants.SP || chr == Constants.HT) {
    // ...
}

// Line 419, 515, 935
if (!(chr == Constants.SP || chr == Constants.HT)) {
    // ...
}
``` 
Because of the above codes, I tried to change Line 1001 of Http11InputBuffer, and Line 577 of ChunkedInputFilter.

### DIFF
--- a/java/org/apache/coyote/http11/Http11InputBuffer.java
+++ b/java/org/apache/coyote/http11/Http11InputBuffer.java
@@ -998,7 +998,7 @@ public class Http11InputBuffer implements InputBuffer, ApplicationBufferHandler 
 
             byte peek = byteBuffer.get(byteBuffer.position());
             if (headerParsePos == HeaderParsePosition.HEADER_MULTI_LINE) {
-                if ((peek != Constants.SP) && (peek != Constants.HT)) {
+                if (!(peek == Constants.SP || peek == Constants.HT)) {
                     headerParsePos = HeaderParsePosition.HEADER_START;
                     break;
                 } else {
--- a/java/org/apache/coyote/http11/Http11InputBuffer.java
+++ b/java/org/apache/coyote/http11/Http11InputBuffer.java
@@ -369,7 +369,7 @@ public class Http11InputBuffer implements InputBuffer, ApplicationBufferHandler 
                     request.setStartTimeNanos(System.nanoTime());
                 }
                 chr = byteBuffer.get();
-            } while ((chr == Constants.CR) || (chr == Constants.LF));
+            } while (chr == Constants.CR || chr == Constants.LF);
             byteBuffer.position(byteBuffer.position() - 1);
 
             parsingRequestLineStart = byteBuffer.position();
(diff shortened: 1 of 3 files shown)


java/org/apache/coyote/http11/Http11InputBuffer.java
```diff
@@ -998,7 +998,7 @@ private HeaderParseStatus parseHeader() throws IOException {
 
             byte peek = byteBuffer.get(byteBuffer.position());
             if (headerParsePos == HeaderParsePosition.HEADER_MULTI_LINE) {
-                if ((peek != Constants.SP) && (peek != Constants.HT)) {
+                if (!(peek == Constants.SP || peek == Constants.HT)) {
```
markt-asf, line 1001: Removing the unnecessary brackets is fine. Applying De Morgan's law is not. Yes, the result is the same but the intention is less obvious than with the original form so please keep it.

java/org/apache/coyote/http11/filters/ChunkedInputFilter.java
```diff
@@ -574,7 +574,7 @@ private boolean parseHeader() throws IOException {
             }
 
             chr = readChunk.get(readChunk.position());
-            if ((chr != Constants.SP) && (chr != Constants.HT)) {
+            if (!(chr == Constants.SP || chr == Constants.HT)) {
```
markt-asf, line 577: As above - remove application of De Morgan's law.

markt-asf (changes requested):
Fixing unnecessary parentheses across multiple files is an improvement on fixing a single instance but this still feels like a relatively small change. I would very much prefer to see such fixes span an entire package.

markt-asf:
See #663 - a single PR would be preferred.

markt-asf (approved):

markt-asf:
Thanks for the PR. It is so much more efficient to review this sort of change at this sort of scale.

Thanks for the De Morgan's law changes. I think those conditions are a lot easier to follow now.

I'm not sure if some of the Panama code is generated or not but I don't think applying this change will hurt even if it is.

PR #662, opened 2023-09-06
Simplify and enhance charset extraction from content type

wonyongChoi05:
Upon reviewing these changes, I've noticed a significant improvement in code readability. This enhanced readability stands out and will make a substantial difference in maintaining and debugging the code, as well as facilitating collaboration among developers.

I believe that this change elevates code quality and simplifies maintenance. By simplifying and making the code more straightforward, we can prevent errors and enhance code comprehensibility.

Once again, thank you for your contribution to this effort. If there are any additional improvements or if you need any further feedback, please feel free to let me know!

### DIFF
--- a/java/org/apache/coyote/Request.java
+++ b/java/org/apache/coyote/Request.java
@@ -16,17 +16,6 @@
  */
 package org.apache.coyote;
 
-import jakarta.servlet.ReadListener;
-import jakarta.servlet.ServletConnection;
-import org.apache.tomcat.util.buf.CharsetHolder;
-import org.apache.tomcat.util.buf.MessageBytes;
-import org.apache.tomcat.util.buf.UDecoder;
-import org.apache.tomcat.util.http.MimeHeaders;
-import org.apache.tomcat.util.http.Parameters;
-import org.apache.tomcat.util.http.ServerCookies;
-import org.apache.tomcat.util.net.ApplicationBufferHandler;
-import org.apache.tomcat.util.res.StringManager;
-
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.nio.charset.Charset;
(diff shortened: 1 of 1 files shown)


java/org/apache/coyote/Request.java
```diff
@@ -16,19 +16,8 @@
...
-import java.util.concurrent.atomic.AtomicBoolean;
-import java.util.concurrent.atomic.AtomicLong;
-import java.util.concurrent.atomic.AtomicReference;
-
 import jakarta.servlet.ReadListener;
 import jakarta.servlet.ServletConnection;
-
 import org.apache.tomcat.util.buf.CharsetHolder;
 import org.apache.tomcat.util.buf.MessageBytes;
 import org.apache.tomcat.util.buf.UDecoder;
```
markt-asf, line 23: This breaks the Checkstyle enforced import order rules.

java/org/apache/coyote/Request.java
```diff
@@ -386,8 +385,7 @@ public void setLocalPort(int port) {
      * Get the character encoding used for this request.
      *
      * @return The value set via {@link #setCharset(Charset)} or if no call has been made to that method try to obtain
-     *             if from the content type.
-     *
+     * if from the content type.
```
markt-asf, line 388: This breaks the standard Javadoc formatting.

java/org/apache/coyote/Request.java
```diff
@@ -853,24 +846,22 @@ public boolean isProcessing() {
...
+        } else {
+            encoding = contentType.substring(start).trim();
         }
-        encoding = encoding.trim();
-        if ((encoding.length() > 2) && (encoding.startsWith("\"")) && (encoding.endsWith("\""))) {
-            encoding = encoding.substring(1, encoding.length() - 1);
-        }
-
-        return encoding.trim();
+        return encoding.replaceAll("\"", "");
```
markt-asf, line 864: This changes functionality. Granted the existing code doesn't strictly implement the requirements of RFC 9110 either but "code clean-up" PRs should not be changing functionality without explicitly drawing attention to that fact.

Separately, I have concerns about the performance implications of this change. Srtring.replaceAll() is not cheap as it uses regular expressions.

If we want to specification compliant here (and we probably should be) the right way to do this would be to use  o.a.t.u.http.parser.MediaType.

markt-asf (changes requested):
PRs should pass the checkstyle checks included in the build.

PRs should not change formatting of unrelated code. If inconsistent formatting is an issue, the code should be formatted in a separate commit use the settings (or their equivalents) found in `res/ide-support/eclipse/clean-up-asf-tomact.xml` and `res/ide-support/eclipse/formatting-asf-tomact.xml`.

java/org/apache/coyote/Request.java
wonyongChoi05: I'm curious whether there are any performance advantages when using MediaType to comply with the specification compared to the existing code. Additionally, I'm interested in knowing whether the existing code already adhered to the specification as well as MediaType does.

java/org/apache/coyote/Request.java
```diff
@@ -852,25 +854,17 @@ public boolean isProcessing() {
      *
      * @param contentType a content type header
      */
-    private static String getCharsetFromContentType(String contentType) {
+    private static String getCharsetFromContentType(String contentType) throws IOException {
```
markt-asf, line 857: The PR won't compile because of this addition.  Look at how other code (e.g. in Response) handles the possibility of an IOException. Hint: the method signature should be unchanged from the original

markt-asf (approved):

PR #663, opened 2023-09-06
Refactoring the main method of a Tomcat class to improve readability

apptie:
Hello! 
While learning about Tomcat, I see something that I think I can contribute to, so I write this PR.

## Explanation

There are two commits here:

- Remove the FQCN when creating a Tomcat instance
- Change the order of equals method calls

### Remove the FQCN when creating a Tomcat instance

I removed the FQCN when creating the Tomcat instance, because I was creating an instance of myself and determined that there were no other Tomcat classes inside the org.apache.catalina package.

### Change the order of equals method calls

Previously, I was comparing strings based on their args(`args[i]`), and I thought I could reverse the order to make them a little more readable.

---

If I've misunderstood something, please let me know.

### DIFF
--- a/java/org/apache/catalina/startup/Tomcat.java
+++ b/java/org/apache/catalina/startup/Tomcat.java
@@ -1303,7 +1303,7 @@ public class Tomcat {
                 break;
             }
         }
-        org.apache.catalina.startup.Tomcat tomcat = new org.apache.catalina.startup.Tomcat();
+        Tomcat tomcat = new Tomcat();
         // Create a Catalina instance and let it parse the configuration files
         // It will also set a shutdown hook to stop the Server when needed
         // Use the default configuration source
--- a/java/org/apache/catalina/startup/Tomcat.java
+++ b/java/org/apache/catalina/startup/Tomcat.java
@@ -1290,9 +1290,9 @@ public class Tomcat {
         // Process some command line parameters
         String[] catalinaArguments = null;
         for (int i = 0; i < args.length; i++) {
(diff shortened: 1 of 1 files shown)


java/org/apache/catalina/startup/Tomcat.java
```diff
@@ -1290,9 +1290,9 @@ public static void main(String[] args) throws Exception {
         // Process some command line parameters
         String[] catalinaArguments = null;
         for (int i = 0; i < args.length; i++) {
-            if (args[i].equals("--no-jmx")) {
+            if ("--no-jmx".equals(args[i])) {
```
markt-asf, line 1293: The change to equals() is subjective. I'd argue the current code is easier to read.

java/org/apache/catalina/startup/Tomcat.java
```diff
@@ -1303,7 +1303,7 @@ public static void main(String[] args) throws Exception {
                 break;
             }
         }
-        org.apache.catalina.startup.Tomcat tomcat = new org.apache.catalina.startup.Tomcat();
+        Tomcat tomcat = new Tomcat();
```
markt-asf, line 1306: The removal of the FQCN is fine but as the only remaining change is rather small for the effort required to review and merge a PR. A PR that addressed this for the entire code base (I believe there is a PMD rule to check for this) would be useful.

markt-asf (changes requested):

markt-asf:
The PMD tool has been used, the use of FQCN identified and unnecessary usages have been fixed.

PR #664, opened 2023-09-06
Improved code readability

sosow0212:
While seeing the code in the Coyote project, I came across this code snippet: 
`if (!(chr == Constants.SP || chr == Constants.HT))`.

It appears to be challenging to read, so I enhanced its readability using De Morgan's law.

---
before
`if (!(chr == Constants.SP || chr == Constants.HT))`.

after
`if (chr != Constants.SP && chr != Constants.HT)`


markt-asf:
See #661 - a single PR would be preferred.

markt-asf:
Included in #661

PR #665, opened 2023-09-06
Update tool-wrapper.bat

ChrissW-R1:
Solved issue with spaces in path.

If you install tomcat on windows in a path with spaces, the tool-wrapper.bat could be executed. This PR adds two missing qoutes to solve this issue.


aooohan:
Thanks for the PR.

PR #667, opened 2023-09-08
Before closing the database connection, roll back the transaction

y987425112:
fix bug https://bz.apache.org/bugzilla/show_bug.cgi?id=64570
Before closing the database connection, roll back the transaction

 I have referred to implementation methods such as Druid and Hikari databases and believe that it is necessary to roll back uncommitted transactions before closing the connection


kkolinko:
From code review point of view: -1

1. No error handling.
2. No configuration.

The `rollbackOnReturn` option should take care of the original BZ issue. And if there is a bug, behaviour of the patch should depend on the value of that option.

See `o.a.t.jdbc.pool.ConnectionPool#terminateTransaction(...)`

markt-asf:
Closing since behaviour is intentionally configurable.

PR #669, opened 2023-09-09
Relocate the useCompression check and apply parts back to back

shin-mallang:
It can be confusing to read the code because the part that checks for HTTP compression and the part that applies the filter are separated from each other
I've placed them back-to-back to improve readability.


rmaucher:
The ordering of these is often important. Please don't try to change this for cosmetic reason since it needs time and expertise to review.

markt-asf:
Re-opening. Discussion on the dev list confirmed that this change was safe.

shin-mallang:
@markt-asf 
I apologize for not sending a PR with an explanation after carefully checking that the change in the code was safe.
I apologize for the inconvenience.
Thank you.

PR #670, opened 2023-09-18
Adding ASL Bundle-License to the manifests

coheigea:
Other ASF projects like CXF that are built using the maven-bundle-plugin have the pom license automatically inserted into the manifest as "Bundle-License: https://www.apache.org/licenses/LICENSE-2.0.txt". This helps tools like syft determine the license when building an SBOM. This PR adds this to the Tomcat jars so that Syft can automatically determine the correct license.

If approved please backport to 10.x + 9.x.

### DIFF
--- a/res/bnd/annotations-api.jar.tmp.bnd
+++ b/res/bnd/annotations-api.jar.tmp.bnd
@@ -15,6 +15,7 @@
 
 -include: build-defaults.bnd, spec-defaults.bnd
 
+Bundle-License: https://www.apache.org/licenses/LICENSE-2.0.txt
 Bundle-Name: tomcat-annotations-api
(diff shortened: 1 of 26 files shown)


res/bnd/annotations-api.jar.tmp.bnd
```diff
@@ -15,6 +15,7 @@
 
 -include: build-defaults.bnd, spec-defaults.bnd
 
+Bundle-License: https://www.apache.org/licenses/LICENSE-2.0.txt
```
ChristopherSchultz, line 18: Could all of these Bundle-License clauses simply be put into the build-detaults.bnd file or similar?

markt-asf:
Fixed by adding that header to the defaults.

PR #677, opened 2023-11-06
Added support for JDBC 4.3 beginRequest and endRequest methods.

fmeheust:
JDBC 4.3 added the methods beginRequest and endRequest to java.sql.Connection. These methods hint the driver of the beginning and completion of a request (independent unit of work). Connection pooling manager should call these methods on the underlying connection prior to returning a connection to the caller, and when the applications returns the connection back to the connection pool respectively.

This pull request adds support for beginRequest and endRequest methods to the Tomcat JDBC Connection Pool.

### DIFF
--- a/modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/ConnectionPool.java
+++ b/modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/ConnectionPool.java
@@ -143,9 +143,13 @@ public class ConnectionPool {
     /**
      * Request boundaries
      */
-    private volatile boolean requestBoundaryMethodsInitialised = false;
-    private Method beginRequest = null;
-    private Method endRequest = null;
+    private static final Method beginRequest;
+    private static final Method endRequest;
+
+    static {
+        beginRequest = getConnectionMethod("beginRequest");
+        endRequest = getConnectionMethod("endRequest");
+    }
 
     //===============================================================================
     //         PUBLIC METHODS
@@ -352,8 +356,8 @@ public class ConnectionPool {
             } else {
                 connection = (Connection)proxyClassConstructor.newInstance(new Object[] {handler});
             }
-            Connection underlyingConnection = con.getConnection();
-            if (hasRequestBoundaryMethods(underlyingConnection)) {
+            if (beginRequest != null) {
+                Connection underlyingConnection = con.getConnection();
                 try {
                     beginRequest.invoke(underlyingConnection);
                 } catch (InvocationTargetException | IllegalAccessException ex) {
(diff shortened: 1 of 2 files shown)


modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/ConnectionPool.java
```diff
@@ -812,6 +829,28 @@ protected PooledConnection createConnection(long now, PooledConnection notUsed,
...
+     */
+    private boolean hasRequestBoundaryMethods(Connection connection) {
+        if (!requestBoundaryMethodsInitialised && connection != null) {
+            try {
+                beginRequest = connection.getClass().getMethod("beginRequest");
+                endRequest = connection.getClass().getMethod("endRequest");
+            } catch (NoSuchMethodException ex) {
+                // begin and end request not implemented, ignore exception
+            } finally {
+                requestBoundaryMethodsInitialised = true;
```
aooohan, line 848: Maybe this block move to static block and assign to some static variables is better?

modules/jdbc-pool/src/test/java/org/apache/tomcat/jdbc/test/ConnectionBoundariesTest.java
```diff
@@ -0,0 +1,53 @@
...
+        ds.getPoolProperties().setUrl(Driver.url);
+        ds.getPoolProperties().setInitialSize(0);
+        ds.getPoolProperties().setMaxIdle(10);
+        ds.getPoolProperties().setMinIdle(10);
+        ds.getPoolProperties().setMaxActive(10);
+        return ds;
+    }
+    @Test
+    public void connectionWithRequestBoundariesTest() throws Exception {
+        datasource.getPoolProperties().getDbProperties().setProperty("requestBoundaries", "true");
```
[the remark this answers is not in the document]
fmeheust: I had created another Connection class that would be returned if that property was set, but finally I just added the methods to the existing Connection class, which makes that line useless. I have removed it

modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/ConnectionPool.java
fmeheust: I have changed the code and moved the code to a static block. The getMethod call is now made on the java.sql.Connection interface and not on the instance if the underlying connection, with this solution request boundaries are only called if the version of java is 9 or later. I have changed the tests accordingly.

aooohan (approved):
LGTM.

ahai-code (approved):

PR #682, opened 2024-01-06
Fix HTTP/2 support wording in HTTP docs

janbrasna:
Trivial wording fix for config docs.

_(Unfortunately goes all the way back to `9.0.x` and `8.5.x`)_


markt-asf:
Tx. I'll back-port this fix shortly.

PR #683, opened 2024-01-07
Fix websocket processor leaks when WsSession expires and server sends close message by WsBackgroundThread

lbliangbin:
Fix websocket processor leaks when WsSession expires and server sends close message by WsBackgroundThread

### DIFF
--- a/java/org/apache/tomcat/websocket/server/WsRemoteEndpointImplServer.java
+++ b/java/org/apache/tomcat/websocket/server/WsRemoteEndpointImplServer.java
@@ -103,12 +103,12 @@ public class WsRemoteEndpointImplServer extends WsRemoteEndpointImplBase {
     protected boolean acquireMessagePartInProgressSemaphore(byte opCode, long timeoutExpiry)
             throws InterruptedException {
 
-        // Only close requires special handling.
-        if (opCode != Constants.OPCODE_CLOSE) {
+        // Only close message in response requires special handling.
+        int socketWrapperLockCount = socketWrapper.getLock().getHoldCount();
+        if (opCode != Constants.OPCODE_CLOSE || socketWrapperLockCount <= 0) {
             return super.acquireMessagePartInProgressSemaphore(opCode, timeoutExpiry);
         }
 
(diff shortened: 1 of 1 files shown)


markt-asf:
No evidence provided to support claim of a leak.
No leak observed when testing with profiler.
No obvious reason that proposed patch would impact `UpgradeProcessor` instances (or anything else that might be called a WebSocket processor).

imario42:
@markt-asf : He described his findings in that tomcat-user post
[https://lists.apache.org/thread/vyog0dyxwq3hlpjgt2vb1wkrxw6x2h1k](url)
However, I do not know if his patch works.

I myself have to workaround that issue by cleaning up the AbstractProtocol.waitingProcessors running a Groovy script every 30 minutes.
Thing is, I simply use SpringBoot and I also do not see how my server code should cause that issues.
Tricky thing it seems.
How can we provide additional infos?

Bildschirmfoto 2024-01-18 um 17 03 59

markt-asf:
Thanks. The reference to the description on the users list helps. I'll look at this again.

markt-asf:
The description of the problem and the analysis of the root cause (the BZ 66508 fix) is correct but the proposed fix is not. It will probably work the majority of the time but there is a timing gap where the problem can still occur. I have an alternative but I need to do some testing before I commit the fix.

markt-asf:
Fixed and back-ported. Will be in the February releases.

imario42:
Thanks a lot @markt-asf  for tackling and fixing it and @lbliangbin for the analysis. 👍😎

chenbiaoit:
>  Fixed and back-ported. Will be in the February releases.

Hi @markt-asf,What channel can I use to know that this problem has been fixed? Thank you very much.

ChristopherSchultz:
> > Fixed and back-ported. Will be in the February releases.
> 
> Hi @markt-asf,What channel can I use to know that this problem has been fixed? Thank you very much.

Are you looking for something other than the commit log, the change log, or a release announcement? Because two of those have already happened, and the third (release announcement) will occur after the next release which contains this fix.

PR #684, opened 2024-01-13
Remove org.apache.catalina.ssi from tomcat-embed-core.jar.tmp.bnd

jjohannes:
This package no longer exits in `tomcat-embed-core.jar`. It is in the `module-info.class` of the current `10.1.18` and earlier `10.x` releases.

Running `jlink` or `jpackage` with a Module Path that contains `tomcat-embed-core-10.1.18.jar` gives the following error:

> jlink failed with: Error: Packages that are exported or open in org.apache.tomcat.embed.core are not present: [org.apache.catalina.ssi]


markt-asf:
Fixed manually in main for 11.0.x (with change log entry) and then back-ported (cherry-picked) to 10.1.x and 9.0.x.

jjohannes:
Thanks @markt-asf!

PR #685, opened 2024-01-19
Update to use JakartaExpressionLanguage osgi.contract

pnicolucci:
I opened: https://bz.apache.org/bugzilla/show_bug.cgi?id=66834 last year and upon looking at the latest M16 release of Expression Language 6.0 here: https://repo1.maven.org/maven2/org/apache/tomcat/tomcat-el-api/11.0.0-M16/ I still see the following in the MANIFEST.MF:

`Provide-Capability: osgi.contract;osgi.contract=JavaEL`

This should now be using the new `JakartaExpressionLanguage` osgi.contract.

### DIFF
--- a/res/bnd/el-api.jar.tmp.bnd
+++ b/res/bnd/el-api.jar.tmp.bnd
@@ -21,7 +21,7 @@ Export-Package: jakarta.el;version=${el.spec.version}
 
 Provide-Capability: \
     osgi.contract;\
-        osgi.contract=JavaEL;\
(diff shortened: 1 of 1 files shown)


markt-asf:
Thanks for reporting this.
Looks like most of the embedded JARs got missed but EL was updated for embedded but not the standalone JAR in BZ 66834. I have a fix I'll push shortly that fixes all of those.

markt-asf:
Broader fix applied.

PR #686, opened 2024-01-31
Add a fallback when people use Parameters class concurrently, to avoid lost params

XenoAmess:
We just found that in most cases, if people use Parameters class with multi-threads, adding such a fallback would make them not suffer from losting params.
It might cause duplicate params, but 1. that shall never happened in single-thread cases, so shall not break something.
2. in most multi-thread cases, getting duplicate params is actually acceptable, but losting params be not acceptable


markt-asf:
WONTFIX - As per section 2.3.3.4 applications are responsible for accessing the request in a thread safe manner.

PR #699, opened 2024-02-26
Remove duplicate cipher element from unimplemented set in TesterOpenSSL.java

dsoumis:


ChristopherSchultz:
@markt-asf Any chance this was a copy/paste error that should be corrected in a different way (e.g. there is actually a missing cipher)?

markt-asf:
@ChristopherSchultz I don't think so. We'd see test failures if that were the case.

PR #700, opened 2024-02-26
Attempt to protect against asynchronous thread safety violations

markt-asf:
Providing this as a PR as it isn't strictly a Tomcat bug or Tomcat's problem to solve. That said, we do see some asynchronous bug reports and this might help highlight application issues early and reduce the number of false positives.

If there are no objections, I'd like to get this merged for the March release round.

### DIFF
--- a/java/org/apache/catalina/connector/OutputBuffer.java
+++ b/java/org/apache/catalina/connector/OutputBuffer.java
@@ -328,7 +328,6 @@ public class OutputBuffer extends Writer {
                 // An IOException on a write is almost always due to
                 // the remote client aborting the request. Wrap this
                 // so that it can be handled better by the error dispatcher.
-                coyoteResponse.setErrorException(e);
                 throw new ClientAbortException(e);
             }
         }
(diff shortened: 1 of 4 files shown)


rmaucher:
Anywhere else can also be similarly synced, then. I don't think it is a good idea to insert random thread safety safeguards in addition to what the specification mandates. Time for a system property if you really want to add it.

markt-asf:
Fair point.

The complicating factor is that there are so many entry points that a non-container thread could use. A complete solution would need to, effectively, wrap the `Response`, `Writer`, `OutputStream`, `Request`, `Reader` and `InputStream`. The resulting code would be quite messy.

It should also be using `SocketWrapper.getLock()` since that is what Tomcat is using with the container threads.

I'll have a think to see if I can come up with a better approach.

Would you feel any differently about an approach that just targeted `Reponse.doWrite()` and `Request.doRead()` with an (optional?) lock using `SocketWrapper.getLock()`?

rmaucher:
I feel ok with this sort of thread safety additions [there could be a lot !] if they have a system property to enable them. Then it becomes useful as a way to detect problems, although I doubt it can actually make things work by itself.

markt-asf:
I'm closing this as I have come up with a better plan that makes slightly better uses of the existing facades to trigger some NPEs when this happens.

PR #702, opened 2024-02-28
Avoid performance issues that may arise when the array is large.

CloudWise-Lukemiao:
Code optimization 1. Avoid performance issues that may arise when the array is large. 2. Use the Arrays.copyOf method to adjust the size of the filter array


rmaucher:
I think FilterDef getAsyncSupported can be optimized. But I'm not super convinced by the rest. Where is the inflection point ?

rmaucher:
I added some optimizations, still based on the array. If you investigate something, then fully using a LinkedHashSet would likely be better than HashSet + array. Please provide actual micro benchmarks to determine when/if it becomes better than the current code.

PR #704, opened 2024-02-29
remove unnecessary code in InputBuffer.available()

carsago:
in this case, to assign ```available = 0``` within the ```if (available == 0)``` by using conditional operator 

remove unnecessary code, improving readability.


rmaucher:
Thanks, but it looks equivalent to me.

PR #707, opened 2024-03-15
Unify comma-separated-value code and optimize the implementation

ChristopherSchultz:
There was repeated code in various places in the source tree to spit simple comma-separated strings using the regular expression pattern `\s*,\s*`. This PR gathers those together into a single utility method and re-implements that method without using a regular expression which improves performance.

Given that this method is often used on request headers, the aggregate performance improvement over many requests is measurable especially for long strings, strings with many components, etc.

Implementation note: there is a subtle change to the behavior of this utility method relative to the original: this _new_ method results in the first and last strings being "trimmed" of whitespace while the _old_ implementation would have preserved leading whitespace in the first string value returned and trailing whitespace in the last string returned. Reviewing the uses of this method, I do not believe this will have any negative effects on the operation of Tomcat.

### DIFF
--- a/java/org/apache/catalina/realm/JNDIRealm.java
+++ b/java/org/apache/catalina/realm/JNDIRealm.java
@@ -963,11 +963,12 @@ public class JNDIRealm extends RealmBase {
         if (cipherSuites == null || cipherSuitesArray != null) {
             return cipherSuitesArray;
         }
-        if (this.cipherSuites.trim().isEmpty()) {
+        this.cipherSuites = this.cipherSuites.trim();
+        if (this.cipherSuites.isEmpty()) {
             containerLog.warn(sm.getString("jndiRealm.emptyCipherSuites"));
             this.cipherSuitesArray = null;
         } else {
-            this.cipherSuitesArray = StringUtils.splitCommaSeparated(cipherSuites.trim());
+            this.cipherSuitesArray = StringUtils.splitCommaSeparated(cipherSuites);
             if (containerLog.isTraceEnabled()) {
                 containerLog.trace(sm.getString("jndiRealm.cipherSuites", Arrays.toString(this.cipherSuitesArray)));
             }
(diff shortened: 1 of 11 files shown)


java/org/apache/catalina/realm/JNDIRealm.java
```diff
@@ -966,7 +967,7 @@ private String[] getCipherSuitesArray() {
             containerLog.warn(sm.getString("jndiRealm.emptyCipherSuites"));
             this.cipherSuitesArray = null;
         } else {
-            this.cipherSuitesArray = cipherSuites.trim().split("\\s*,\\s*");
+            this.cipherSuitesArray = StringUtils.splitCommaSeparated(cipherSuites.trim());
             if (containerLog.isTraceEnabled()) {
```
markt-asf, line 971: Is this trim necessary here? I know it was there before but it looks redundant.

java/org/apache/tomcat/util/buf/StringUtils.java
```diff
@@ -94,4 +94,28 @@ public static <T> void join(Iterable<T> iterable, char separator, Function<T,Str
...
+     *
+     * @return An array of String values.
+     */
+    public static String[] splitCommaSeparated(String s) {
+        if (s == null || s.length() == 0) {
+            return new String[0];
+        }
+
+        String[] splits = s.split(",");
+        for (int i=0; i<splits.length; ++i) {
```
markt-asf, line 115: The formatting police are going to complain about spaces around operators here.

markt-asf (approved):
Looks good to me.

I think you are actually fixing a bug here. Leading and trailing spaces on the input stream while unlikely could cause issues for some of these uses.

Just a few trivial suggested changes.

java/org/apache/catalina/realm/JNDIRealm.java
ChristopherSchultz: It turns out its doubly-redundant. It's "gone" but it's still there because the condition above also needs to trim the string to check for empty-string. But now we are only trimming once in this method instead of twice.

ChristopherSchultz:
Hmm... I've merged this PR but it occurs to me that maybe I should have deprecated all those methods instead of removing them.

Thoughts?

rmaucher:
Ok, so add back the protected static methods in 10.1, just in case.

ChristopherSchultz:
> Ok, so add back the protected static methods in 10.1, just in case.

Sounds good.

PR #710, opened 2024-03-22
it will drop first header if any header cause exceptio during process

foremans:


markt-asf:
Thanks for spotting this. I've applied this fix along with a simple test case.

PR #711, opened 2024-03-23
Make ConnectionHandler accessible through API from AbstractProtocol to ultimately acess RequestGroupInfo

rmannibucau:
Goal is to be able to have metrics without having to re-register them or to use reflection, even when JMX is disabled - including GraalVM case.

### DIFF
--- a/java/org/apache/coyote/AbstractProtocol.java
+++ b/java/org/apache/coyote/AbstractProtocol.java
@@ -458,7 +458,7 @@ public abstract class AbstractProtocol<S> implements ProtocolHandler, MBeanRegis
     }
 
 
-    protected Handler<S> getHandler() {
+    public Handler<S> getHandler() {
         return handler;
     }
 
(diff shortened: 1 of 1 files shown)


java/org/apache/coyote/AbstractProtocol.java
```diff
@@ -769,7 +769,7 @@ private void logPortOffset() {
 
     // ------------------------------------------- Connection handler base class
 
-    protected static class ConnectionHandler<S> implements AbstractEndpoint.Handler<S> {
+    public static class ConnectionHandler<S> implements AbstractEndpoint.Handler<S> {
```
markt-asf, line 772: Why do you need this change? Isn't the change to 'getHandler()' sufficient to expose the global request processor?

rmannibucau: Had the idea to make getGlobal returning RequestInfo directly (`public RequestGroupInfo getGlobal() { return global; }`) but didn't push it so you are right, will revert this change

PR #712, opened 2024-04-02
Changed simple synchronization to ReentrantReadWriteLock

LupusMKW:
Due to https://github.com/apache/tomcat/commit/4f33be682fda02a616baa0fd9b4965d248cfe1c1 our application does no longer work. 

The commit improved the situation with respect to concurrent mutability of the Services array. Previously, it was possible to read the array while it was being modified concurrently. 
Unfortunately, it can now happen that Tomcat runs into a deadlock if a starting application attempts to gather information from the Services array (which all our applications do).

The proposed change replaces the simple lock with a ReentrantReaderWriter lock, so that readers can read the Services array and only writers are blocked.

### DIFF
--- a/java/org/apache/catalina/core/StandardServer.java
+++ b/java/org/apache/catalina/core/StandardServer.java
@@ -35,6 +35,7 @@ import java.util.concurrent.ScheduledExecutorService;
 import java.util.concurrent.ScheduledFuture;
 import java.util.concurrent.ScheduledThreadPoolExecutor;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.ReentrantReadWriteLock;
 
 import javax.management.InstanceNotFoundException;
 import javax.management.MBeanException;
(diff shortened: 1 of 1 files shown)


ChristopherSchultz:
Thank you for this PR, though I was already working on a patch and the unit tests pass, so I'm going to commit my patch instead of yours. I will credit you for the idea in the changelog, however.

I like my patch better than yours primarily because you are placing the call to `lock.lock()` within the `try` block, and it should be placed before the `try` block instead. Also, I use two separate `Lock` objects instead of calling `ReentrantReadWriteLock.writeLock()` or `.readLock()` each time, which makes it easier to read the code to see that the correct lock is being used.

I _could_ just add commits to your PR to make it work, but the work is already done on my end so I'm gonna commit mine.

wapkch:
@ChristopherSchultz Could this fix be ported to 8.5.x? Cause the latest 8.5.100 has the same problem.

n828cl:
Tomcat 8.5 reached end-of-life in March, so there won't be any more updates. You should update to a supported level at your earliest convenience.

PR #713, opened 2024-04-07
Sessioncounter

pangxianhai:
sessionCounter change to AtomicLong. Avoid inaccurate counting in concurrent situations


markt-asf:
Is this fixing an observed issue or a theoretical one?

pangxianhai:
> Is this fixing an observed issue or a theoretical one?

observed issue. On the monitoring platform, we found that the sessionCounter is less than the expiredSessions. The expiredSessions is AtomicLong type, so We guess the sessionCounter is wrong using ++

rmaucher:
I think this field should simply be dropped in Tomcat 11, it's redundant with the others. Shouldn't it be expired + active ? For the other branches I would deprecate the field.

rmaucher:
Thanks for pointing this out. Since tracking too many redundant stats is not super useful, I decided to remove this one (it is now approximated by adding the active sessions to the expired sessions).

PR #718, opened 2024-04-12
fix CVE-2024-23672

achibear:
Tomcat 8 fixed the CVE-2024-23672 vulnerability through commit [3631adb1](https://github.com/apache/tomcat/commit/3631adb1342d8bbd8598802a12b63ad02c37d591). However, we have discovered similar unpatched code in Tomcat 7. Should we apply similar code changes to fix the vulnerability in this branch?

### DIFF
--- a/java/org/apache/tomcat/websocket/Constants.java
+++ b/java/org/apache/tomcat/websocket/Constants.java
@@ -19,7 +19,7 @@ package org.apache.tomcat.websocket;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
-
+import java.util.concurrent.TimeUnit;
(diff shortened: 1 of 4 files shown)


hazendaz:
Not speaking for team but as onlooker, Tomcat 7 is end of life since March 2021.  There are no further updates.  So it wouldn't make sense to patch it.  It wouldn't be released anyways.  Only tomcat 8.5 (note not 8.0) and above are supported with 10.0 also being end of life.  Only supported versions would get patched with issues.

See https://tomcat.apache.org/whichversion.html

aooohan:
Tomcat 7 is end of life since March 2021, so we won't merge this change, thanks.

PR #719, opened 2024-04-13
remove obsolete bean attributes

qauck:
These attributes no longer existing in corresponding org.apache.tomcat.util.net.NioEndpoint, which will cause java.lang.NoSuchMethodException at run time.


markt-asf:
Thanks for the PR. I did a broader review and found a few more attributes that could be removed too.

PR #720, opened 2024-04-17
Re-adding support for fractional seconds in access log

jose-galvez:
commit [a8575ef](https://github.com/apache/tomcat/commit/a8575efbc1de64fe58c8fd99e53a0011b59a8e86) removed support for seconds with floating point for the Extended Access Log Valve and made it only seconds. The documentation for Tomcat 10 says this should be seconds with fractions so this change brings back that behavior

### DIFF
--- a/java/org/apache/catalina/valves/AbstractAccessLogValve.java
+++ b/java/org/apache/catalina/valves/AbstractAccessLogValve.java
@@ -19,7 +19,9 @@ package org.apache.catalina.valves;
 
 import java.io.CharArrayWriter;
 import java.io.IOException;
+import java.math.RoundingMode;
 import java.net.InetAddress;
+import java.text.DecimalFormat;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Date;
@@ -1311,6 +1313,7 @@ public abstract class AbstractAccessLogValve extends ValveBase implements Access
     protected static class ElapsedTimeElement implements AccessLogElement {
         private final boolean micros;
         private final boolean millis;
+        private final DecimalFormat df;
 
         /**
          * @param micros <code>true</code>, write time in microseconds - %D
@@ -1320,6 +1323,8 @@ public abstract class AbstractAccessLogValve extends ValveBase implements Access
         public ElapsedTimeElement(boolean micros, boolean millis) {
             this.micros = micros;
             this.millis = millis;
+            df = new DecimalFormat("#.###");
+            df.setRoundingMode(RoundingMode.CEILING);
         }
 
         @Override
@@ -1330,7 +1335,7 @@ public abstract class AbstractAccessLogValve extends ValveBase implements Access
                 buf.append(Long.toString(TimeUnit.NANOSECONDS.toMillis(time)));
             } else {
                 // second
-                buf.append(Long.toString(TimeUnit.NANOSECONDS.toSeconds(time)));
+                buf.append(df.format(time / 1000000000.00));
             }
         }
     }


markt-asf:
From memory that change was to align Tomcat's access log configuration with httpd. It may well be that the correct fix here is to correct the documentation.
Separately, comparing this PR and the original code I suspect that, if we did want to merge something along these lines, that there would be performance concerns with the current proposal. As a minimum we'd need to see some performance figures for the proposed formatting.

ChristopherSchultz:
Why not use `%D` which is in milliseconds (Tomcat <10) or microseconds (Tomcat >= 10)?

jose-galvez:
> Why not use `%D` which is in milliseconds (Tomcat <10) or microseconds (Tomcat >= 10)?

I wanted to keep parity on the behavior here with previous versions, especially because in our case we have scripts that use this value for monitoring and are expecting this format. Updating them could be complicated since we would need to support both the Tomcat 9  and Tomcat 10+ format, and also some of our customers have their own scripts that would need to be updated too so it makes things complicated

ChristopherSchultz:
So.... nothing can ever change, then?

jose-galvez:
It can, but in this case the documentation states that it didn't so I was trying to bring it back to how it behaved before (which as an added bonus helps us not change our scripts). 

If the answer is that this is the correct behavior and the documentation is the one that should change it's fine, we'll figure out a way to adapt or patch it on our end. I just figured it was worth sharing since maybe this specific change was not intended

jose-galvez:
> Are you specifically talking about the ExtendedAccessLogValve's `time-taken` token? If so, you're right, that documentation did not change yet the underlying implementation did change.

Yes, sorry I didn't state this better 😞  but the change I made was specifically for the `time-taken` token to go back to the Tomcat 9 behavior and report the seconds with fractions as the documentation states

> IMHO, "fractional seconds" isn't worth the computation power to format it in the first place, then parse it later back into milliseconds or whatever. I would recommend changing your tooling to look for an embedded `.` character and changing behavior to be backward- and forward-compatible by simply using milliseconds or microseconds, whichever you prefer.

I'll talk to our team to see if they can make these changes. For now we created a patch for our application with this PR's change so we load the modified `ExtendedAccessLogValve` class in our application for Tomcat 10. Of course this is meant to be temporary since it could become a bit of a nightmare to maintain long term
 
> We should update the documentation for `ExtendedAccessLogValve` to make it clear that `time-taken` has changed along with `%T` to be whole-seconds.

This would be great, that way we can also send a notice to our customers pointing to it so they can update any tooling they have that relies on the current format

> We can continue this conversation to determine what _else_ we might do, such as continuing to support "fractional seconds" as an output format. The problem (for you) with supporting fractional-seconds is that it won't work from 10.1.0 - 10.1.23 at least, so you will have a specific set of Tomcat versions that are supported by your product and tooling.

We are fine right now because our current production versions are running Tomcat 9.x and we have the patch for 10.1.x ready for when it goes to production so if there's a new output in the near future to have the fractional seconds we could just move to that and remove our internal patch 🙂  but like I said I'll bring it up with the team to see if we can start moving to using milliseconds instead since it probably would also make calculations faster on our scripts and monitoring systems.

Thank you so much for your comments and your insight on this issue 😃

ChristopherSchultz:
Please have a look at a different PR which builds on some refactoring I just pushed: https://github.com/apache/tomcat/pull/721

I think I like mine better.

jose-galvez:
I like it!! Definitely way better than my change which now seems like a bad hack 🤣  I can close this PR in favor of yours 😃

PR #721, opened 2024-04-19
Add support for timescales with time-taken access log token.

ChristopherSchultz:
Also reintroduce support for fractional-seconds and add nanosecond timescale.


java/org/apache/catalina/valves/AbstractAccessLogValve.java
```diff
@@ -1756,10 +1776,14 @@ protected AccessLogElement createAccessLogElement(String name, char pattern) {
                 return new DateAndTimeElement(name);
             case 'T':
                 // ms for milliseconds, us for microseconds, and s for seconds
-                if ("ms".equals(name)) {
-                    return new ElapsedTimeElement(false, true);
+                if ("ns".equals(name)) {
+                    return new ElapsedTimeElement(ElapsedTimeElement.Style.MILLISECONDS);
```
markt-asf, line 1780: I think you want `NANOSECONDS` here.

java/org/apache/catalina/valves/AbstractAccessLogValve.java
```diff
@@ -1756,10 +1776,14 @@ protected AccessLogElement createAccessLogElement(String name, char pattern) {
...
                 } else if ("us".equals(name)) {
-                    return new ElapsedTimeElement(true, false);
+                    return new ElapsedTimeElement(ElapsedTimeElement.Style.MICROSECONDS);
+                } else if ("ms".equals(name)) {
+                    return new ElapsedTimeElement(ElapsedTimeElement.Style.MILLISECONDS);
+                } else if ("fs".equals(name)) {
+                    return new ElapsedTimeElement(ElapsedTimeElement.Style.SECONDS_FRACTIONAL);
```
markt-asf, line 1786: Strictly, that is fs is femto seconds which doesn't look right. Maybe sf?

markt-asf (changes requested):

java/org/apache/catalina/valves/AbstractAccessLogValve.java
ChristopherSchultz: If course I did. Thanks for spotting that.

java/org/apache/catalina/valves/AbstractAccessLogValve.java
ChristopherSchultz: Good question. I actually thought of femtoseconds and then thought "who actually knows about those?" and now I guess I have my answer.

We could use `sf`, sure... also `s.` which is a little obscure. There is no reason is has to be two-letters, either. `fracsec`?

markt-asf: It is the electronics engineer in me - capacitors have really small values.

`fracsec` works for me.

ChristopherSchultz: Done/

markt-asf (approved):

PR #724, opened 2024-04-22
Configurable HTML subtitle in Manager web application

xsimo:
Managing multiple tomcats on multiple servers, the need came to have a way of customizing the Manager HTML interface. 
This PR proposes to add a simple configurable html subtitle.
Example_of_customized_subtitle


markt-asf:
Thanks for the PR. I have applied a variation of this that uses a servlet init parameter rather than a context init parameter since a) the configuration is specific to the servlet and b) to be consistent with existing configuration.

Fixed in:
- 11.0.x for 11.0.0-M21 onwards
- 10.1.x for 10.1.25 onwards
-  9.0.x for  9.0.90 onwards

PR #731, opened 2024-05-31
Fix ide-support for idea

dsoumis:


isapir:
These lines were commented out on purpose, so that they are available if you need them but not enabled by default.

I am not sure that uncommenting them is the right thing to do.  The files should not be excluded in some environments/versions.

dsoumis:
Intellij won't compile the sources if those exclusions are not made, making the "ant ide-intellij" step useless. 
In case it's not the right thing to do, we should redefine the support for intellij and document some steps in order to be functional again.

isapir:
@dsoumis It won't compile if the Java version etc. are not compatible with that code. This is not making it useless. Perhaps a better change would be to add a comment about it in the output, or document it.

Maybe @rmaucher or @markt-asf can clarify what exactly is required in order for those directories to compile.

dsoumis:
@isapir It seemed weird to me, but even with Java 22 defined in Intellij the compilation didn't succeed. Maybe something else is missing as well.
+1 For adding a comment in the output

isapir:
@dsoumis I agree that it's weird, but my concern is that it would break other environments that work.

Let's wait for Remy/Mark/others to chime in as they are more familiar with how this works.

markt-asf:
I'm not an Idea user but I'd expect that code to compile using Java 22. Not much help I know but those packages aren't showing any errors in Eclipse with Java 22.

What is the error if you compile those with Java 22 (maybe check the compile target is 22 as well as Java 22 being used)?

dsoumis:
@markt-asf 
Compilation errors with Java22 (compile target is 22 as well as Java 22 being used):
> 
> /home/dsoumis/Programs/github/tomcat/java/org/apache/tomcat/util/net/openssl/panama/OpenSSLEngine.java:23:25
> java: package java.lang.foreign does not exist
> /home/dsoumis/Programs/github/tomcat/java/org/apache/tomcat/util/net/openssl/panama/OpenSSLEngine.java:116:41
> java: cannot find symbol
>   symbol:   class MemorySegment
>   location: class org.apache.tomcat.util.net.openssl.panama.OpenSSLEngine
> /home/dsoumis/Programs/github/tomcat/java/org/apache/tomcat/util/net/openssl/panama/OpenSSLEngine.java:121:19
> java: cannot find symbol
>   symbol:   class Arena
>   location: class org.apache.tomcat.util.net.openssl.panama.OpenSSLEngine

markt-asf:
Hmm. Exactly which version of Java 22 are you using?

dsoumis:
22.0.1

markt-asf:
That isn't it then. I think we need someone else with Idea (or who can set it up) to test this.

isapir:
@dsoumis I got it to work by setting the Project's `Language level` to "X - Experimental features"

image

isapir:
@dsoumis I pushed a patch that sets the Language Level to Experimental:
https://github.com/apache/tomcat/commit/f8eecc13b53fdca998565455d3c27e9b9a6cd8dc

Can you pull it and check if it resolves the issue for you?

dsoumis:
@isapir It doesn't work for me. After pulling your patch I am getting errors such as:
`java: java.lang.foreign.Arena is a preview API and is disabled by default.
  (use --enable-preview to enable preview APIs)`
  
I tried to resolve this by unchecking the --release option for cross compilation and add the following compiler arguments in Java Compiler settings as seen below 
Screenshot from 2024-06-03 10-59-22

However, now I am getting the following error:
`java: invalid source release 20 with --enable-preview
  (preview language features are only supported for release 22)`

My opinion is that we should not set language level to experimental, as sooner or later that will not be the case anymore.
I think the best approach would be to add a comment, as you initially proposed, to build.xml "ide-intellij" target that will guide the developer how to handle this case.

isapir:
Strange, I don't have those settings

image

But OK, I will revert the latest change.  Feel free to write a patch with the comments that you want to add.

dsoumis:
On second thoughts, I believe ide-support should strictly follow build.xml rules. 
As build.xml [excludes](https://github.com/apache/tomcat/blob/main/build.xml#L1014) openssl and panama directories from compilation, so should the IDE as well. Through reading the intellij manual, I [found out](https://www.jetbrains.com/help/idea/content-roots.html#exclude-files-folders) that the proper way to achieve that is by excluding the directories in Module Structure settings. This change allows for building with Java 17 as intended (IDE doesn't complain for missing Java 22) without compiling the excluded directories, and everything works as expected.
@isapir please verify this works for you as well

dsoumis:
@isapir  My expectation was that "ant ide-intellij" should work out of the box without requiring any adjustments.
Currently, to use an SDK version lower than 22, you must exclude `/java/org/apache/tomcat/util/net/openssl/panama` and `/java/org/apache/tomcat/util/openssl`, since the recommended build version is 17.
 Users can then adjust the settings to include these directories and use Java 22.

>Also, I left the commented lines in `res/ide-support/idea/compiler.xml` as an easy to toggle setting by the user, so I am against removing them.

 I suggest making these replacements in the` tomcat.iml` file rather than in `compiler.xml`. Keeping them in `compiler.xml` will prevent the user from reaching the compilation stage due to failing inspections.

isapir:
@dsoumis Tomcat 11 takes advantage of new technology.  The build script checks if the Java version is 22 and if so it compiles the Panama packages.

This is for a development environment, which should allow you to debug and devlop the project "out of the box", including the Panama packages. By commenting out the lines you are effectively making them not debuggable.  Developers who use Tomcat on an IDE are not regular users and it is perfectly reasonable IMO to expect that they use a new Java version.

If they run into the error then they can refer to the documentation or comments.  The lines should be kept commented out, regardless of which file (I didn't check the implication of having them in compiler.xml).  A user that wishes to exclude the packages can simply uncomment them.  It's a one time tweak that is easy to change if needed.

Excluding the files will not take advantage of advanced features.  Users who want to use older features can run Tomcat 10.1 or 9.0 rather than the main branch.

rmannibucau:
Assuming using uses the highest java version sounds okish but shouldnt preview mode be enabled if the packages are not excluded, otherwise it leads to the same output: it does not work in Idea.

markt-asf:
I'm not an Idea user so I have no strong view on this. I'd be happy with whatever consensus the Idea users reach.

I am an Eclipse user and the approach I use there (which I think the other Eclipse users are happy with) is:
- Eclipse is configured to use the same Java version as the build (Java 22 in this case)
- I aim to keep the project warning free if used with the current, latest version of Eclipse (I'm currently using the RC for the next release which triggers a couple of extra warnings but I haven't fixed them yet as that will create different warnings for users still on the current release)

There are times when I have manually tweaked the build to exclude the packages that require Java 22 as I wanted to check something I was doing was still going to be Java 17 compatible and tweaking the IDE was easier than constantly running Ant.

To be honest, configuring the Eclipse IDE from the Ant task isn't something I do that often. Keeping the project error free and the configuration of the auto-format tools is far more important to me than whether the FFM code is included in the IDE by default or not.

isapir:
Thank you for your input on this @markt-asf - it makes sense to me, and is consistent with my proposed solution.

> - Eclipse is configured to use the same Java version as the build (Java 22 in this case)

This is what we have now.  The reason that it is not working "out of the box" is that we also need to enable "Experimental" or "Preview" features as the FFM is not enabled in Java 22 by default, and should be enabled with a switch.  I added that to the project at https://github.com/apache/tomcat/commit/f8eecc13b53fdca998565455d3c27e9b9a6cd8dc but reverted it after getting feedback above.

We might want to reconsider adding back that flag, but if the user's IDE does not specify the latest Java version it breaks with an error that Experimental features can only be used with the latest Java SDK version, ATM 22.

IMO it is reasonable to expect a user that is developing on the latest version of Tomcat, i.e. the `main` branch, to be using the latest Java SDK, which will work with the Experimental features flag.  Users who use an older SDK will have to either disable that flag, or update to the latest SDK.  But it is up to the user to choose how to configure their local environment. 

> - There are times when I have manually tweaked the build to exclude the packages that require Java 22 as I wanted to check something I was doing was still going to be Java 17 compatible and tweaking the IDE was easier than constantly running Ant.

That is exactly the reason that I left the commented out lines that exclude the FFM directories in the project- making it easy to toggle the FFM directories by uncommenting or commenting back the lines.

Keeping the exclusions in `res/ide-support/idea/compiler.xml` (or `res/ide-support/idea/tomcat.iml`, though I haven't tested that) commented out, allows the user to test the project with all of its features (e.g. FFM), and still make it easy to uncomment the lines and test without FFM on older SDKs.

> As build.xml excludes openssl and panama directories from compilation, so should the IDE as well. 

@dsoumis That is not correct.  The build.xml only excludes the FFM directories for Java versions < 22.  See lines 1018 - 1033 which make use of the `use-ffm` flag at https://github.com/apache/tomcat/blob/main/build.xml#L1018

No settings will fit everyone 100% as IDEs set local environments with opinionated preferences, but the settings should be reasonable and fit most, and users can easily change the settings that do not fit their needs.  

We can improve the documentation and instructions to advise users how to configure the IDE, but disabling these directories in the IDE makes them unreachable when debugging/developing, and that's the primary reason to run Tomcat in an IDE.

dsoumis:
> IMO it is reasonable to expect a user that is developing on the latest version of Tomcat, i.e. the main branch, to be using the latest Java SDK, which will work with the Experimental features flag.

> @dsoumis That is not correct. The build.xml only excludes the FFM directories for Java versions < 22. See lines 1018 - 1033 which make use of the use-ffm flag at https://github.com/apache/tomcat/blob/main/build.xml#L1018

Since IDE support is [configured](https://github.com/apache/tomcat/blob/main/build.xml#L3830) with build.java.version, which currently is [17](https://github.com/apache/tomcat/blob/main/build.xml#L111), then should be aligned with build.xml.

> The reason that it is not working "out of the box" is that we also need to enable "Experimental" or "Preview" features as the FFM is not enabled in Java 22 by default, and should be enabled with a switch.

We can guide the developer by adding instructions to the target's comment in build.xml.

---
> Keeping the exclusions in res/ide-support/idea/compiler.xml (or res/ide-support/idea/tomcat.iml, though I haven't tested that) commented out, allows the user to test the project with all of its features (e.g. FFM), and still make it easy to uncomment the lines and test without FFM on older SDKs.

In its current state this will not work with @markt-asf proposal:
> Keeping the project error free and the configuration of the auto-format tools is far more important to me than whether the FFM code is included in the IDE by default or not.

Therefore, I propose that the exclusions be uncommented by default, and we provide guidance in the target's comment on how to include the FFM code. This would specify what needs to be commented out and the minimum SDK to use with  "Experimental" or "Preview" features enabled. 

(It's noteworthy that I haven't managed to make this work in Intellij even when using Java22 with preview features enabled, which might be a bug in IDEA and could affect other users as well. From my point of view, it's better to have something that works flawlessly out of the box than trying to figure out what is going wrong, which led to the existence of this PR :) )

markt-asf:
> Since IDE support is [configured](https://github.com/apache/tomcat/blob/main/build.xml#L3830) with build.java.version, which currently is [17](https://github.com/apache/tomcat/blob/main/build.xml#L111), then should be aligned with build.xml.

Fair point. I think that should probably be release.java.version and was missed when we split build version and release version but that probably needs to be a separate discussion.

> (It's noteworthy that I haven't managed to make this work in Intellij even when using Java22 with preview features enabled, which might be a bug in IDEA and could affect other users as well. From my point of view, it's better to have something that works flawlessly out of the box than trying to figure out what is going wrong, which led to the existence of this PR :) )

Personally, I'd always rather fix the root cause than the symptom. Since I'm not an Idea user I don't think my views should hold much weight. That said, what I'd suggest is:
- do nothing for now - folks can always tweak their own configuration if they want/need to
- figure out the root cause of why this isn't working for you
- discuss the next steps once we know what the root cause is

dsoumis:
The change we could make so that all opinions are generally satisfied are:

1. Move the exclusions from `res/ide-support/idea/compiler.xml` to `res/ide-support/idea/tomcat.iml` commented out.
2. Provide guidance through target's comment

Something like:
```
<echo>IntelliJ IDEA project directory created.
The SDK was set to &quot;${build.java.version}&quot; so make sure that your IDE has an SDK with that name,
or update the Project Settings accordingly. Consider uncommenting the exclusion lines in ./idea/tomcat.iml or use an SDK >= 22 with "Experimental" / "Preview" features enabled.</echo>
```

ChristopherSchultz:
My preference is for the ide-idea ant target to work for *most people*. _Most people_ are not Tomcat developers, but people who have downloaded our sources and are building them locally for whatever reason. I would want this to break in a way that is more easily-solved by most of these users.

Tomcat devs can ask their colleagues how to get around a problem that is specific to them.

PR #732, opened 2024-06-04
Fix typo in Czech translation

jajik:
At the same time, I've updated the translation in the POEditor.


jajik:
Closing after a discussion with @rmaucher , POEditor changes should be sufficient.

PR #733, opened 2024-06-05
Fix property ant.tstamp.now.iso ignored when building Tomcat JDBC poo…

dsoumis:
…l libraries

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -1598,7 +1598,7 @@
     <echo message="Building Tomcat JDBC pool libraries"/>
     <ant antfile="${tomcat.jdbc.dir}/build.xml" dir="${tomcat.jdbc.dir}"
          inheritAll="false" target="build">
-      <property name="ant.tstamp.now.iso" value ="${ant.tstamp.now.iso}" />
+      <property name="ant.tstamp.now.iso" value ="${tstamp.iso.release}" />
       <property name="tstamp.file" value ="${tstamp.file}" />
       <property name="tomcat.pool" value="${tomcat.pool}" />
       <property name="tomcat.juli.jar" value="${tomcat-juli.jar}" />


ChristopherSchultz (commented):
For an actual release, `tstamp.iso.release` and `ant.tstamp.now.iso` will be set to the same value. I'm not sure this change is necessary.

Under what conditions will `tstamp.iso.release` and `ant.tstamp.now.iso` have different values? Not during a "real release" (`ant pre-release` followed by `ant release` fixes the timestamp at the time of `pre-release`) and not doing a "non-real release" (`ant release` without `ant pre-release` fixes the timestamp at the time of `release`).

dsoumis:
When running `ant deploy` to reproduce the build, `pre-release` target will not have been executed, ant.tstamp.now.iso has no value assigned and the following output is written:

> build-tomcat-jdbc:
     [echo] Building Tomcat JDBC pool libraries
   [tstamp] magic property ant.tstamp.now.iso ignored as '${ant.tstamp.now.iso}' is not in valid ISO pattern
   [tstamp] magic property ant.tstamp.now.iso ignored as '${ant.tstamp.now.iso}' is not in valid ISO pattern


   
   May not be an actual issue, but it disrupts the flawlessness of the process.

isapir:
Is `tstamp.iso.release` defined at that spot @dsoumis ?  If so, maybe we can set a default for `ant.tstamp.now.iso` to be `tstamp.iso.release`?

dsoumis:
> Is tstamp.iso.release defined at that spot @dsoumis ?
Yes.

ChristopherSchultz:
> When running `ant deploy` to reproduce the build, `pre-release` target will not have been executed

If you are trying to reproduce the build, they you will have `build.properties.release` as generated by the release manager. That file contains (e.g.):

    ant.tstamp.now.iso=2024-06-11T20:57:36Z

When running "deploy", I see this:

```
Setting project property: ant.tstamp.now.iso -> 2024-06-11T20:57:36Z
Setting project property: ant.tstamp.now.iso -> 2024-06-11T20:57:36Z
Override ignored for property "ant.tstamp.now.iso"
Setting project property: ant.tstamp.now.iso -> 2024-06-11T20:57:36Z
Override ignored for property "ant.tstamp.now.iso"
... [many times repeated]
```

However, when *not* using a source-release package:
```
$ ant -d deploy | grep ant.tstamp.now.iso
Property "ant.tstamp.now.iso" has not been set
Setting ro project property: ant.tstamp.now.iso -> ${ant.tstamp.now.iso}
   [tstamp] magic property ant.tstamp.now.iso ignored as '${ant.tstamp.now.iso}' is not in valid ISO pattern
   [tstamp] magic property ant.tstamp.now.iso ignored as '${ant.tstamp.now.iso}' is not in valid ISO pattern
```

> May not be an actual issue, but it disrupts the flawlessness of the process.

+1

The irony of the situation is that `ant.tstamp.now.iso` can be used to "set" the build timestamp for ant but ant does not set that property itself if it's not specified e.g. on the command-line. So it's basically never a good idea to use `${ant.tstamp.now.iso}` in your own build file.

ChristopherSchultz (approved):

PR #734, opened 2024-06-14
Rename method `getUseVirtualThreads` to `isUseVirtualThreads`

quaff:


markt-asf:
No prior discussion, no explanation. Therefore close without merging.

quaff:
> no explanation.
I thought it's common sense that boolean property getter should be `is*`.

markt-asf:
No, it is a coding style choice. The Tomcat code uses both. Currently about 70% use `isXxx` and 30% use `getXxx`. Personally, I tend to use whichever one feels more natural at the time. `getXxx` feels more natural to me for this case.

While users of the standard distribution won't notice this change (`IntrospectionUtils` will take care of it), it will break backwards compatibility for programmatic users. Any change like this needs to retain the old version with a deprecation annotation for at least the current major version before removal.

There needs to be a good reason for changing the API and breaking backwards compatibility. I don't see any such reason here.

ChristopherSchultz:
The Javabean specification says that boolean properties must have accessors (for readable properties) and mutators (for writable ones) called `boolean getFoo()` and/or `boolean getFoo()` and `void setFoo(boolean b)`. There is no requirement for `isFoo` versus `getFoo`. Ar Mark says, it's entirely a matter of style.

Some examples including references:
https://stackoverflow.com/a/39307501/276232

In English "get use-virtual-threads" is more readable and natural to me than "is use-virtual-threads". The best method name would probably be "should use-virtual-threads", but that breaks all accessor naming conventions, and therefore we wouldn't choose it.

PR #735, opened 2024-06-20
Refactor setLocale reduce code depth

lee-jeong-geun:
Hello
I refactored the setLocale method and the CharsetMapper's getCharset method it calls.


ChristopherSchultz:
Meh. I don't see this as a significant improvement. In at least one case it's slightly less optimized though the CPU probably won't notice the difference. It also adds multiple returns from a single function which *sometimes* makes the code more difficult to follow.

markt-asf:
Agree with Chris. At best this is just a change in coding style, at worst it is marginally less optimal.

PR #736, opened 2024-06-23
Increase test coverage of WsServerContainer

dsoumis:


rmaucher:
Did you add "execute.validate=true" to your build.properties file ? Normally we don't use wildcard imports (please avoid that part).
Other than that, more test coverage = good.

dsoumis:
Thanks for pointing that out. I've applied the fixes required.

JanFojtik21:
okey, then all good

rmaucher:
I'll merge this one.

However, it would be good to avoid adding pure spec tests in the future, as they are now supposed to be handled by the now-more-useable TCK (the TCK is to test all of the spec's assertions, so hopefully it would test most of these). OTOH, it is good to show reasonable coverage, so unless it is possible to add a TCK run into the coverage report [to see where we actually really stand], adding more tests to the testsuite still makes sense. I would say adding some basic spec tests to our testsuite is going to remain a good idea for a little while.

More stuff to think about basically ...

PR #739, opened 2024-07-05
Merged two loops into one to simplify the logic

lee-jeong-geun:
Hello.
Reduced the filter addition logic in createFilterChain to a single loop.


n828cl:
The proposed change would violate the servlet spec, which requires that filters matching URL patterns must precede those matching servlet names, even when using the same filter for both. See section 6.2.4 of the current spec:
https://jakarta.ee/specifications/servlet/6.1/jakarta-servlet-spec-6.1

markt-asf:
This won't be merged for the reason Chuck explained.

PR #740, opened 2024-07-10
Add OpenSSL FFM code

rmaucher:
Also port the TLS testsuite improvements from 10.1.
APR TLS testing uses only the OpenSSL parameter (and skips JSSE and OpenSSL FFM) to avoid looking weird while running only once.


rmaucher:
The rationale behind this is long term support of Tomcat 9.0.
- FFM support is likely desirable starting with the Java 25 LTS release
- Java 25 cannot be used to compile Tomcat 9.0
- Testsuite improvements for TLS from 10.1+ are needed in the long run

Ant javac task provides a solution to these issues by specifying a javac executable, which will then be used to compile the FFM related classes. Requiring a FFM enabled javac makes building releases more complicated but works and allows moving forward.
A new ${java-ffm.home} property needs to be specified and needs to point to a JDK 22+ with javac. The javac version used will be logged in the release properties so that the build can be reproduced (although it will obviously be more error prone than before).

PR #742, opened 2024-07-18
Correctly Proxy Statement returned from ResultSet

mc1arke:
The tomcat-jdbc connection pool creates reflective proxies around Connections, Statements, and ResultSets to allow interception of 'close' calls, and consistency of traversing the link back from a child element to its parent - e.g. Statement to Connection. However, the link from ResultSet to Statement does not intercept the call so a non-proxied Statement is returned, as well as the `equals` method not handling the comparison against a non-proxied class correctly. To overcome this, an invocation handler is being created for wrapping a ResultSet returned from any of the relevant Statement methods, with that invocation handler intercepting the `getStatement` call and returning the proxied creating statement. The `equals` checks have also been altered to handle non-proxied instances being passed in to them.

https://bz.apache.org/bugzilla/show_bug.cgi?id=69206

PR #744, opened 2024-08-08
Unwrap InvocationTargetException from underlying statement execution

mc1arke:
When executing a wrapped statement, any resulting SQLException is wrapped in an InvocationTargetException by the reflective invocation, which is rethrown by the InvocationHandler as an
UndeclaredThrowableException. To ensure the expected SQLException is thrown, the invocation of the wrapped object is being moved into the existing try-catch block, thereby ensuring the InvocationTargetException is unwrapped and the underlying exception thrown by the handler.

https://bz.apache.org/bugzilla/show_bug.cgi?id=69255

### DIFF
--- a/modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/StatementFacade.java
+++ b/modules/jdbc-pool/src/main/java/org/apache/tomcat/jdbc/pool/StatementFacade.java
@@ -117,21 +117,21 @@ public class StatementFacade extends AbstractCreateStatementInterceptor {
                 throw new SQLException("Statement closed.");
             }
 
-            if (compare(GET_RESULTSET, method)) {
-                return getConstructor(RESULTSET_IDX, ResultSet.class)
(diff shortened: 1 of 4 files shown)


markt-asf:
There will be one in the [snapshot repo](https://repository.apache.org/content/repositories/snapshots/org/apache/tomcat/tomcat/11.0.0-M25-SNAPSHOT/) in about 40 mins once the CI system finishes testing it. (There is a snapshot build after ~every commit).

VRBogdanov:
Of course, thanks a lot!

Furytron:
I just wanted to point out that we had a similar issue to [VRBogdanov](https://github.com/apache/tomcat/pull/742#issuecomment-2275626750)  where upgrading to 9.0.93 from 9.0.91, starting throwing `java.lang.reflect.UndeclaredThrowableException` on application start-up. 

Upgrading to the 9.0.94-dev (Snapshot) release confirmed that the changes here fixed the issue. So we will wait for 9.0.94 to be released before upgrading. Thanks all.

PR #749, opened 2024-09-09
Update Http11Processor.java

Oelje:
To avoid duplicate headers in the response.


Oelje:
We are using a reverse proxy on TomCat. This proxy passes the responses from the backend 1:1, including the TE header.  This wasn't a problem for a long time because the upstream firewall had ignored the double header, but with the current version of the software it acknowledged this with error messages.  In the meantime, we have also found articles on the Internet that say that a (reverse) proxy should not pass this header on. We are now trying to stop sending the header ourselves.

markt-asf:
Reverse proxies are a lot more complicated than simply echoing bytes from input to output as you are discovering.

Closing this PR as this is an application issue.

PR #750, opened 2024-09-11
Resize elements array and update index in SynchronizedStack#setLimit when needed

Chenjp:
SynchronizedStack#setLimit: when set a new limit which is less than current capacity, shrink elements array and update index are expected.

pls see TestSynchronizedStack#testResetLowerLimit()

### DIFF
--- a/java/org/apache/tomcat/util/collections/SynchronizedStack.java
+++ b/java/org/apache/tomcat/util/collections/SynchronizedStack.java
@@ -19,7 +19,7 @@ package org.apache.tomcat.util.collections;
 /**
  * This is intended as a (mostly) GC-free alternative to
  * {@link java.util.Stack} when the requirement is to create a pool of re-usable
- * objects with no requirement to shrink the pool. The aim is to provide the
+ * objects. The aim is to provide the
  * bare minimum of required functionality as quickly as possible with minimum
  * garbage.
  *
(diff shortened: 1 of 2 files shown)


markt-asf:
This PR will not be applied.

`SynchronizedStack` is deliberately designed not to support shrinking the stack because:
- Tomcat doesn't need the capability to shrink the stack
- shrinking the stack creates garbage and this class is intended to avoid unnecessary garbage

Chenjp:
> This PR will not be applied.
> 
> `SynchronizedStack` is deliberately designed not to support shrinking the stack because:
> 
>     * Tomcat doesn't need the capability to shrink the stack
> 
>     * shrinking the stack creates garbage and this class is intended to avoid unnecessary garbage

Then method #setLimit should not be public:

```java
public class Http2Protocol implements UpgradeProtocol {
    @Override
    public void setHttp11Protocol(AbstractHttp11Protocol<?> http11Protocol) {
        this.http11Protocol = http11Protocol;
        recycledRequestsAndResponses.setLimit(http11Protocol.getMaxConnections());
```
@markt-asf What would happen if we reconfigure a  stack(with 200 elements) limitation from -1 to 100?

Chenjp:
@markt-asf Would you please review detail of ```Http2Protocol#setHttp11Protocol```? reconfigure stack limitation occurs.

markt-asf:
`maxConnections` is rarely changed at runtime. When it is changed, it is more likely to be increased than decreased. In the extremely rare cases it is decreased, the stack will remain larger than strictly necessary. And that is acceptable.

PR #751, opened 2024-09-11
fix bz #69316 FastHttpDateFormat - getCurrentDate() returns inaccurate result

Chenjp:
FastHttpDateFormat#getCurrentDate() returns same string for two different second datetime.
For following date:

1st:1726041009036, Wed, 11 Sep 2024 07:50:09 GMT
2nd:1726041009940, Wed, 11 Sep 2024 07:50:09 GMT

1st:1726041011501, Wed, 11 Sep 2024 07:50:11 GMT
***2nd:1726041012101, Wed, 11 Sep 2024 07:50:11 GMT*** --it is incorrect.

Fix: check whether cached date and current date are in same second.

### DIFF
--- a/java/org/apache/tomcat/util/http/FastHttpDateFormat.java
+++ b/java/org/apache/tomcat/util/http/FastHttpDateFormat.java
@@ -71,9 +71,9 @@ public final class FastHttpDateFormat {
     }
 
     /**
-     * Instant on which the currentDate object was generated.
+     * Instant on which the currentDate object was generated. 
      */
-    private static volatile long currentDateGenerated = 0L;
+    private static volatile long currentDateGeneratedInSeconds = 0L;
 
 
     /**
@@ -103,12 +103,11 @@ public final class FastHttpDateFormat {
      * @return the HTTP date
      */
     public static String getCurrentDate() {
-        long now = System.currentTimeMillis();
-        // Ignore millisecond part.
-        now -= now % 1000L;
-        if (now != currentDateGenerated) {
-            currentDate = FORMAT_RFC5322.format(new Date(now));
-            currentDateGenerated = now;
+        // according rfc5322, date/time data is accurate to the second.
+        long nowInSeconds = System.currentTimeMillis() / 1000L;
+        if (nowInSeconds != currentDateGeneratedInSeconds) {
+            currentDate = FORMAT_RFC5322.format(new Date(nowInSeconds * 1000L));
+            currentDateGeneratedInSeconds = nowInSeconds;
         }
         return currentDate;
     }
--- a/java/org/apache/tomcat/util/http/FastHttpDateFormat.java
+++ b/java/org/apache/tomcat/util/http/FastHttpDateFormat.java
@@ -71,7 +71,7 @@ public final class FastHttpDateFormat {
     }
 
     /**
-     * Instant on which the currentDate object was generated. 
+     * Instant on which the currentDate object was generated.
      */
     private static volatile long currentDateGeneratedInSeconds = 0L;
 
--- a/java/org/apache/tomcat/util/http/FastHttpDateFormat.java
+++ b/java/org/apache/tomcat/util/http/FastHttpDateFormat.java
@@ -104,8 +104,10 @@ public final class FastHttpDateFormat {
      */
     public static String getCurrentDate() {
         long now = System.currentTimeMillis();
+        // Ignore millisecond part.
+        now -= now % 1000L;
         // Handle case where time moves backwards (e.g. system time corrected)
-        if (now - now%1000L != currentDateGenerated - currentDateGenerated%1000L) {
+        if (now != currentDateGenerated) {
             currentDate = FORMAT_RFC5322.format(new Date(now));
             currentDateGenerated = now;
         }
--- a/java/org/apache/tomcat/util/http/FastHttpDateFormat.java
+++ b/java/org/apache/tomcat/util/http/FastHttpDateFormat.java
@@ -106,7 +106,6 @@ public final class FastHttpDateFormat {
         long now = System.currentTimeMillis();
         // Ignore millisecond part.
         now -= now % 1000L;
-        // Handle case where time moves backwards (e.g. system time corrected)
         if (now != currentDateGenerated) {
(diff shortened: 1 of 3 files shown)


java/org/apache/tomcat/util/http/FastHttpDateFormat.java
```diff
@@ -105,7 +105,7 @@ public final class FastHttpDateFormat {
     public static String getCurrentDate() {
         long now = System.currentTimeMillis();
         // Handle case where time moves backwards (e.g. system time corrected)
-        if (Math.abs(now - currentDateGenerated) > 1000) {
+        if (now - now%1000L != currentDateGenerated - currentDateGenerated%1000L) {
```
markt-asf, line 108: There are faster ways of achieving the same results. This adds ~ 9% to getCurrentDate(). Look as dividing now by 1000 instead.

test/org/apache/tomcat/util/http/TesterFastHttpDateFormat.java
```diff
@@ -0,0 +1,48 @@
+package org.apache.tomcat.util.http;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+public class TesterFastHttpDateFormat {
```
markt-asf, line 6: This needs to be called `TestFastHttpDateFormat` so it gets included in standard test runs.

markt-asf (changes requested):
In addition to the specific comments on the code there is the more general question of whether we want to fix this issue at all. HTimestamps might be out by up to 1s (plus the system clock skew) with the current code. How much of a problem does that cause?

There is a performance cost of fixing this. It is only ~1ns per request so almost certainly not an issue but I think we need to explicitly make any such decisions else we'll keep doing this and lots of little performance impacts will add up over time.

java/org/apache/tomcat/util/http/FastHttpDateFormat.java
rainerjung: Rough idea: add "current" variables currentDateGeneratedLower and currentDateGeneratedSecsUpper
Only set them when we generate a new currentDate, and then use:
currentDateGeneratedLower = currentDateGenerated - currentDateGenerated % 1000
currentDateGeneratedUpper = currentDateGeneratedLower + 1000
Finally change the new comparison expression to
if (now < currentDateGeneratedLower || now >= currentDateGeneratedUpper)
which involved only two comparisons in the hot path.
Untested, but I could try and provide a formal patch.

Chenjp: to minimize costs , ignore ms part to avoid divide ops:
```java
    public static String getCurrentDate() {
        long now = System.currentTimeMillis();
        // Ignore millisecond part.
        now -= now % 1000L;
        // Handle case where time moves backwards (e.g. system time corrected)
        if (now != currentDateGenerated) {
            currentDate = FORMAT_RFC5322.format(new Date(now));
            currentDateGenerated = now;
        }
        return currentDate;
    }
```

isapir: Modulus is also a slow operation.  If there is a chance of this slowing every request by 1ms as @markt-asf suggested then I'm a `-1` on this.  A `1ms` slowdown means 1,000 less requests-per-second in throughput.

Any unnecessary change should be benchmarked to evaluate the overhead that it adds.

Chenjp: @isapir On my laptop pc, new implementation elapse ***3.32 nanoseconds / per call*** (single thread), comparing to original implementation's ***3.36 ns / per call***, even performance improved.

Please try it.

isapir: @Chenjp If it's a few nanoseconds then that's fine by me.  How did you measure that?  Using JMH?

Chenjp: TesterFastHttpDateFormatPerformance, change threadCount to 1.
result: Duration: 331058600 (***total callCount = 100000000***)
per call: 331058600 ns/ 100000000 (callCount) = ***3.31 ns***;

java/org/apache/tomcat/util/http/FastHttpDateFormat.java
```diff
@@ -104,8 +104,9 @@ public final class FastHttpDateFormat {
      */
     public static String getCurrentDate() {
         long now = System.currentTimeMillis();
-        // Handle case where time moves backwards (e.g. system time corrected)
-        if (Math.abs(now - currentDateGenerated) > 1000) {
+        // Ignore millisecond part.
+        now -= now % 1000L;
```
markt-asf, line 108: Is `long now = System.currentTimeMillis() / 1000L;` not faster?

Chenjp: Without significant improvement .
1. solution1: Duration: ***322245800ns*** (100000000 times)
```java
    public static String getCurrentDate() {
        long now = System.currentTimeMillis() / 1000L;
        if (now != currentDateGenerated) {
            currentDate = FORMAT_RFC5322.format(new Date(now*1000L));
            currentDateGenerated = now;
        }
        return currentDate;
    }
```
2. solution2: Duration: ***328164000ns*** (100000000 times)
```java
    public static String getCurrentDate() {
        long now = System.currentTimeMillis();
        // Ignore millisecond part.
        now -= now % 1000L;
        if (now != currentDateGenerated) {
            currentDate = FORMAT_RFC5322.format(new Date(now));
            currentDateGenerated = now;
        }
        return currentDate;
    }
```

Chenjp: @markt-asf Updated, PTAL

markt-asf:
The code looks good to me but I'd still like to see an answer to Rémy's question above before we decide whether to merge this PR.

rmaucher:
It it doesn't much perform worse, I'm fine with it.

markt-asf:
OK. I'll merge this then fix the missing ALv2 header and any other issues before porting it to the other versions.

markt-asf (approved):

PR #752, opened 2024-09-12
Tests the compatibility with rewrite from httpd for non existent headers

PaulLodge:
This test adds code coverage for compatibility with rewrite from httpd for non existent headers


n828cl:
The new method testRewriteEmptyHeader() does not appear to be invoked from anywhere. Is some of the change missing?

PR #753, opened 2024-09-12
Add ParameterLimitValve to enforce request parameter limits for specific URLs

dsoumis:
This is an effort of introducing Parameter Limit Valve to allow limiting the number of parameters in HTTP requests, but explicitly allowing more parameters for specific URLs.

It's worth to be noted that if the Parameter Limit Valve is configured, it operates independently of the Connector's maxParameterCount attribute. The Connector's maxParameterCount sets a global limit, while the Parameter Limit Valve offers additional flexibility by allowing different limits for specific URLs. However, if the maxParameterCount defined in the Connector is lower, it effectively overrides the valve by preventing large requests from ever reaching it.

### DIFF
--- a/java/org/apache/catalina/valves/ParameterLimitValve.java
+++ b/java/org/apache/catalina/valves/ParameterLimitValve.java
@@ -1,0 +1,153 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.catalina.valves;
+
+
+import java.io.IOException;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+import jakarta.servlet.ServletException;
+
+import org.apache.catalina.connector.Request;
+import org.apache.catalina.connector.Response;
+
+
+/**
+ * This is a concrete implementation of {@link ValveBase} that enforces a limit on the number of HTTP request parameters
+ * allowed in POST requests. The features of this implementation include:
+ * <ul>
+ * <li>Global parameter limit that applies to all requests</li>
+ * <li>URL-specific parameter limits that can be defined using regular expressions</li>
+ * <li>Configurable through Tomcat's <code>server.xml</code> or <code>context.xml</code></li>
+ * </ul>
+ * <p>
+ * The global limit, specified by <code>maxGlobalParams</code>, applies to all requests unless a more specific
+ * URL pattern is matched. URL patterns and their corresponding limits can be configured via a regular expression
+ * mapping through the <code>urlPatternLimits</code> attribute.
+ * </p>
+ * <p>
+ * The Valve checks each incoming request and enforces the appropriate limit. If a request exceeds the allowed number
+ * of parameters, it returns a <code>400 Bad Request</code> response with a descriptive error message.
+ * </p>
+ * <p>
+ * Example configuration in <code>context.xml</code>:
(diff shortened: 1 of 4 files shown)


markt-asf:
This approach could be implemented as a Filter. If we were going to do this, I think something that enforces the limit at the point the parameters are parsed - rather than after - is the way to go. It would mean some refactoring of `Request.doParseParameters()` and I haven't thought it all the way through but it looks doable. Maybe add a maxParameterCount field to the request that is reset from the current connector value on recycle but the Valve could then change it. The added bonus is that if the app doesn't trigger parameter parsing, the parameters won't get parsed.

ChristopherSchultz:
While I haven't actually looked at the code (for either `request.parseParameters` or this Valve), I thought the point of `maxParameterCount` was that the request would actually fail to be fully-parsed and recovery was possible neither via Valve nor Filter. The parameter count limit is there to protect Tomcat from a DoS caused by hash collisions (right?).

Allowing a Valve/Filter to recover from this means that that DoS protection is effectively gone, even if the request is ultimately rejected with a 4xx response. The damage has already been done.

dsoumis:
I will refactor the implementation as follows:
1. The maxParameterCount limit will be enforced directly within Request.doParseParameters(). If the Valve is set and the number of parameters exceeds the configured limit, the request will fail immediately, preventing parameter parsing operations from taking place.
2. The ParameterLimitValve will only adjust the maxParameterCount on the Request object before parameter parsing occurs. The actual enforcement of the limit will remain in the parsing stage, ensuring that no recovery happens post-parsing, and thus preserving the intended DoS protection.

Please let me know if this sounds like a reasonable solution, or if there’s anything else to consider that I am missing.

markt-asf:
> The parameter count limit is there to protect Tomcat from a DoS caused by hash collisions (right?).

Hash collisions was why the 10k limit was put in place - [CVE-2012-0022](http://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2012-0022). That was chosen as the largest round number that was low enough to avoid the hash collision issue.

The further reduction to 1k was on the basis that very few apps need the limit that high and there is a memory cost to handling parameters. The aim was to reduce the minimum amount of RAM Tomcat needed to have and still be able to handle default maximum concurrent requests with default maximum parameters each.

Like all the Tomcat defaults, it is a trade-off.

dsoumis:
I have refactored the valve and the request as discussed here.
There were 2 failing tests after the change to Request's constructor which introduced method accessing of Connector. 
The change at TestPersistentManager.java is just a reordering of when the mock Connector object is activated.
Regarding TestSSLValve.java, I introduced a singleton implementation of Mockrequest giving more flexibility for Connector values.

test/org/apache/catalina/valves/TestSSLValve.java
```diff
@@ -314,12 +352,14 @@ public void testSslCipherUserKeySizeHeaderPresent() throws Exception {
...
     @Test(expected = NumberFormatException.class)
     public void testSslCipherUserKeySizeHeaderBadFormat() throws Exception {
+        setUp();
         mockRequest.setHeader(valve.getSslCipherUserKeySizeHeader(), "not-an-integer");
 
         try {
             valve.invoke(mockRequest, null);
         } catch (NumberFormatException e) {
             Assert.assertNull(mockRequest.getAttribute(Globals.KEY_SIZE_ATTR));
+            mockRequest.setHeader(valve.getSslCipherUserKeySizeHeader(), null);
```
dsoumis, line 362: Due to the random ordering of the execution of the tests, I had to reset the header at this point because other tests were throwing NumberFormatException.

java/org/apache/catalina/valves/ParameterLimitValve.java
```diff
@@ -0,0 +1,119 @@
...
+     *
+     * @param request  The servlet request to be processed
+     * @param response The servlet response to be created
+     *
+     * @exception IOException      if an input/output error occurs
+     * @exception ServletException if a servlet error occurs
+     */
+    @Override
+    public void invoke(Request request, Response response) throws IOException, ServletException {
+        String requestURI = request.getRequestURI();
```
rmaucher, line 103: This returns the original URI as sent by the client (= not decoded). If some mapping is done, then `request.getDecodedRequestURI()` must be used instead.
Maybe the best place to look as an example would be the rewrite valve which uses:
```
            MessageBytes urlMB = context ? request.getRequestPathMB() : request.getDecodedRequestURIMB();
            urlMB.toChars();
            CharSequence urlDecoded = urlMB.getCharChunk();
```
The `context` flag is there to allow a different matching behavior if the valve is associated with a context (usually when this happens the match is expected to be relative to the context root instead of the full path).

If doing all that, the XML configuration should allow some kind of encoding (probably URL encoding), and the patterns should also be URL decoded when they are parsed.

PR #755, opened 2024-09-19
Fix: evict cached MBean when bean descriptor content changed to ensure ManagedBean#getMBeanInfo result is correct.

Chenjp:
1. ensure evict cache for each modification like returnType, name, desc, addAttribute etc. 
2. code refactoring.
3. testcase added.

### DIFF
--- a/java/org/apache/tomcat/util/modeler/AttributeInfo.java
+++ b/java/org/apache/tomcat/util/modeler/AttributeInfo.java
@@ -120,19 +120,10 @@ public class AttributeInfo extends FeatureInfo {
 
     // --------------------------------------------------------- Public Methods
 
-
-    /**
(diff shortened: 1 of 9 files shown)


markt-asf:
Why would MBean info change at runtime? The MBeans pretty much all represent Tomcat internals and they don't change structure at runtime. I'm not seeing the use case for this.

rmaucher:
In theory this is an upstream issue (commons-modeler), however it is no longer maintained.

Chenjp:
> In theory this is an upstream issue (commons-modeler), however it is no longer maintained.

got it.

markt-asf:
Unless a valid use case is provided for this change, this PR will be closed without being applied.

Chenjp:
No valid use case.

PR #756, opened 2024-09-24
Unit test ensuring that the Expires and Date headers use GMT

PaulLodge:
This Unit test makes sure that the expires and date headers are consistent and both use GMT, as per the bug report https://bz.apache.org/bugzilla/show_bug.cgi?id=62476


test/org/apache/catalina/authenticator/TestFormAuthenticatorA.java
```diff
@@ -216,6 +221,38 @@ public void testSelectedMethods() throws Exception {
     }
 
 
+    /*
+     * Test to ensure that the Expir and
```
dsoumis, line 225: Comment seems incomplete here :)

PR #762, opened 2024-10-10
Report HTTP/2 header parsing errors earlier

LQxdu:
We found that the 10.0.x branch is also affected by CVE-2024-24549(https://github.com/advisories/GHSA-7w75-32cg-r6g2), but it has not been fixed yet. Therefore, we have created this patch for 10.0.x based on the fixes applied in other branches.

### DIFF
--- a/java/org/apache/coyote/http2/Http2Parser.java
+++ b/java/org/apache/coyote/http2/Http2Parser.java
@@ -289,6 +289,9 @@ class Http2Parser {
 
         swallowPayload(streamId, FrameType.HEADERS.getId(), padLength, true, buffer);
 
+        // Validate the headers so far
+        hpackDecoder.getHeaderEmitter().validateHeaders();
+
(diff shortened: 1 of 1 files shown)


isapir:
Is there a reason you are using Tomcat 10.0 and not Tomcat 10.1?  Tomcat 10.0 is no longer supported

See https://tomcat.apache.org/tomcat-10.0-eol.html

LQxdu:
During a recent network asset mapping assessment, we identified that a significant number of services are still deployed on Tomcat 10.0.x. Given the continued widespread use of this version, we think it is valuable to provide patch support for these legacy deployments as well.

n828cl:
As previously stated, the 10.0.x branch is well past end-of-life (over three years ago); there will be no further updates to this level. Anyone not updating a product in that length of time clearly has no interest in security, regardless of whatever lip service they might pay.

PR #763, opened 2024-10-10
Test if a custom 404 page is configured but the error jsp redirects to the project root

PaulLodge:
This test covers the regression mentioned in https://svn.apache.org/r1830547


rmaucher:
Thanks. I rewrote this to avoid using files and JSP, but with the same general idea.

PR #764, opened 2024-10-10
Allow applications to trigger sending of 103 early hints

ChristopherSchultz:


ChristopherSchultz:
The Servlet 6.2 spec doesn't actually exist yet... at least not in any published form.

That reference is speculative, assuming that the work being done here will ultimately be released as Servlet 6.2:
https://github.com/jakartaee/servlet/tree/master/spec

PR #765, opened 2024-10-12
Bug/bug 55470 class not found exception during startup patch

emileplas:
https://bz.apache.org/bugzilla/show_bug.cgi?id=55470


markt-asf:
The test case isn't (directly) related to the fix in the PR but it does improve code coverage so I'll commit that separately.

markt-asf:
The remaining changes are essentially a copy of Ralf Hauser's patch on BZ 55470. I'll follow up on that on the BZ issue.

PR #771, opened 2024-10-24
webdav lock and unlock: ensure response must not be cached

Chenjp:
Make sure that Responses of lock / unlock methods MUST NOT be cached, per rfc4918. 
see #disableCacheable()
testcase added.

### DIFF
--- a/java/org/apache/catalina/servlets/WebdavServlet.java
+++ b/java/org/apache/catalina/servlets/WebdavServlet.java
@@ -581,8 +581,10 @@ public class WebdavServlet extends DefaultServlet implements PeriodicEventListen
         } else if (method.equals(METHOD_MOVE)) {
             doMove(req, resp);
         } else if (method.equals(METHOD_LOCK)) {
+            disableCacheable(resp);
             doLock(req, resp);
         } else if (method.equals(METHOD_UNLOCK)) {
+            disableCacheable(resp);
             doUnlock(req, resp);
         } else {
             // DefaultServlet processing
@@ -776,6 +778,20 @@ public class WebdavServlet extends DefaultServlet implements PeriodicEventListen
         }
 
         return methodsAllowed.toString();
+    }
+    
+    /**
+     * Explicitly prevent the response from being cached by middle tier, e.g., <type>ExpiresFilter</type>, http server, or
+     * proxy server, browser.
+     * @param resp The HTTP Servlet response
+     * @see <a href="https://datatracker.ietf.org/doc/html/rfc7234">RFC 7234</a>
+     */
+    protected void disableCacheable(HttpServletResponse resp) {
+        resp.setHeader("Cache-Control", "no-cache, no-store, max-age=0"); // HTTP 1.1
+        /*
+         * Typically other components follow 'Expires' header.
+         */
+        resp.setDateHeader("Expires", 0);
     }
 
 
--- a/test/org/apache/catalina/servlets/TestWebdavServletRfcSection.java
+++ b/test/org/apache/catalina/servlets/TestWebdavServletRfcSection.java
@@ -1,0 +1,203 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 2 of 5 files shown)


rmaucher:
All of the WebDAV methods are "must not cache" actually (except PROPFIND, but IMO it's a bad idea to cache it), and it is 100% obvious why. If the proxy feels like it doesn't have to send the MKCOL to Tomcat, things are not going to work. I don't know if the WebDAV Servlet should have built in headers for that or the users should use a valve, filter or something instead (like the obvious: DO NOT use a caching proxy on the WebDAV location !).

Personally I am "done" with the WebDAV Servlet. The area had no tests (Litmus was sometimes run manually by Mark it seems) so according to our previous meetings there was a need to improve that. Along the way I found design problems and bugs I had to fix, and it now seems good enough (Litmus passes and the testcase scenarios are rather complex). I don't believe there are lots of users for this, so as far as I am concerned it is more than enough.

When adding tests, you might want to look at that: https://nightlies.apache.org/tomcat/tomcat-12.0.x/coverage/org.apache.catalina.servlets/index.html#dn-a
You can set test.coverage=true (and also add execute.validate=true) in your build.properties to see if new tests improve that or if things were already covered.

markt-asf:
This change is unnecessary. As per RFC 9111, methods other than GET are only cached if the method definition allows it. RFC 4918 is explicit that these methods are not cacheable.

rmaucher:
OTOH, +1 if you want to rewrite (they are too verbose right now) and expand the WebDAV tests.

Chenjp:
> This change is unnecessary. As per RFC 9111, methods other than GET are only cached if the method definition allows it. RFC 4918 is explicit that these methods are not cacheable.

Currently, expiresFilter with default conf will enable cache for those LOCK/UNLOCK etc. If source server do not explicitly disable it, then other component like ExpiresFilter, Nginx/Apache Httpd, F5, Proxy Server will have to be responsible for that. 
If the proxy server of our user, how to resolve it - Ask their IT team to modify proxy setting?

markt-asf:
If the `ExpiresFilter` is setting cache control headers when it should then that Filter should be fixed. The same goes for any other component setting cache headers incorrectly.

Chenjp:
> If the `ExpiresFilter` is setting cache control headers when it shouldn't then that Filter should be fixed. The same goes for any other component setting cache headers incorrectly.



> OTOH, +1 if you want to rewrite (they are too verbose right now) and expand the WebDAV tests.

I started reading rfc4918 three weeks ago, and yesterday when i submitted the code i found that @rmaucher you had almost finished WebDAV rfc support.

Chenjp:
@markt-asf @rmaucher please advise following issue need fix or not? -- If necessary, I will submit PR.
```java
java.lang.AssertionError: 9.10.2. Refreshing Locks - Values other than 0 or infinity MUST NOT be used with the Depth header on a LOCK method. Depth header `1` is invalid. expected:<400> but was:<200>
	at org.junit.Assert.fail(Assert.java:89)
	at org.junit.Assert.failNotEquals(Assert.java:835)
	at org.junit.Assert.assertEquals(Assert.java:647)
	at org.apache.catalina.servlets.TestWebdavServletRfcSection_9_10.testRfc_9_10_2(TestWebdavServletRfcSection_9_10.java:199)

java.lang.AssertionError: 9.10.3. Depth and Locking - The Depth header may be used with the LOCK method. Depth header `-1` is invalid. 4xx status coude expected, actual status code was 200
	at org.apache.catalina.servlets.WebdavServletRfcSectionBase$WebdavClient.lockResource(WebdavServletRfcSectionBase.java:164)
	at org.apache.catalina.servlets.TestWebdavServletRfcSection_9_10.testRfc_9_10_3(TestWebdavServletRfcSection_9_10.java:280)	

java.lang.AssertionError: 9.11. UNLOCK Method - 400 (Bad Request) expected due to `No lock token was provided`., Failed to unlock(lockToken=null), actual status code was 409
	at org.apache.catalina.servlets.WebdavServletRfcSectionBase$WebdavClient.unlockResource(WebdavServletRfcSectionBase.java:289)
	at org.apache.catalina.servlets.TestWebdavServletRfcSection_9_11.testRfc_9_11_1(TestWebdavServletRfcSection_9_11.java:190)	
```

rmaucher:
+1 for the third one (the request will not work so it's better to send the specified 400 code immediately), I just fixed it.
For the other two, about depth, I left it as is on purpose. There's no status code mentioned for a bad depth, and there's a default value if absent. IMO pretending it's either 0 if specified or inifinity otherwise is not bad.
Note: There are some items where you must not be strict like the lock timeout just after the depth header in the code.

BTW, if you submit tests, having a WebDAV compliance testsuite is not a goal for Tomcat (it was structured that way with HTTP/2 because HTTP/2 is critical for Tomcat), although we're probably not going to refuse it. The general idea is to have good coverage and cover some complex scenarios. Maybe you can submit code to Litmus and see what happens if something is not covered by it (evidently the three items here are not since Litmus is all ok), since the goal there is to be a WebDAV compliance testsuite.

WebdavServletRfcSectionBase (renamed to WebdavServletBaseTest IMO) could be a good idea to make the existing tests less verbose. Verifying it is really the case would be start, of course.

rmaucher:
Ok, I give up on that one, since it's already done for copyResource. I gave you credit in the commit message.

PR #772, opened 2024-10-28
ExpiresFilter enhance should-not-cacheable verification

Chenjp:
- Skip response with Cache-Control: no-store.
- Introduce "ExpiresIncludedMethods", "ExpiresIncludedResponseStatusCodes" initial parameters to improve configurability.
- Testcase added.


markt-asf:
This doesn't need to be this complicated.

RFC 9110 does not define the list of cacheable status codes, it defines a list of "heuristically cacheable" status codes. Also, the filter already has a mechanism to exclude some status codes.

For the purposes of this filter it is sufficient to only add the cache headers for GET and HEAD. I don't see a need to make that list configurable.

If the application has set any sort of Cache-Control header, this filter shouldn't be adding to or changing that.

markt-asf:
Having looked at the spec again, we don;t need to be that strict about Cache-Control headers set by the application.

Chenjp:
Commit ```c65d66d``` has prevented most of unexpected cases.

PR #776, opened 2024-11-11
fix message parameters - expiresFilter.skippedStatusCode

Chenjp:
expiresFilter.skippedStatusCode: fix message parameters.


markt-asf:
Thanks for the PR. Translations are handled a little differently to the rest of the code. We use POEditor. See https://cwiki.apache.org/confluence/display/TOMCAT/Managing+translations for details.

I've committed a fix for the original English version and updated the translations in POEditor.

PR #777, opened 2024-11-12
bz69439 - fix: improper handling of conflicting cache-control directives

Chenjp:
Currently ExpiresFilter pickup the first-added cache-control header as the effective field, cause follow-up directives (like no-store or max-age) to be ignored.

rfc9111 - 4.2.1 - If directives conflict (e.g., both max-age and no-cache are present), the most restrictive directive should be honored.

* remove cacheControlHeader from servlet response wrapper ```ExpiresFilter.XHttpServletResponse```.
* go through all cache-control fields when verify add-expire-header possibility.
* test case added.

### DIFF
--- a/java/org/apache/catalina/filters/ExpiresFilter.java
+++ b/java/org/apache/catalina/filters/ExpiresFilter.java
@@ -528,11 +528,6 @@ public class ExpiresFilter extends FilterBase {
     public class XHttpServletResponse extends HttpServletResponseWrapper {
 
         /**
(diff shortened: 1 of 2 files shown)


markt-asf:
Thanks for the PR. I have applied a fix based on this PR.

Removing the cache-control header caching removes a deliberate performance optimisation (and doesn't update the Javadoc to reflect that change either). I opted for the simpler approach of merging multiple headers into a single header in the cache. I also slightly changed the test cases.

PR #778, opened 2024-11-15
Partial put - enhance content range verification

Chenjp:
per [rfc 9110 - 14.4](https://httpwg.org/specs/rfc9110.html#field.content-range)

- ContentRange - add isValid(), more length check.
- ContentRange - parse(): make sure the not-null return of parse() is valid. 
- force content-range unit in lowercase internally.
- TestDefaultServletPut - cover issues by adding 2 sets of parameters.


rmaucher:
-1
Other that the lower case range, the shuffling around of the checks introduces a much more serious bug (&& length > end).

rmaucher:
Ok, that's correct, the end is also inclusive.

PR #779, opened 2024-11-18
bz69442: Content type testcase

Chenjp:
add tests.


markt-asf:
Thanks for the PR. I have applied a variation to 9.0.x, 10.1.x 11.0.x and main. The main changes were:

- use a Servlet rather than a JSP
- refactored the code to make `application/x-www-form-urlencoded` a constant

PR #780, opened 2024-11-18
fix multipart/byteranges response body - length attri of ContentRange

Chenjp:
content-range of partial content: length is the target resource ***full content length***.
```java
    ostream.println("Content-Range: bytes " + start + "-" + end + "/" + length);
```
add parameters set to cover this issue.


rmaucher:
Merged in main, 11 and 10.1. Test case added to 9.0.

PR #783, opened 2024-11-22
revoke early hints support on HTTP 1.1

Chenjp:
For compatibility and security reasons, it is recommended to only send HTTP 103 Early Hints responses over HTTP/2 or later unless the client is known to handle informational responses correctly

* see https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/103

### DIFF
--- a/java/org/apache/coyote/http11/Http11Processor.java
+++ b/java/org/apache/coyote/http11/Http11Processor.java
@@ -151,7 +151,12 @@ public class Http11Processor extends AbstractProcessor {
      */
     private final HttpParser httpParser;
 
-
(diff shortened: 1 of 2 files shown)


rmaucher:
Not interested in that one, sorry, Tomcat will implement what the Servlet specification says.

markt-asf:
The requirement to correctly handle 1xx responses has been in the HTTP specs for over 25 years. Any user agent (or proxy) that still can't handle this is broken and needs to be fixed. We will not remove useful features from Tomcat to work around buggy user agents.

PR #784, opened 2024-11-22
enhanced http/2 early hints behavior

Chenjp:
automatically flush when earlyHints called, to enable the possibility of server-think-time  (programmatic code in web app) between the Early Hints headers and the final response.  Speeding up page loads is the main goal of Early hints feature, and the "server think-time" is the key.

### DIFF
--- a/java/org/apache/coyote/http2/Stream.java
+++ b/java/org/apache/coyote/http2/Stream.java
@@ -611,13 +611,13 @@ class Stream extends AbstractNonZeroStream implements HeaderEmitter {
     final void writeEarlyHints() throws IOException {
         MimeHeaders headers = coyoteResponse.getMimeHeaders();
         String originalStatus = headers.getHeader(":status");
-        headers.setValue(":status").setString(String.valueOf(HttpServletResponse.SC_EARLY_HINTS));
+        headers.setValue(":status").setString(Integer.toString(HttpServletResponse.SC_EARLY_HINTS));
 
         try {
             MimeHeaders earlyHintsHeaders = new MimeHeaders();
             earlyHintsHeaders.duplicate(headers);
             earlyHintsHeaders.filter(HTTP_EARLY_HINTS_HEADERS);
-            earlyHintsHeaders.setValue(":status").setString(String.valueOf(HttpServletResponse.SC_EARLY_HINTS));
+            earlyHintsHeaders.setValue(":status").setString(Integer.toString(HttpServletResponse.SC_EARLY_HINTS));
             handler.writeHeaders(this, earlyHintsHeaders, false, Constants.DEFAULT_HEADERS_FRAME_SIZE);
             // to take advantage of the Early Hints feature, the server-think-time is needed between the Early Hints
             // headers and the final response. Therefore, we need emit early hints status and headers via flush.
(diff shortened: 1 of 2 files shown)


java/org/apache/coyote/http2/Stream.java
```diff
@@ -604,10 +611,23 @@ final void writeAck() throws IOException {
...
         MimeHeaders headers = coyoteResponse.getMimeHeaders();
         String originalStatus = headers.getHeader(":status");
-        headers.setValue(":status").setString("103");
+        headers.setValue(":status").setString(String.valueOf(HttpServletResponse.SC_EARLY_HINTS));
+
         try {
-            handler.writeHeaders(this, headers, false, Constants.DEFAULT_HEADERS_FRAME_SIZE);
+            MimeHeaders earlyHintsHeaders = new MimeHeaders();
+            earlyHintsHeaders.duplicate(headers);
+            earlyHintsHeaders.filter(HTTP_EARLY_HINTS_HEADERS);
```
markt-asf, line 619: This isn't how early hints are meant to work. The headers sent via early hints are expected to be in the final response.

java/org/apache/coyote/http2/Stream.java
```diff
@@ -604,10 +611,23 @@ final void writeAck() throws IOException {
...
+            handler.writeHeaders(this, earlyHintsHeaders, false, Constants.DEFAULT_HEADERS_FRAME_SIZE);
+            // to take advantage of the Early Hints feature, the server-think-time is needed between the Early Hints
+            // headers and the final response. Therefore, we need emit early hints status and headers via flush.
+            flush(false);
         } finally {
+            // Once early hints emitted, clean up early hints relative headers.
+            // The final response is responsible for resetting Link/CSP/etc related headers
+            for(String header:HTTP_EARLY_HINTS_HEADERS) {
+                headers.removeHeader(header);
+            }
```
markt-asf, line 630: These should not be removed.

java/org/apache/coyote/http2/Stream.java
```diff
@@ -604,10 +611,23 @@ final void writeAck() throws IOException {
     final void writeEarlyHints() throws IOException {
         MimeHeaders headers = coyoteResponse.getMimeHeaders();
         String originalStatus = headers.getHeader(":status");
-        headers.setValue(":status").setString("103");
+        headers.setValue(":status").setString(String.valueOf(HttpServletResponse.SC_EARLY_HINTS));
```
markt-asf, line 614: I am in favour of replacing 103 with the correct constant but that is this only part of the patch that I currently think should be retained. `Integer.toString()` would be better here though. There are also other places where 103 should be replaced.

java/org/apache/coyote/http2/Stream.java
markt-asf: I am well aware of all of that. That is all an application concern, not a container concern. My previous comments stand.

markt-asf: `HttpServletResponse.setHeader(String,String)` to remove (set to `null`) or replace (set to new value).

markt-asf:
The acceptable changes of this PR (literal 103 to constant)  have been applied separately.

PR #786, opened 2024-11-27
BZ 69477: Wrong 'allowedInternalProxies' param name in the Remote IP Filter documentation 

jorgediaz-lr:
https://bz.apache.org/bugzilla/show_bug.cgi?id=69477

I have detected that the 'allowedInternalProxies' param name in the Remote IP Filter documentation here https://tomcat.apache.org/tomcat-11.0-doc/config/filter.html#Remote_IP_Filter is incorrect.

See correction in the first commit e716e7a2536c571e6bc8ddc5628133bb325e4ab8

I have also added some missing filter-mapping sections in some of the RemoteIpFilter examples

Thank you


jorgediaz-lr:
Thank you!

PR #787, opened 2024-11-27
Fix unsetting of httpOnly and secure flag

kriths:
https://bz.apache.org/bugzilla/show_bug.cgi?id=69478


markt-asf:
Sorry. Looks like our work on this overlapped.

PR #788, opened 2024-11-28
minor update on .gitignore

Chenjp:
ignore ds_store and ant test related outputs.


markt-asf:
I've removed the `.DS_Store` file and added it to `.gitignore`. The presence of the other files is an indicator that something has gone wrong so they should not be ignored. I can see an issue with `derby.log` and I'll get that fixed shortly.

markt-asf:
Apache projects prefer to communicate via public channels - typically the mailing list - rather than IM or other non-public channels. What we lose in terms of speed we more than gain in terms of working as a community.

PR #789, opened 2024-11-28
fix typo

Chenjp:
minor typo fix


java/org/apache/catalina/servlets/DefaultServlet.java
```diff
@@ -1487,7 +1487,7 @@ protected Ranges parseRange(HttpServletRequest request, HttpServletResponse resp
             return FULL;
         }
 
-        // Retrieving the range header (if any is specified
+        // Retrieving the range header if any is specified
```
ebourg, line 1490: I think there was actually a missing parenthesis at the end

PR #790, opened 2024-11-29
fix incorrect partial content response to HEAD request with-Range-header, or from Not-accept-range server

Chenjp:
- Bug: incorrect partial content response to HEAD request with-Range-header
  ***Tomcat*** - 206:
  ```
    C:\Users\chenjp>curl http://localhost:55263/index.html -i -H "Range: bytes=0-10" -I
    HTTP/1.1 206
    vary: accept-encoding
    Accept-Ranges: bytes
    ```
  ***www.apache.org*** offcial site - 200: 
  ```
  C:\Users\chenjp>curl https://www.apache.org -i -H "Range: bytes=0-10" -I
  HTTP/1.1 200 OK
  Connection: keep-alive
  Content-Length: 64483
  ```
- Bug: receives a partial content sc from a Not-accept-range server (DefaultServlet with init param useAcceptRanges=false)
  ```
  C:\Users\chenjp>curl http://localhost:55463/index.html -i -H "Range: bytes=0-10"
  HTTP/1.1 206
  vary: accept-encoding
  ETag: W/"957-1732881015126"
  ```

ChangeLog:
* per RFC 9110 - Section 14, currently GET is the only method.
* A server that turn off accept-range support should ignore Range header field.

### DIFF
--- a/java/org/apache/catalina/servlets/DefaultServlet.java
+++ b/java/org/apache/catalina/servlets/DefaultServlet.java
@@ -1444,9 +1444,15 @@ public class DefaultServlet extends HttpServlet {
     protected Ranges parseRange(HttpServletRequest request, HttpServletResponse response, WebResource resource)
             throws IOException {
 
-        // Range headers are only valid on GET requests. That implies they are
-        // also valid on HEAD requests. This method is only called by doGet()
-        // and doHead() so no need to check the request method.
+        if (useAcceptRanges == false) {
+            // Not supported feature, ignore
+            return FULL;
+        }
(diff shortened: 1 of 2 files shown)


markt-asf:
Needs some small changes but I'll do that after merging but before back-porting.

markt-asf:
Note: It is not required for a server to send `Accept-Ranges` for it to process a request with `Range`.

markt-asf:
No.

markt-asf:
Look at the history of the code. You'll need to go back a long way. There is definitely an argument for deprecating this option and removing it in Tomcat 12.

PR #791, opened 2024-11-29
Reject Range-Request if those ranges are not strictly in ascending order

Chenjp:
Request that ranges are not strictly in ascending order, indicates either a broken client or a deliberate denial-of-service attack.


markt-asf:
Yes. This change has been reverted. It is not RFC 9110 compliant.

rmaucher:
Konstantin has provided examples of requests out there with these kind of ranges. Rejecting them provides no value to Tomcat, so the change was reverted.

PR #793, opened 2024-12-05
Use UTF-8 for logging in Http11InputBuffer

izeye:
I'm not sure if there's any reason not to use UTF-8 there, but for non-ASCII characters, their encoding is broken, so it would be nice to use UTF-8 unless there's other reason not to do so.

### DIFF
--- a/java/org/apache/coyote/http11/Http11InputBuffer.java
+++ b/java/org/apache/coyote/http11/Http11InputBuffer.java
@@ -811,7 +811,7 @@ public class Http11InputBuffer implements InputBuffer, ApplicationBufferHandler,
 
         if (log.isTraceEnabled()) {
             log.trace("Received [" + new String(byteBuffer.array(), byteBuffer.position(), byteBuffer.remaining(),
-                    StandardCharsets.ISO_8859_1) + "]");
+                    StandardCharsets.UTF_8) + "]");
         }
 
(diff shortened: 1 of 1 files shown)


markt-asf:
HTTP headers are expected to be ASCII, not UTF-8.

izeye:
@markt-asf Thanks for the feedback!

Did you mean to say that the log only contains headers?

With [this sample](https://github.com/izeye/hello-spring-boot/tree/web), if I send the following request:

```
% curl -H "Content-Type: application/json" "http://localhost:8080/persons" --data '{"firstName": "테스트1", "lastName": "테스트2"}'
{"firstName":"테스트1","lastName":"테스트2"}
%
```

I can see the following log:

```
2024-12-05T19:37:31.320+09:00 DEBUG 13943 --- [nio-8080-exec-1] o.a.coyote.http11.Http11InputBuffer      : Received [POST /persons HTTP/1.1
Host: localhost:8080
User-Agent: curl/7.87.0
Accept: */*
Content-Type: application/json
Content-Length: 53

{"firstName": "íì¤í¸1", "lastName": "íì¤í¸2"}]
```

It contains the whole request including the request body in addition to the headers.

markt-asf:
We are only interested in the headers. The body could be using any encoding so whatever is chosen the body will be corrupted for some requests.

PR #802, opened 2024-12-18
for delete failure case, respond with SC_CONFLICT rather than SC_METHOD_NOT_ALLOWED

Chenjp:
Since allowed methods check has been performed previously, failure status code switch to 409 / SC_CONFLICT.

Root cause may be insufficient privileges, OS file locking, or already deleted by another concurrent request.


rmaucher:
Changing things just for the sake of it is not a good plan. The return code is acceptable.

PR #803, opened 2024-12-22
Fix BZ69508 - Javasource generator - explicitly set literal false when call printParams to process jsp include page

Chenjp:
In GeneratorVisitor#generateIncludeWithParameters, since we introduce an extra variable urlVariableName to store include page param, then we need explicit specify literal=false when call #printParams.


markt-asf:
Tx for the PR.

Merged manually so I could make some minor changes.

PR #806, opened 2025-01-13
fix BZ 69446 - Propagate maxSwallowSize to enforce partial PUT request target file size

Chenjp:
maxSwallowSize is semantically equivalent to max file size of upload file size. Send 413 when length of Content-Range request exceeds maxSwallowSize.


rmaucher:
As per the response in the BZ, we're not going to address this (esp not using an unrelated configuration setting).

PR #807, opened 2025-01-15
Webdav special path handling fix

Chenjp:
1. fix: "/web-information.xml" is not a special path. 
2. method name refactoring: from isSpecialPath to checkSpecialPath.


rmaucher:
I did not pick up the test since it doesn't add much.
Also because of:
        File appDir = new File(getBuildDirectory(), "webapps" + contextPath);
        try (FileWriter fw = new FileWriter(new File(appDir, "WEB-INFORMATION.xml"))) {

PR #808, opened 2025-01-16
webdav testcase for special path

Chenjp:
add cases for allowSpecialPaths=true.


rmaucher:
This is trivial, mostly already tested, so I did not bother adding them.

PR #809, opened 2025-01-17
Fix memory leak in CrawlerSessionManagerValve

brmautorola:
Remove use of sessionIdClientId map, which is only ever updated, but never queried - causing memory leaks over time


rmaucher:
This is a leftover following a refactoring and got unnoticed. I added a test fix and the changelog.

PR #810, opened 2025-01-22
Ensure partial put data range not exceed ContentRange declared

Chenjp:
Ensure updated content size not exceeds bounds ***(end-start+1)*** declared by Content-Range header.

For backwards compatibility, just discard the rest part instead of a sending 400.


rmaucher:
Surprisingly, I verified that there's no specification language on what to do if content-length (or whatever similar like chunking) conflicts with content-range.

rmaucher:
I'm not sure. Since this is writing to a temporary file rather than the real resource, it is possible to back out without consequences. As a result, I think I will implement throwing an IOE if there's an inconsistency.

rmaucher:
Made changes to try it. Thanks.

PR #812, opened 2025-01-29
Forward toString in MessageBytes toStringType

adwsingh:
toString doesn't check if the value has already been converted to String while toStringType does this. I couldn't find a reason not to do the same for toString

### DIFF
--- a/java/org/apache/tomcat/util/buf/MessageBytes.java
+++ b/java/org/apache/tomcat/util/buf/MessageBytes.java
@@ -163,20 +163,7 @@ public final class MessageBytes implements Cloneable, Serializable {
      */
     @Override
     public String toString() {
-        switch (type) {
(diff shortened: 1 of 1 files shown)


adwsingh:
My use case is I want to avoid having to convert byte to String on repeated invocation of [MimeHeaders.getHeader](https://github.com/apache/tomcat/blob/main/java/org/apache/tomcat/util/http/MimeHeaders.java#L383). 

If this change is not acceptable I can change the getHeader to invoke `mh.toStringType`

rmaucher:
toStringType changes the MB type. toString does not.
This should indicate that it is wise to not change the MB type in some cases. There's a long history with these changes, you should look into the git history.

adwsingh:
@rmaucher in that case is the proposal in my comment acceptable?

PR #813, opened 2025-01-29
Invoke toStringType instead of toString in MimeHeaders

adwsingh:
This is to avoid having to convert from Bytes to String again for the same header.


markt-asf:
This should be safe. I'm just running some tests and reviewing the code before committing.

PR #814, opened 2025-01-30
log connectionId

Dmole:
Add connectionId to the access log options so the connectionId in the error log can be cross referenced to an IP for debug, fail2ban, etc.

### DIFF
--- a/java/org/apache/catalina/valves/AbstractAccessLogValve.java
+++ b/java/org/apache/catalina/valves/AbstractAccessLogValve.java
@@ -111,6 +111,7 @@ import org.apache.tomcat.util.net.IPv6Utils;
  * <li><code>%{xxx}i</code> for incoming headers
  * <li><code>%{xxx}o</code> for outgoing response headers
  * <li><code>%{xxx}c</code> for a specific cookie
+ * <li><code>%{xxx}L</code> xxx is an attribute in the Connection
  * <li><code>%{xxx}r</code> xxx is an attribute in the ServletRequest
  * <li><code>%{xxx}s</code> xxx is an attribute in the HttpSession
  * <li><code>%{xxx}t</code> xxx is an enhanced SimpleDateFormat pattern (see Configuration Reference document for
@@ -157,6 +158,10 @@ public abstract class AbstractAccessLogValve extends ValveBase implements Access
     private enum PortType {
         LOCAL,
         REMOTE
+    }
+
+    private enum ConnectionType {
+        ID
     }
 
     /**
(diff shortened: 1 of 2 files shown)


markt-asf:
This is not the way to do this. Look at the ExtendedAccessLogValve. Look at `x-H(...)` options.

Dmole:
I added it to both AccessLogValve and ExtendedAccessLogValve because it is non-trivial for users to switch between them as the documentation does not provide a mapping (EG: %v = ?), and the extended valve does not permit arbitrary formatting (EG: JSON log).

is that OK?

rainerjung:
Concerning analogies in Apache httpd land:

ErrorLogFormat has

```
%{c}L 	Log ID of the connection
%{C}L 	Log ID of the connection if used in connection scope, empty otherwise
```

and explains them as

`The log ID format %L produces a unique id for a connection or request. This can be used to correlate which log lines belong to the same connection or request, which request happens on which connection. A %L format string is also available in [mod_log_config](https://httpd.apache.org/docs/2.4/en/mod/mod_log_config.html) to allow to correlate access log entries with error log lines. If [mod_unique_id](https://httpd.apache.org/docs/2.4/en/mod/mod_unique_id.html) is loaded, its unique id will be used as log ID for requests.`

See: https://httpd.apache.org/docs/2.4/en/mod/core.html#errorlogformat

And as mentioned here, mod_log_config documents:
`
%L 	The request log ID from the error log (or '-' if nothing has been logged to the error log for this request). Look for the matching error log line to see what request caused what error.`

I haven't checked what the connection id in Tomcat land is and how well this analogy applies.

Furthermore mod_https provides some IDs via httpd env vars (%{...}e):

```
Variable Name: 	Value Type: 	Description:
HTTP2	flag	HTTP/2 is being used.
H2PUSH	flag	HTTP/2 Server Push is enabled for this connection and also supported by the client.
H2_PUSH	flag	alternate name for H2PUSH
H2_PUSHED	string	empty or PUSHED for a request being pushed by the server.
H2_PUSHED_ON	number	HTTP/2 stream number that triggered the push of this request.
H2_STREAM_ID	number	HTTP/2 stream number of this request.
H2_STREAM_TAG	string	HTTP/2 process unique stream identifier, consisting of connection id and stream id separated by 
```-.

Dmole:
Add %{c}L and x-H(connectionId) to AccessLogValve and ExtendedAccessLogValve to cross reference errors from catalina.log

markt-asf:
Manually applied as there were a few things I needed to add/change
- %{xxx}L is for ID attributes not connection attributes
- new settings needed to be added to the docs
- change log entry was required
- The CachedElement interface does not apply here

Dmole:
https://github.com/apache/tomcat/commit/3cfc9bc048109c5a566fc24539e5bb9759c114b6

https://github.com/apache/tomcat/commit/987f737afe4f08da939d870b5b1ec4b96010d926

PR #816, opened 2025-01-31
Fix null stream issue for resource loading based on relative names

AmshikaH:
This addresses a case in the `getResourceAsStream` [1] method which falsely identifies resources to be absent when looking them up with relative names (names that don't start with a slash) if they have already been looked up with the absolute name (names that start with a slash). 

In the getResourceAsStream [1] method, Tomcat checks the recently introduced `notFoundClassResources` based on the "path" variable alone, which is always the absolute name.

However, there are cases such as [2] in the Apache CXF core component where the class loading first tries calling the `getResourceAsStream` [1] method with the absolute name as the “name” argument and if this returns a null stream, it then tries the relative name by calling the `getResourceAsStream` [1] method with the relative name as the “name”.

In both calls to the `getResourceAsStream` method, the “path” variable set at [3] is the absolute name and therefore, Tomcat checks the cache for the absolute name both times.

As a result, the second attempt based on the relative name gets skipped, and a null stream is returned even though the resource is present.

In this PR, the block where the resource is loaded based on the "name" variable has been moved to a separate if block where the cache is checked for the "name" variable instead of the "path" variable.

[1] https://github.com/apache/tomcat/blob/08206671035f05e205806140b11c10c6eb3ca5c3/java/org/apache/catalina/loader/WebappClassLoaderBase.java#L1070
[2] https://github.com/apache/cxf/blob/cxf-3.5.9/core/src/main/java/org/apache/cxf/version/Version.java#L38
[3] https://github.com/apache/tomcat/blob/08206671035f05e205806140b11c10c6eb3ca5c3/java/org/apache/catalina/loader/WebappClassLoaderBase.java#L1100

### DIFF
--- a/java/org/apache/catalina/loader/WebappClassLoaderBase.java
+++ b/java/org/apache/catalina/loader/WebappClassLoaderBase.java
@@ -1098,14 +1098,12 @@ public abstract class WebappClassLoaderBase extends URLClassLoader
             log.trace("  Searching local repositories");
         }
         String path = nameToPath(name);
-        if (!notFoundClassResources.contains(path)) {
+        if (name.startsWith("/") || !notFoundClassResources.contains(path)) {
             WebResource resource = resources.getClassLoaderResource(path);
             if (resource.exists()) {
                 stream = resource.getInputStream();
                 trackLastModified(path, resource);
             }
-        }
-        if (!notFoundClassResources.contains(name)) {
             try {
                 if (hasExternalRepositories && stream == null) {
                     URL url = super.findResource(name);
@@ -1116,18 +1114,13 @@ public abstract class WebappClassLoaderBase extends URLClassLoader
             } catch (IOException e) {
                 // Ignore
             }
-        }
-        if (stream != null) {
-            if (log.isTraceEnabled()) {
-                log.trace("  --> Returning stream from local");
+            if (stream != null) {
+                if (log.isTraceEnabled()) {
+                    log.trace("  --> Returning stream from local");
+                }
+                return stream;
             }
-            return stream;
-        }
-        if (!notFoundClassResources.contains(path)) {
             notFoundClassResources.add(path);
-        }
-        if (!notFoundClassResources.contains(name)) {
-            notFoundClassResources.add(name);
         }
 
         // (3) Delegate to parent unconditionally
--- a/java/org/apache/catalina/loader/WebappClassLoaderBase.java
+++ b/java/org/apache/catalina/loader/WebappClassLoaderBase.java
@@ -1104,6 +1104,8 @@ public abstract class WebappClassLoaderBase extends URLClassLoader
                 stream = resource.getInputStream();
                 trackLastModified(path, resource);
             }
+        }
+        if (!notFoundClassResources.contains(name)) {
             try {
                 if (hasExternalRepositories && stream == null) {
                     URL url = super.findResource(name);
@@ -1114,13 +1116,18 @@ public abstract class WebappClassLoaderBase extends URLClassLoader
             } catch (IOException e) {
                 // Ignore
             }
-            if (stream != null) {
-                if (log.isTraceEnabled()) {
-                    log.trace("  --> Returning stream from local");
-                }
-                return stream;
+        }
+        if (stream != null) {
+            if (log.isTraceEnabled()) {
+                log.trace("  --> Returning stream from local");
             }
+            return stream;
+        }
+        if (!notFoundClassResources.contains(path)) {
             notFoundClassResources.add(path);
+        }
+        if (!notFoundClassResources.contains(name)) {
+            notFoundClassResources.add(name);
         }
 
         // (3) Delegate to parent unconditionally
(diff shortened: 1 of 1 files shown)


markt-asf:
`getResourceAsStream()` is documented to accept a name, not a URL. I'd like to understand the use case better before making any changes. Can you provide some sample code or a test case or something similar?

AmshikaH:
Sorry, I meant absolute and relative paths, not URLs. I have updated the description body to reflect this.

An example case would be the calling the [`getResourceAsStream` method in Apache CXF Core's Version class](https://github.com/apache/cxf/blob/cxf-3.5.9/core/src/main/java/org/apache/cxf/version/Version.java#L38) (I use cxf-3.5.9) where the class loader is Tomcat's webapp class loader. The relevant snippet from Apache CXF Core's Version class is given below:

Screenshot 2025-02-04 at 16 33 40

In line 40, Apache CXF Core calls Tomcat's `getResourceAsStream` method by passing the absolute path `/org/apache/cxf/version/version.properties` as the name.

In line 42, Tomcat's `getResourceAsStream` method is called a second time by passing the relative path `org/apache/cxf/version/version.properties` as the name.

In both calls, the variables within Tomcat's `getResourceAsStream` are as follows.

- First call: 
  - name: `/org/apache/cxf/version/version.properties` (absolute)
  - path: `/org/apache/cxf/version/version.properties` (absolute)
- Second call:
  - name: `org/apache/cxf/version/version.properties` (relative)
  - path: `/org/apache/cxf/version/version.properties` (absolute)

Regardless of whether the `name ` variable is relative or absolute, the `path` variable is always absolute as it is derived using Tomcat's [`nameToPath` method](https://github.com/apache/tomcat/blob/9647a025443861f6fb6159b9c90f91dba6c7e7b4/java/org/apache/catalina/loader/WebappClassLoaderBase.java#L2379) in Tomcat's [`getResourceAsStream` method](https://github.com/apache/tomcat/blob/08206671035f05e205806140b11c10c6eb3ca5c3/java/org/apache/catalina/loader/WebappClassLoaderBase.java#L1100):

Screenshot 2025-02-04 at 17 06 43

Since the caching logic within Tomcat's `getResourceAsStream` only checks the `path` variable which is absolute, the second call based on the relative path gives a false positive that the resource has already been looked up and found to be unavailable with Tomcat 9.0.97 and 9.0.98.

markt-asf:
These are names not paths.

I'm not convinced that 'relative' and 'absolute' are concepts that apply here.

Putting that to one side, where is the target resource located so that the 'absolute' lookup fails but the 'relative' one works? Also, relative to what? What would be the 'absolute' path be that did work?

markt-asf:
The Java documentation for resource names also states:

> The methods in ClassLoader use the given String as the name of the resource without applying any absolute/relative transformation (see the methods in Class). The name should not have a leading "/".

That doesn't mean the issue isn't going to be fixed in Tomcat. But it does mean it looks like CXF has a bug that should be fixed.

markt-asf:
And given that `addURL()` is a protected method, how are you calling it?

markt-asf:
I think I understand everything that is going on now.

The root cause is that Tomcat accepts name for `ClassLoader.getResource(String name)` and friends with a leading '/' whereas the JRE provided class loaders do not.

My plan to deal with this is:
- Tomcat 12 onwards return `null` if the name starts with `/`
- Tomcat 11 and earlier, skip the `notFoundClassResources` cache if the name starts with '/'

Apps that use incorrect resource names may see a small performance degradation if they try loading the same resource multiple times using the wrong format for a resource name. The fix for that would be for the app to use the correct format.

AmshikaH:
> But it does mean it looks like CXF has a bug that should be fixed.

Thank you for pointing this out. I will make a report of this via CXF's Jira board.

> And given that addURL() is a protected method, how are you calling it?

We are using a custom class loader that extends Tomcat's WebappClassLoader, which makes our custom class loader a subclass of Tomcat's WebappClassLoaderBase. 

> * Tomcat 12 onwards return `null` if the name starts with `/`
> * Tomcat 11 and earlier, skip the `notFoundClassResources` cache if the name starts with '/'
>
> Apps that use incorrect resource names may see a small performance degradation if they try loading the same resource multiple times using the wrong format for a resource name. The fix for that would be for the app to use the correct format.

Thank you, a fix would be much appreciated.

AmshikaH:
> The Java documentation for resource names also states:
> 
> > The methods in ClassLoader use the given String as the name of the resource without applying any absolute/relative transformation (see the methods in Class). The name should not have a leading "/".
> 
> That doesn't mean the issue isn't going to be fixed in Tomcat. But it does mean it looks like CXF has a bug that should be fixed.

Please find the issue on CXF's Jira board here: https://issues.apache.org/jira/browse/CXF-9108

I have credited you by linking your GitHub profile and the quoted comment in the report. If you have an account in the Jira issue board that you would prefer being tagged instead, please do let me know and I can update the issue.

AmshikaH:
> * Tomcat 11 and earlier, skip the `notFoundClassResources` cache if the name starts with '/'

I have updated the PR according to this approach.

PR #817, opened 2025-02-02
Replace \r\n with SimpleHttpClient.CRLF in Http2TestBase & Cleanup RequestUtil

koust6u:
- replace "\r\n\" to SimpleHttpClient.CRLF in Http2TestBsae
- unnecessary code deletion and secure origin comparison in RequestUtil


markt-asf:
1 change per PR please.

Note that we try to avoid static imports if at all possible.

PR #820, opened 2025-02-08
Enhance case sensitivity check

Chenjp:
fix incorrect case sensitivity result when the target base directory (dir name without [a-z|A-Z] / or NON-ENGLISH language) on windows is changed to case sensitive.

Previous behavior:
1. Windows file system.
2. directory "tomcat/webapps" is case insensitive.
3. sub directory "tomcat/webapps/1" (maybe "tomcat/webapps/应用1“) is changed to case sensitive, via ```fsutil.exe file setCaseSensitiveInfo <path> enable```
4. DirResourceSet#isCaseSensitive returns case insensitive of directory "tomcat/webapps/1".
5. Unexpected.

### DIFF
--- a/java/org/apache/catalina/webresources/DirResourceSet.java
+++ b/java/org/apache/catalina/webresources/DirResourceSet.java
@@ -343,7 +343,10 @@ public class DirResourceSet extends AbstractFileResourceSet implements WebResour
      */
     private boolean isCaseSensitive() {
         try {
-            String canonicalPath = getFileBase().getCanonicalPath();
+            // Some file system (e.g. windows) treat files and directories as case insensitive by default, and also support
+            // setting case sensitivity per directory. Then we have to handle case sensitivity for each directory.
+            // Ensure file path contain ENGLISH string, otherwise its upper and lower maybe are same.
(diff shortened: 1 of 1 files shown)


markt-asf:
This enhanced check doesn't work.

If the target file doesn't exist, the part of the canonical name that represents the file will be whatever value is provided to the File constructor. Applying this PR would have the effect of every file system being treated as case insensitive.

rmaucher:
If we start going crazy and creating some file, would it be enough to simply try creating a second file with a different case, or check if the file with another case exists ?
But I don't want to be creating files in the first place. Since this is about covering some edge case, let's do nothing.

markt-asf:
There are still issues with the updated approach. That Windows allows case sensitivity to be controlled on a per directory basis is ... unhelpful. I've updated the case sensitivity check and for the few edge cases it still gets wrong, it fails the right way.

markt-asf:
Mounting case insensitive file systems on Linux adds yet more complexity. And I don't see an easy way to address that - especially if we want to avoid creating files to test case sensitivity (we do). I am beginning to think skipping the check and following the case insensitive code path is the way to go.

PR #821, opened 2025-02-10
BZ69504 move force recycle from log access method

Chenjp:
extract recycling ops on external request/response from CoyoteAdapter#log.


rmaucher:
The PR does a lot more than what it claims to do. I will not merge it.

markt-asf:
I share Rémy's concerns. I see multiple concerns being addressed in this PR.
Correct recycling is critical to the correct operation of Tomcat and log still recycles the request/response in some circumstances.

I'm not against refactoring to make the code cleaner and/or clearer but this needs to be done in small steps that can be verified as the refactoring progresses.

PR #823, opened 2025-02-10
BZ69446 - add parameter maxPutFileSize in DefaultServlet

Chenjp:
Enhance PUT destination resource size restriction on both Fully and Partial PUTs.

### DIFF
--- a/java/org/apache/catalina/servlets/DefaultServlet.java
+++ b/java/org/apache/catalina/servlets/DefaultServlet.java
@@ -260,6 +260,11 @@ public class DefaultServlet extends HttpServlet {
      */
     private boolean useStrongETags = false;
 
+    /**
+     * Maximum size of PUT target resource.
+     */
+    private long maxPutFileSize = -1L;
+
 
     // --------------------------------------------------------- Public Methods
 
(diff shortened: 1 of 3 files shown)


rmaucher:
The feedback was that this serves no purpose. I believe it is correct.

For processing of POST and some others like that, processing takes memory. Also we do not know what the app will do with the data. Hence limiting could be important and it justifies having some setting.

But PUT in the DefaultServlet is not the same. Let's say a security conscious user such as you already disabled partial PUT. A malicious user can then write the same amount of data on the server storage using a trivial increase of the amount of input even if your setting is in place. The main difference would be using separate HTTP requests. So then you would be expected to catch that pattern using some kind of extra protection (where it would also be possible to check the request length) ?

As a result, adding this setting still does not make sense to me.

rmaucher:
We answered this quite a few times already, so let's leave it at that. Write enabled allows editing webapp contents, which is reserved to administrators.

ChristopherSchultz:
> > The feedback was that this serves no purpose. I believe it is correct.
> > For processing of POST and some others like that, processing takes memory. Also we do not know what the app will do with the data. Hence limiting could be important and it justifies having some setting.
> > But PUT in the DefaultServlet is not the same. Let's say a security conscious user such as you already disabled partial PUT. A malicious user can then write the same amount of data on the server storage using a trivial increase of the amount of input even if your setting is in place. The main difference would be using separate HTTP requests. So then you would be expected to catch that pattern using some kind of extra protection (where it would also be possible to check the request length) ?
> > As a result, adding this setting still does not make sense to me.
> 
> Rate limiter filter, or WAF can detect and block those requests.

Either of those two can also provide the limits you are requesting, here. Adding a feature to Tomcat which requires additional protection from e.g. WAF is kinda silly when the WAF could provide the new feature as well.

Chenjp:
@ChristopherSchultz Currently, no OWASP DRS rule against this pattern.

PR #826, opened 2025-02-13
Refactor CORS Request Type Checking Logic for Simplicity and Clarity

koust6u:
This PR refactors the checkRequestType method in the CORS filter to simplify and improve the readability of the request type determination logic. The main changes are:

+ Early Returns: The code has been simplified to return early for various conditions (e.g., invalid CORS requests, same-origin requests, missing or invalid headers). This reduces nesting and enhances readability.
+ Consolidation of Conditions: Some conditional blocks that check for empty or invalid headers were merged to avoid redundant checks, making the code more concise.
+ Improved Flow: The logic now more clearly distinguishes between the different CORS request types by handling the OPTIONS method, GET/HEAD methods, and POST method (with content type checks) in a more straightforward manner.

### DIFF
--- a/java/org/apache/catalina/filters/CorsFilter.java
+++ b/java/org/apache/catalina/filters/CorsFilter.java
@@ -538,58 +538,41 @@ public class CorsFilter extends GenericFilter {
      * @return the CORS type
      */
     protected CORSRequestType checkRequestType(final HttpServletRequest request) {
-        CORSRequestType requestType = CORSRequestType.INVALID_CORS;
         if (request == null) {
             throw new IllegalArgumentException(sm.getString("corsFilter.nullRequest"));
         }
         String originHeader = request.getHeader(REQUEST_HEADER_ORIGIN);
-        // Section 6.1.1 and Section 6.2.1
-        if (originHeader != null) {
-            if (originHeader.isEmpty()) {
-                requestType = CORSRequestType.INVALID_CORS;
-            } else if (!RequestUtil.isValidOrigin(originHeader)) {
-                requestType = CORSRequestType.INVALID_CORS;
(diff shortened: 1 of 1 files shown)


n828cl:
I'm not sure that the revised code properly implements the spec. (Unfortunately, the current spec is much harder to follow than the previous iteration.) Also, calling IsSameOrigin() before isValidOrigin() seems just a tad risky.

dsoumis:
The revised code, though less nested, adheres to the CORS flow as specified. In my opinion, this structure offers a more maintainable and understandable implementation.
Also the ordering of the methods appears safe. In isSameOrigin(), any malformed header would not erroneously be classified as same-origin. Thus, will result in the subsequent check with isValidOrigin().

koust6u:
@n828cl 
Which part do you find harder to follow? If you could clarify, it would help me understand better.  

I also believe that calling `isSameOrigin()` before `isValidOrigin()` shouldn't cause any issues. However, could you explain your concerns in more detail? If there's a potential problem, I'll make the necessary adjustments.

markt-asf:
I echo n828cl's concerns about calling `isSameOrigin()` before `isValidOrigin()`. I can't see an obvious issue but swapping the order of those two checks is going to be a lot faster than doing all the thinking and analysis to convince ourselves that the current order is safe. It will mean 3 `if` statements rather than 2 but I'm fine with that. Also, if we ever need to extend the checks in `isValidOrigin()`, the current order risks bypassing those additional checks.

I am happy that the logic (apart from the above) is the same.

dsoumis:
Indeed Mark, that's a fair point. I've refactored the order as suggested in: [commit](https://github.com/apache/tomcat/commit/e451872) .

PR #827, opened 2025-02-15
Refactor CSRF Filter: Simplify Logic, Improve Logging, and Enhance Code Readability

koust6u:
1. **Simplified CSRF Filter Logic**:
   - Applied `early return` to eliminate unnecessary `else` statements, improving code readability.
   - Simplified conditionals to reduce duplication and streamline logic.

2. **Extracted Logging Logic**:
   - Moved CSRF-related logging logic into a separate `CsrfLoggingHelper` class to improve reusability and maintainability.
   - Ensured consistent logging levels (debug, trace) are managed within the helper class.

4. **Simplified Return Handling**:
   - Streamlined return statements to improve control flow and eliminate unnecessary condition checks.

### DIFF
--- a/java/org/apache/catalina/filters/CsrfPreventionFilter.java
+++ b/java/org/apache/catalina/filters/CsrfPreventionFilter.java
@@ -327,7 +327,8 @@ public class CsrfPreventionFilter extends CsrfPreventionFilterBase {
         HttpSession session = req.getSession(false);
 
         String requestedPath = getRequestedPath(req);
-        boolean skipNonceCheck = skipNonceCheck(req);
+        CsrfLoggingHelper csrfLoggingHelper = new CsrfLoggingHelper(session, requestedPath);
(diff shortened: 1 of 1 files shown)


dsoumis:
1) I am not in favor of the helper class. Conditions like if (log.isDebugEnabled()) and if (log.isTraceEnabled()) are widely used in the code and I believe it makes it more transparent when the logging occurs.

2) Since early returns (guard clauses) are not applied to all conditions, removing the else blocks makes the code harder to analyze.

3) The filter is being invoked twice

koust6u:
Thank you for the thorough review and detailed feedback. I now understand your concerns regarding the logging helper and the importance of explicitly showing when logging occurs. Your point about the removal of else blocks making the code harder to analyze also makes sense, especially given that early returns were not applied consistently. Additionally, I see the issue with the filter being invoked twice.

PR #830, opened 2025-02-24
Fix info-then-debug logging in Http2UpgradeHandler

bergander:
Fix info-then-debug logging in `Http2UpgradeHandler` by making the `UserDataHelper` static. Otherwise the _lastInfoTime_ in `UserDataHelper` will be per HTTP2-connection, but it should be a global scoped timestamp.

This also makes it consistant with how `UserDataHelper` is used in `Cookie`.


markt-asf:
Tx. Well-spotted.

PR #832, opened 2025-03-19
Expanding Parallel Deployment section of Tomcat Docs

usmazat:
Added a paragraph in the Parallel deployment section of Tomcat docs to include an example of getting current webappVersion via property 
Original example was  provided by @ChristopherSchultz


markt-asf:
If you mention one request attribute you should mention them all. You might need a table.

usmazat:
> If you mention one request attribute you should mention them all. You might need a table.

I put together a table of attributes. wasn't sure where to put it so I created a separate xml document and put a link in a context.xml

markt-asf:
I like the idea. I don't think we have a list of ServletContext attributes in a single place. The implementation needs a few tweaks to fit in with how we build the Tomcat docs. I'll re-work this using systemprops.xml as a basis.

Also, the list needs to be limited to ServletContext attributes. Other values - such as system properties - are documented elsewhere.

markt-asf:
Nothing to do on your part. Thanks for the work you put into this. I've added a new runtime attribute section to the configuration documentation that documents the request and servlet context attributes.

PR #834, opened 2025-03-26
Changed string comparison - need to use equals() instead of == or !=

dk2k:


rmaucher:
This is most likely on purpose.

dk2k:
@rmaucher Purpose like what? == comparison is OK when both strings are interned, and I don't think it's the case. 
And it's safe to use equals() for interned strings anyway

markt-asf:
There should be an explanation for each of those in the spot bugs exclusion file.

java/org/apache/catalina/realm/DigestCredentialHandlerBase.java
```diff
@@ -288,7 +289,7 @@ protected String mutate(String inputCredentials, byte[] salt, int iterations, in
      * @return <code>true</code> if the strings are equal to each other, <code>false</code> otherwise.
      */
     public static boolean equals(final String s1, final String s2, final boolean ignoreCase) {
-        if (s1 == s2) {
+        if (Objects.equals(s1, s2)) {
```
ChristopherSchultz, line 292: No. This is obviously a performance optimization and should not be replaced.

java/org/apache/jasper/compiler/Parser.java
```diff
@@ -1471,8 +1471,8 @@ private void parseBody(Node parent, String tag, String bodyType) throws JasperEx
                 err.jspError(start, "jasper.error.emptybodycontent.nonempty", tag);
             }
         } else if (bodyType.equalsIgnoreCase(TagInfo.BODY_CONTENT_JSP) ||
-                bodyType.equalsIgnoreCase(TagInfo.BODY_CONTENT_SCRIPTLESS) || (bodyType == JAVAX_BODY_CONTENT_PARAM) ||
-                (bodyType == JAVAX_BODY_CONTENT_TEMPLATE_TEXT)) {
+                bodyType.equalsIgnoreCase(TagInfo.BODY_CONTENT_SCRIPTLESS) || (bodyType.equals(JAVAX_BODY_CONTENT_PARAM)) ||
+                (bodyType.equals(JAVAX_BODY_CONTENT_TEMPLATE_TEXT))) {
```
ChristopherSchultz, line 1475: The only time this will be true is when this particular constant is being used, internally. Using .equals is needless pedantry.

java/org/apache/tomcat/util/digester/Digester.java
```diff
@@ -2037,7 +2038,7 @@ private StringBuilder updateBodyText(StringBuilder bodyText) {
             return bodyText; // return unchanged data
         }
 
-        if (out == in) {
+        if (Objects.equals(out, in)) {
```
ChristopherSchultz, line 2041: No. This is a performance de-optimization.

java/org/apache/tomcat/util/net/openssl/panama/OpenSSLContext.java
```diff
@@ -503,7 +504,7 @@ public void init(KeyManager[] kms, TrustManager[] tms, SecureRandom sr) throws K
                 }
             }
             // Check if the ciphers have been changed from the defaults
-            if (maxTlsVersion >= TLS1_3_VERSION() && (sslHostConfig.getCiphers() != SSLHostConfig.DEFAULT_TLS_CIPHERS)) {
+            if (maxTlsVersion >= TLS1_3_VERSION() && (!Objects.equals(sslHostConfig.getCiphers(), SSLHostConfig.DEFAULT_TLS_CIPHERS))) {
```
ChristopherSchultz, line 507: This one is at least not-harmful but unnecessary. The default value of Ciphers is DEFAULT_TLS_CIPHERS (exactly, as in "the exact same object reference").

ChristopherSchultz (commented):
Running a static code analyzer on Tomcat source code will likely produce warnings of this nature, because the static code analyzer isn't smart enough to know what the intent is of the comparisons. It just says "== considered harmful" and makes all these recommendations.

I don't believe any of these changes are warranted.

java/org/apache/tomcat/util/net/openssl/panama/OpenSSLContext.java
rmaucher: The is supposed to check if the value was explicitly configured, even if it stayed the same as the default in the end. This is for TLS 1.3 cipher suite stuff (I'm not an expert on that).

java/org/apache/catalina/realm/DigestCredentialHandlerBase.java
dk2k: memory adresses are compared here instead of string values

PR #835, opened 2025-03-26
Changed string concatenations to chained calls of append()

dk2k:
Justification: each string concatenation becomes an anonimous stringbuilder, from which toString() result is used.
It's about performance.


rmaucher:
Doing this should not be needed for constant strings (so the IDE does not complain about it).

PR #836, opened 2025-04-01
webapps/docs: fix spelling in XML files

jbampton:


markt-asf:
Tx. for the PR. I'll get it back-ported to the earlier versions as well.

PR #839, opened 2025-04-01
Fix spelling in Java and JSP files

jbampton:


rmaucher:
I don't agree with using PRs for trivial fixes like this. A committer has to do them (in batches, like I did), otherwise it creates a mountain of spam (and CI runs, and so on).

markt-asf:
I'd rather have these fixes in the code base than not. I didn't think the batch size had been too bad. Small enough it was easy to review but large enough there haven't been that many today. I wouldn't object to fewer larger PRs along these lines. The mailing list noise is always going to be there when we have four active branches.

PR #842, opened 2025-04-04
Adding nonstandard support for c:set and c:remove

jengebr:
Enables optimized, non-standard behavior for the most common JSP tags, when configured to do so.

**Impact:** benchmark (included)shows 99% reduction in cpu, 100% reduction in object allocation

**Activation:** off by default, enabled via an init param to the JSP servlet, such as:
```
      <init-param>
        <param-name>useNonstandardTagOptimizations</param-name>
        <param-value>c:set,c:remove</param-value>
      </init-param>
```

**Example:**
Old code (standard): one method, two parameters, 10+ lines of code, one object allocated (and it's large!).
```
  private boolean _jspx_meth_c_005fset_005f39(javax.servlet.jsp.tagext.JspTag _jspx_th_c_005fwhen_005f11, javax.servlet.jsp.PageContext _jspx_page_context)
          throws java.lang.Throwable {
    javax.servlet.jsp.PageContext pageContext = _jspx_page_context;
    javax.servlet.jsp.JspWriter out = _jspx_page_context.getOut();
    //  c:set
    org.apache.taglibs.standard.tag.rt.core.SetTag _jspx_th_c_005fset_005f39 = new org.apache.taglibs.standard.tag.rt.core.SetTag();
    _jsp_getInstanceManager().newInstance(_jspx_th_c_005fset_005f39);
    try {
      _jspx_th_c_005fset_005f39.setPageContext(_jspx_page_context);
      _jspx_th_c_005fset_005f39.setParent((javax.servlet.jsp.tagext.Tag) _jspx_th_c_005fwhen_005f11);
      // /WEB-INF/views/jsp/features/buybox/offerDisplayGroupLayout.jsp(230,12) name = var type = java.lang.String reqTime = false required = false fragment = false deferredValue = false expectedTypeName = null deferredMethod = false methodSignature = null
      _jspx_th_c_005fset_005f39.setVar("groupName");
      // /WEB-INF/views/jsp/features/buybox/offerDisplayGroupLayout.jsp(230,12) name = value type = javax.el.ValueExpression reqTime = true required = false fragment = false deferredValue = true expectedTypeName = java.lang.Object deferredMethod = false methodSignature = null
      _jspx_th_c_005fset_005f39.setValue(new org.apache.jasper.el.JspValueExpression("/WEB-INF/views/jsp/features/buybox/offerDisplayGroupLayout.jsp(230,12) '${tabContents.keySet().toArray()[0]}'",_jsp_getExpressionFactory().createValueExpression(_jspx_page_context.getELContext(),"${tabContents.keySet().toArray()[0]}",java.lang.Object.class)).getValue(_jspx_page_context.getELContext()));
      int _jspx_eval_c_005fset_005f39 = _jspx_th_c_005fset_005f39.doStartTag();
      if (_jspx_th_c_005fset_005f39.doEndTag() == javax.servlet.jsp.tagext.Tag.SKIP_PAGE) {
        return true;
      }
    } finally {
      org.apache.jasper.runtime.JspRuntimeLibrary.releaseTag(_jspx_th_c_005fset_005f39, _jsp_getInstanceManager(), false);
    }
    return false;
  }

```

New code (nonstandard) - one parameter, one line of code, zero objects allocated.
```
  private void _jspx_meth_c_005fset_005f0(jakarta.servlet.jsp.PageContext _jspx_page_context)
          throws java.lang.Throwable {
    //  c:set
    // /jsp/generator/nonstandard/set-01.jsp(18,0) name = var type = java.lang.String reqTime = false required = false fragment = false deferredValue = false expectedTypeName = null deferredMethod = false methodSignature = null
    // /jsp/generator/nonstandard/set-01.jsp(18,0) name = value type = jakarta.el.ValueExpression reqTime = true required = false fragment = false deferredValue = true expectedTypeName = java.lang.Object deferredMethod = false methodSignature = null
    org.apache.jasper.runtime.JspRuntimeLibrary.nonstandardSetTag(_jspx_page_context, "testVar", new org.apache.jasper.el.JspValueExpression("/jsp/generator/nonstandard/set-01.jsp(18,0) '${'testValue'}'",_jsp_getExpressionFactory().createValueExpression(_jspx_page_context.getELContext(),"${'testValue'}",java.lang.Object.class)).getValue(_jspx_page_context.getELContext()), jakarta.servlet.jsp.PageContext.PAGE_SCOPE);
  }
```

Caller code is simplified by eliminating the boolean check, from:
```
          if (_jspx_meth_c_005fset_005f39(_jspx_th_c_005fwhen_005f11, _jspx_page_context))
            return true;
```
to
```
        _jspx_meth_c_005fset_005f125(_jspx_page_context);
```

### DIFF
--- a/java/org/apache/jasper/compiler/Generator.java
+++ b/java/org/apache/jasper/compiler/Generator.java
@@ -51,6 +51,7 @@ import org.apache.jasper.JasperException;
 import org.apache.jasper.JspCompilationContext;
 import org.apache.jasper.TrimSpacesOption;
 import org.apache.jasper.compiler.Node.ChildInfoBase;
+import org.apache.jasper.compiler.Node.JspAttribute;
 import org.apache.jasper.compiler.Node.NamedAttribute;
 import org.apache.jasper.runtime.JspRuntimeLibrary;
 import org.xml.sax.Attributes;
@@ -69,6 +70,8 @@ class Generator {
     private static final Pattern PRE_TAG_PATTERN = Pattern.compile("(?s).*(<pre>|</pre>).*");
 
     private static final Pattern BLANK_LINE_PATTERN = Pattern.compile("(\\s*(\\n|\\r)+\\s*)");
+
+    private static final String CORE_LIBS_URI = "http://java.sun.com/jsp/jstl/core";
 
     private final ServletWriter out;
 
@@ -101,6 +104,8 @@ class Generator {
     private final DateFormat timestampFormat;
 
     private final ELInterpreter elInterpreter;
+
+    private final Set<String> nonstandardCustomTagNames;
 
     private final StringInterpreter stringInterpreter;
 
@@ -1480,6 +1485,9 @@ class Generator {
 
         @Override
         public void visit(Node.CustomTag n) throws JasperException {
+            if (visitPotentiallyNonstandardCustomTag(n)) {
+                return;
+            }
 
             // Use plugin to generate more efficient code if there is one.
             if (n.useTagPlugin()) {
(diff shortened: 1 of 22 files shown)


jengebr:
> While only targeting `c:set/c:remove` - Do you also need to check the tag belongs to `<uri>http://java.sun.com/jsp/jstl/core</uri>` ? I suspect the user wouldn't enable the flag then use a different library with the same tag names, but it seems like a possible weird outcome

Great suggestion, thank you.  Edge cases are a big concern so please keep them coming!

java/org/apache/jasper/runtime/JspRuntimeLibrary.java
```diff
@@ -957,4 +957,21 @@ public static void releaseTag(Tag tag, InstanceManager instanceManager) {
...
     }
+
+    public static void nonstandardSetTag(jakarta.servlet.jsp.PageContext pageContext, String var, Object value) {
+        if (value == null) {
+            pageContext.removeAttribute(var);
+        } else {
+            pageContext.setAttribute(var, value);
+        }
+    }
+
```
markt-asf, line 968: This only appears to be used in test code. If true, it should be removed.

java/org/apache/jasper/compiler/Generator.java
```diff
@@ -3028,6 +3036,195 @@ public String generateNamedAttributeJspFragment(Node.NamedAttribute n, String ta
...
+            out.pushIndent();
+            // Generated body of method
+            String varValue = jspAttributes.get("var").getValue();
+            JspAttribute scope = jspAttributes.get("scope");
+            if (scope == null) {
+                out.printil("pageContext.removeAttribute(\"" + varValue + "\");");
+            } else {
+                String scopeValue = translateScopeToConstant(jspAttributes);
+                out.printil("pageContext.removeAttribute(\"" + varValue + "\", " + scopeValue + ");");
+            }
```
markt-asf, line 3218: This takes a different approach to handling `null` scope compared to `set`. It isn't wrong but it is inconsistent. Why? If there is a good reason, it should be documented else folks coming later are going to wonder why. If there isn't a good reason, `set` and `remove` should be consistent (and I have no preference for either approach).

markt-asf (changes requested):
Thinking long term, I am wondering whether we need to provide the ability for users to customise these optimisations. If we do then we can add something like a `GeneratorFactory` and users can  provide their own generator that overrides the built-in one along the lines of `ELInterpreterFactory`. For now, I don't think is necessary.

java/org/apache/jasper/runtime/JspRuntimeLibrary.java
jengebr: The generated JSP code calls this method, why is why it's not called within Tomcat itself.

java/org/apache/jasper/compiler/Generator.java
jengebr: I modeled the code here after the tag implementations (SetTag and RemoveTag), with the goal of maintaining compatibility... so any mismatch is either an error in how I copied the logic, or how the tags themselves are implemented.  What specific differences do you see?

java/org/apache/jasper/runtime/JspRuntimeLibrary.java
markt-asf: Can you point out where. I only see calls to the 4-arg method, not the 3-arg method.

java/org/apache/jasper/compiler/Generator.java
markt-asf: `set` calls
```
String scopeValue = translateScopeToConstant(jspAttributes);
```
whereas `remove` treats `scope == null` as a special case
```
if (scope == null) {
  out.printil("pageContext.removeAttribute(\"" + varValue + "\");");
} else {
    String scopeValue = translateScopeToConstant(jspAttributes);
    out.printil("pageContext.removeAttribute(\"" + varValue + "\", " + scopeValue + ");");
}
```
I'd probably remove the special case treatment from `remove`.

markt-asf (changes requested):
I'm fine with `GeneratorFactory` being something we consider later if necessary. Just want to make sure we don't do anything now to make it harder to do that if we need to.

java/org/apache/jasper/runtime/JspRuntimeLibrary.java
jengebr: Thank you, sorry, I missed the 3-arg vs. 4-arg distinction.  You're correct and I'll remove the 3-arg.

java/org/apache/jasper/compiler/Generator.java
markt-asf: Understood. I didn't realise there was a difference until I looked at the `PageContext` Javadoc. It is probably worth a short comment on both set and remove on how no/null scope is handled since it is different.

markt-asf (approved):

PR #843, opened 2025-04-14
BUGFIX: fix an incorrect range size validation

Chenjp:
**semantics fix:** range size = end_pos - start_pos + 1

## See error log
```bash
curl http://localhost:64438/one.txt -d c -H "Content-Range: bytes 0-0/1" -i
HTTP/1.1 405
Allow: OPTIONS, GET, HEAD, PUT, DELETE
Content-Type: text/html;charset=utf-8
Content-Language: zh-CN
Content-Length: 637
Date: Mon, 14 Apr 2025 08:44:27 GMT
```


rmaucher:
Ok so this looks like it cleans up all the off by one issue. Merged.

PR #844, opened 2025-04-23
not recycle when external

qingdaoheze:
Fix for https://bz.apache.org/bugzilla/show_bug.cgi?id=69655


markt-asf:
Does not fix the issue (which has already been fixed).

qingdaoheze:
You can see my screenshots. I think the root cause is that the length of org.apache.coyote.http2.Http2 Protocol#recycledRequestsAndResponses gradually increases when the upgrade request comes.

image

PR #849, opened 2025-05-10
improve log context in order to add thread name information.

felix-zhang-0818:


felix-zhang-0818:
From source code level.  
if (isRequestThread(thread)) {
    log.warn(sm.getString("webappClassLoader.stackTraceRequestThread", getContextName(), threadName,
            getStackTrace(thread)));
}

it already provides the parameter threadName, but the log context is not using it.  So here, it's enough to just apply log context properties change.

PR #850, opened 2025-05-12
improve log context to add thread name for webappClassLoader.stackTra…

felix-zhang-0818:
…ceRequestThread


markt-asf:
There is a comment at the top of all of the non-English resource files that states:
```
# Do not edit this file directly.
# To edit translations see: https://tomcat.apache.org/getinvolved.html#Translations
```
If you do decide to ignore that comment, you need to manually update POEditor.com else your changes will get overwritten the next time the resource files are sync'd.

I have applied these updated to POEditor.com

ChristopherSchultz:
Oh, sorry @markt-asf I forgot about POEditor.com. I'll coordinate off-line about this.

PR #854, opened 2025-05-12
Refactor: Use enum for parsingRequestLinePhase in Http11InputBuffer

devjohnpark:
## Refactor `parsingRequestLinePhase` in `Http11InputBuffer` to use `enum RequestLineParsePhase`

*   Changed the type of the `parsingRequestLinePhase` field from `int` to the `enum RequestLineParsePhase`.
*   This is a pure refactoring and introduces no functional changes.
*   All tests pass.

#### Motivation for Replacing the int-typed parsingRequestLinePhase with an Enum

1.  Readability
Using the `enum RequestLineParsePhase` instead of `int` makes the purpose of the `parsingRequestLinePhase` variable much clearer. Developers can immediately understand the current parsing state (e.g., `RequestLineParsePhase.METHOD`) just by looking at the enum constant.

2. Type Safety
The `enum` type ensures that only valid `RequestLineParsePhase` constants can be assigned to the `parsingRequestLinePhase` variable. The compiler prevents accidental assignment of arbitrary or invalid integer values (e.g., `9`, which is outside the range of valid phase values), which could lead to runtime errors if the variable were still an `int`.

#### Summary of Changes

*   The type of the `parsingRequestLinePhase` field was changed from `int` to `enum RequestLineParsePhase`.
*   To ensure compatibility with the original integer values, each `enum` constant explicitly retains its corresponding integer representation via an internal `int phase` field.
*   The existing `public int getParsingRequestLinePhase()` method is still used to provide the phase value to `Http11Processor`. To support this gradually and maintain API compatibility for now, the `int phase` field within `enum RequestLineParsePhase` was kept. The `getParsingRequestLinePhase()` method now returns the `phase` value from the current enum constant.
*   **Future Intention:** If this PR is merged, a potential follow-up task could be to remove the `getParsingRequestLinePhase()` method and provide more specific, intention-revealing methods for `Http11Processor`, further improving readability and decoupling.


n828cl:
You can't seriously think that someone is going to even review, let alone merge, a PR with 3,271 files changed...

Seems like a rebase is warranted first.

PR #855, opened 2025-05-13
[Bug 69693] - Improve readability and type safety by changing parsingRequestLinePhase to an enum in Http11InputBuffer 

devjohnpark:
## Refactor `parsingRequestLinePhase` in `Http11InputBuffer` to use `enum RequestLineParsePhase`

*   Changed the type of the `parsingRequestLinePhase` field from `int` to the `enum RequestLineParsePhase`.
*   This is a pure refactoring and introduces no functional changes.
*   All tests pass.

#### Motivation for Replacing the int-typed parsingRequestLinePhase with an Enum

1.  Readability
Using the `enum RequestLineParsePhase` instead of `int` makes the purpose of the `parsingRequestLinePhase` variable much clearer. Developers can immediately understand the current parsing state (e.g., `RequestLineParsePhase.METHOD`) just by looking at the enum constant.

2. Type Safety
The `enum` type ensures that only valid `RequestLineParsePhase` constants can be assigned to the `parsingRequestLinePhase` variable. The compiler prevents accidental assignment of arbitrary or invalid integer values (e.g., `9`, which is outside the range of valid phase values), which could lead to runtime errors if the variable were still an `int`.

#### Summary of Changes

*   The type of the `parsingRequestLinePhase` field was changed from `int` to `enum RequestLineParsePhase`.
*   To ensure compatibility with the original integer values, each `enum` constant explicitly retains its corresponding integer representation via an internal `int phase` field.
*   The existing `public int getParsingRequestLinePhase()` method is still used to provide the phase value to `Http11Processor`. To support this gradually and maintain API compatibility for now, the `int phase` field within `enum RequestLineParsePhase` was kept. The `getParsingRequestLinePhase()` method now returns the `phase` value from the current enum constant.
*   **Future Intention:** If this PR is merged, a potential follow-up task could be to remove the `getParsingRequestLinePhase()` method and provide more specific, intention-revealing methods for `Http11Processor`, further improving readability and decoupling.

### DIFF
--- a/java/org/apache/coyote/http11/Http11InputBuffer.java
+++ b/java/org/apache/coyote/http11/Http11InputBuffer.java
@@ -146,7 +146,7 @@ public class Http11InputBuffer implements InputBuffer, ApplicationBufferHandler,
         lastActiveFilter = -1;
 
         parsingRequestLine = true;
-        parsingRequestLinePhase = RequestLineParsePhase.NEW;;
+        parsingRequestLinePhase = RequestLineParsePhase.PHASE_1_NEW;;
         parsingRequestLineEol = false;
         parsingRequestLineStart = 0;
         parsingRequestLineQPos = -1;
@@ -253,7 +253,7 @@ public class Http11InputBuffer implements InputBuffer, ApplicationBufferHandler,
 
         chr = 0;
         prevChr = 0;
-        parsingRequestLinePhase = RequestLineParsePhase.NEW;;
+        parsingRequestLinePhase = RequestLineParsePhase.PHASE_1_NEW;;
         parsingRequestLineEol = false;
         parsingRequestLineStart = 0;
         parsingRequestLineQPos = -1;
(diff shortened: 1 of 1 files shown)


YongGoose:
Personally, I think this PR is worth reviewing!
Using enums is definitely easier to understand than using integers like `0` and `1`.

markt-asf:
I'm not convinced. With `int` the order is clear and obvious. With the proposed `enum` the correct order is less clear and depends on the reader knowing the detail of the HTTP/1.1 spec for request lines. Better names for the constants (that explicitly include the order) might help.

YongGoose:
> I'm not convinced. With `int` the order is clear and obvious. With the proposed `enum` the correct order is less clear and depends on the reader knowing the detail of the HTTP/1.1 spec for request lines. 

I hadn’t considered it from the order perspective — your point makes a lot of sense.

> Better names for the constants (that explicitly include the order) might help.

That’s a great idea!

YongGoose:
@markt-asf 
This is just a simple question out of curiosity — doesn’t the Tomcat project use Copilot for code reviews?

I’m a committer for the Seata project, and I’ve found that Copilot does a pretty good job of catching typos or awkward grammar. 👍🏻
- https://github.com/apache/incubator-seata/pull/7354#discussion_r2098160942

For simple refactoring tasks like this one, it’s actually been quite helpful.
Have a great day :)

devjohnpark:
@markt-asf I agree that explicitly including the order in enum constant names, like PHASE_0_NEW, enhances clarity without requiring HTTP/1.1 spec knowledge. So I updated RequestLineParsePhase names to PHASE_1_NEW, etc. style according to feedback.

Additionally, how about I refactor the fields and methods needed for parsing the Request Line into an inner or outer class to separate the responsibilities?

rmaucher:
-1, let's keep an int instead. It works and nobody will change this code at this point. If they do, they should have a good understanding of it.

PR #857, opened 2025-05-17
[fix] ensure that when acceptor is stopped it doesn't log an error for nothing

rmannibucau:
Proposal to have a clean shutdown (follow up of the mail on the list).

This only handles the case bindOnInit=false.

### DIFF
--- a/java/org/apache/tomcat/util/net/NioEndpoint.java
+++ b/java/org/apache/tomcat/util/net/NioEndpoint.java
@@ -235,9 +235,9 @@ public class NioEndpoint extends AbstractNetworkChannelEndpoint<NioChannel,Socke
                 }
             }
         } else {
-            serverSock = ServerSocketChannel.open();
-            socketProperties.setProperties(serverSock.socket());
             InetSocketAddress addr = new InetSocketAddress(getAddress(), getPortWithOffset());
+            serverSock = ServerSocketChannel.open(addr.getAddress().getAddress().length == 4 ? StandardProtocolFamily.INET : StandardProtocolFamily.INET6);
+            socketProperties.setProperties(serverSock.socket());
             serverSock.bind(addr, getAcceptCount());
         }
         serverSock.configureBlocking(true); //mimic APR behavior


markt-asf:
This appears to be addressing a symptom (various error messages) rather than the cause (no local address available to unlock the `Acceptor`). We should be looking at the root cause. What connector configuration are you using? Any idea why `localAddress` is `null`?

rmannibucau:
@markt-asf http11nio protocol handler, the root cause is what i mentionned in my mail to the list, the acceptor thread doesn't handle the fact stop is being called properly so when server channel is closed it consider the endpoint is still running so it is an issue whereas it is actually desired to stop the thread so logging was abusive in that shutdown case IMHO, the shutdown lifecycle was just not handled by acceptor event loop

rmannibucau:
@markt-asf agree it fails similarly with `bindOnInit` or not set:

here is a reproducer:

```
package com.github.rmannibucau;

import org.apache.catalina.LifecycleException;
import org.apache.catalina.LifecycleState;
import org.apache.catalina.core.StandardContext;
import org.apache.catalina.core.StandardHost;
import org.apache.catalina.startup.Tomcat;
import org.apache.catalina.valves.ErrorReportValve;
import org.apache.tomcat.util.modeler.Registry;

import java.io.IOException;
import java.net.URI;
import java.net.http.HttpClient;
import java.net.http.HttpRequest;
import java.nio.file.Files;
import java.nio.file.Path;

import static java.net.http.HttpResponse.BodyHandlers.discarding;

// leads to:
// WARNING: The acceptor thread [http-nio-8080-Acceptor] did not stop cleanly
//
// tested with java 17 and 24 with maven dep (+transitives):
//     <dependency>
//      <groupId>org.apache.tomcat</groupId>
//      <artifactId>tomcat-catalina</artifactId>
//      <version>11.0.7</version>
//    </dependency>
public final class StartAndStopTomcat {
    private StartAndStopTomcat() {
        // no-op
    }

    public static void main(final String... args) throws LifecycleException, InterruptedException, IOException {
        Registry.disableRegistry();

        // setup "Tomcat"
        final var tomcat = new Tomcat();
        tomcat.setBaseDir("target/work");
        Files.createDirectories(Path.of("target/work"));
        tomcat.setPort(8080);

        final var host = new StandardHost();
        host.setAutoDeploy(false);
        host.setName("localhost");
        tomcat.getEngine().addChild(host);

        try {
            tomcat.init();
        } catch (final LifecycleException e) {
            try {
                tomcat.destroy();
            } catch (final LifecycleException ex) {
                // no-op
            }
            throw new IllegalStateException(e);
        }
        try {
            tomcat.start();
        } catch (final LifecycleException e) {
            // "stop"
            throw new IllegalStateException(e);
        }

        // tomcat.getConnector().setProperty("bindOnInit", "false");
        // or next line which just enables to trigger the init of the connector:
        tomcat.getConnector();

        // "create a context
        final var context = new StandardContext();
        context.setPath("");
        context.setName("");
        context.setFailCtxIfServletStartFails(true);

        context.addLifecycleListener(new Tomcat.FixContextListener());

        final var errorReportValve = new ErrorReportValve();
        errorReportValve.setShowReport(false);
        errorReportValve.setShowServerInfo(false);
        context.getPipeline().addValve(errorReportValve);

        // no need of all these checks in general since we use a flat classpath
        context.setClearReferencesThreadLocals(false);
        context.setClearReferencesRmiTargets(false);
        context.setClearReferencesHttpClientKeepAliveThread(false);
        context.setClearReferencesStopThreads(false);
        context.setClearReferencesStopTimerThreads(false);
        context.setSkipMemoryLeakChecksOnJvmShutdown(true);

        tomcat.getHost().addChild(context);

        // ensure it is started
        final var state = context.getState();
        if (state == LifecycleState.STOPPED || state == LifecycleState.FAILED) {
            // "stop"
            throw new IllegalStateException("Context didn't start");
        }

        // do a request to use the acceptor
        System.out.println("Started");
        try (final var client = HttpClient.newHttpClient()) {
            client.send(HttpRequest.newBuilder().GET().uri(URI.create("http://localhost:8080")).build(), discarding());
        }

        //
        // stop
        //
        System.out.println("Stopping");
        final var server = tomcat.getServer();
        tomcat.stop();
        tomcat.destroy();
        if (server != null) { // give a change to stop the utility executor otherwise it just leaks and stop later
            final var utilityExecutor = server.getUtilityExecutor();
            if (utilityExecutor != null) {
                utilityExecutor.close();
            }
        }
    }
}
```

markt-asf:
That test case doesn't log any exceptions or warnings for me with Tomcat 12.0.x.

rmannibucau:
@markt-asf did it using "one line formatter" and only copied from the http request reception to the stop state to not flood with useless data: https://gist.githubusercontent.com/rmannibucau/1e30a6506f5d23ec3cb58139b9986aac/raw/13db877bb89c558d1f3220ecda7fbb36ae55ec5b/gistfile1.txt

markt-asf:
Tx. That helps. The unlock is trying to use an IPv6 address to unlock the Acceptor which Tomcat doesn't appear to be listening on. I'll see if I can reproduce this but if I can't, I'll add some debug logging to `AbstractEndpoint.getUnlockAddress()` so we can figure out what is going wrong.

rmannibucau:
Interesting, so adding `-Djava.net.preferIPv4Stack=true` i do not have the error anymore and indeed the error seems to come from the fact i'm using "localhost" so i enter into `localAddress.getAddress().isAnyLocalAddress()`. Alternatively i can also workaround it by setting before init/start `tomcat.getConnector().setProperty("address", "localhost");` but both are not aligned on my computer.

markt-asf:
Just downloading the iso to try and reproduce this. Any networking configured beyond the default using DHCP such as a VPN?

rmannibucau:
@markt-asf no vpn, just a plain default setup with this /etc/hosts:

```
127.0.0.1 localhost local.dev
127.0.1.1 romain

# The following lines are desirable for IPv6 capable hosts
::1     localhost ip6-localhost ip6-loopback
fe00::0 ip6-localnet
ff00::0 ip6-mcastprefix
ff02::1 ip6-allnodes
ff02::2 ip6-allrouters
```

rmannibucau:
@markt-asf any reason to not enforce family in org.apache.tomcat.util.net.NioEndpoint#initServerSocket based on addr ? updated the PR in that sense

markt-asf:
May have unintended consequences for dual-stack systems.

markt-asf:
I'm very reluctant to make changes until I understand the root cause of an issue - especially in an area such as networking where there can be subtle behaviour differences across platforms.

I have Ubuntu 25.04 installed and IPv6 enabled but I still don't see the issue you are seeing. Can you email me the output of `ifconfig` for your system? I want to see how your setup compares to mine so I can hopefully tweak my config to recreate the issue you are seeing.

markt-asf:
Thanks for the `ifconfig` output. Configuration seems very similar. I installed Docker in case that was a factor but that didn't help trigger the issue. I still can't recreate this issue.

Tomcat appears to be finding a valid IP address to unlock the Acceptor for your system.

Given that it works when you force IPv4, might there be some configuration somewhere (`iptables`?) blocking IPv6 connections? If you start Tomcat and `netstat` shows it is listening on IPv6, can you connect to each of the IPv6 addresses on port 8080 with a browser?

At this point, this looks more like an environmental issue than a Tomcat issue.

rmannibucau:
@markt-asf you nailed it, after way deeper investigation I realized ipv6 was bound but only ipv4 (all) were working. It led me to checking some conf os. gai.conf was plain default so nothing special there (all commented actually) but realized `/proc/sys/net/ipv6/conf/all/disable_ipv6` was enabled.
Would it be ok to try to read this file if on linux and if readable and "1" not respect the address class in getUnlockAddress?
My goal would be to have something working by default without having to do the configuration of all the tomcats (someties hidden) on the system.

markt-asf:
This is almost certainly an issue with the system being tested.

I am still unable to recreate the reported issue.

If I enable `/proc/sys/net/ipv6/conf/all/disable_ipv6` then the IPv6 addresses no longer appear in `ifconfig`. I can even start Tomcat with IPv6 enabled, disable IPv6, stop Tomcat and Tomcat stops cleanly.

There is an error message when this situation is detected. Debug level logging is provided that provides sufficient information to debug further.

Apart from exceptional circumstances, Tomcat does not attempt to work around bugs in other components or OS configuration issues. This is a design decision to avoid bloat.

I'll leave this open for now but unless some evidence emerges in the next few days that Tomcat is somehow at fault, this is going to be closed as "Not a Tomcat issue".

rmannibucau:
> Apart from exceptional circumstances, Tomcat does not attempt to work around bugs in other components or OS configuration issues. This is a design decision to avoid bloat.

well this is not an OS bug there but more a JVM bug since ipv6 should never be used for anything but it is there
since tomcat clearly workarounds the JVM bugs I think it should there as well, i'm not sure there is another way using java api to check this setting, if so i'm happy with that, else reading a file is not crazy and avoid warnings/errors which are hard to track and polluting shutdown for nothing so worth it IMHO

markt-asf:
Looks like an OS issue (or more likely a network driver issue) since `ifconfig` is reporting IPv6 addresses for a network adaptor when IPv6 is disabled. I don't see that on my Ubuntu 25.04 install.

rmannibucau:
@markt-asf not really, the issue is really on tomcat side and its any handling at a time which is decoralated from the binding time. If using a specific interface there is no issue but with multiple interfaces tomcat just randomly assumes the family. The disable setting is something you can set at runtime so tomcat really has a bug between the bind logic and the shutdown logic there even if I can agree it is probably something rare (that said with kubernetes or net related config it is not rare this is a setting is also customized after startup with a (systemctl) service.
So still think tomcat shouldn't just think it is fine but either collect the shutdown ip earlier (just after bind), enforce the family since java will do it anyway and it solves all issues or just hide the warn/error when shutting down since it knows it is fine cause the thread is no more alive anyway.

markt-asf:
On my clean install, Tomcat correctly handles disabling of IPv4 via `/proc/sys/net/ipv6/conf/all/disable_ipv6` at runtime.

All the indications are that an issue with your environment that is causing the OS to report that an IPv6 interface is available when it is not. Tomcat correctly reports this as an error condition.

PR #859, opened 2025-05-22
Acceptor & Poller Thread should print error logs when it crashes now

jmilktea:
As the title says, this is to facilitate troubleshooting. When the core thread crashes, we need to print the error log.


rmaucher:
I don't understand what this tries to do. If ExceptionUtils.handleThrowable rethrows a fatal error, you will notice it.

jmilktea:
Yes sir, the service will hang, but there is no log to help me find it quickly. I need to find it through some other means. If there is a log, I can find it at a glance, and it is reasonable to print the error log. @rmaucher

PR #860, opened 2025-05-26
Fix issue on multiple path parameters processing in RequestUtil

Chenjp:
RequestUtil#stripPathParams: enable support for multiple k=v pairs


markt-asf:
Thanks for the PR. I merged this manually as I wanted to modify the variable naming slightly and add a couple of additional tests.

rmaucher:
I verified this should not cause problems with the rewrite code.

PR #862, opened 2025-05-29
BZ 69699 - Propagate Session ID path parameter during rewrite redirect

Chenjp:
redirect target url: append session id path parameter

### DIFF
--- a/java/org/apache/catalina/valves/rewrite/RewriteValve.java
+++ b/java/org/apache/catalina/valves/rewrite/RewriteValve.java
@@ -428,19 +428,29 @@ public class RewriteValve extends ValveBase {
                     StringBuilder urlStringEncoded =
                             new StringBuilder(REWRITE_DEFAULT_ENCODER.encode(urlStringRewriteEncoded, uriCharset));
 
-                    // Currently session ID is the only propagated path parameter
+                    // Per servlet spec#url-rewriting, when a client will not accept a cookie, URL rewriting may be used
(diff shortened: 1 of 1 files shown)


rmaucher:
Thanks. Fixing this did not seem to be a bad idea, but the proposed patch seemed too convoluted. I did not add the test since encodeRedirectURL requires a real session (and it is already tested elsewhere).

Chenjp:
> Thanks. Fixing this did not seem to be a bad idea, but the proposed patch seemed too convoluted. I did not add the test since encodeRedirectURL requires a real session (and it is already tested elsewhere).

@rmaucher  per servlet spec, When server support Session Tracking by COOKIE, and client browser does send cookie, it is not suitable to sends the encodeRedirectURL with path parameter jsessionid.

markt-asf:
@Chenjp You should look carefully at the implementation of `encodeRedirectURL()` before you comment  further.

rmaucher:
Servlet containers have to implement that (very complex) helper method that does everything to achieve session tracking, depending on how the session tracking was sent by the client, and the webapp configuration. Yes, I kinda forgot about it too.

Chenjp:
> @Chenjp You should look carefully at the implementation of `encodeRedirectURL()` before you comment further.

en...Current implementation checks session id source(COOKIES or PathParameter), what SHALL we do if a client browser send a request with jsessionid in path parameter and other COOKIES except JSESSIONID? Apparently the client does support COOKIES.

PR #864, opened 2025-06-04
fix arbitrary file access during archive extraction zipslip

cur1pro:
https://github.com/apache/tomcat/blob/d54e8e0a4c52e6f00472a45d8f51ae141d9df5c0/java/org/apache/tomcat/util/scan/JarFileUrlJar.java#L192-L192

https://github.com/apache/tomcat/blob/d54e8e0a4c52e6f00472a45d8f51ae141d9df5c0/java/org/apache/tomcat/util/scan/JarFileUrlJar.java#L192-L192

Fix the issue, we need to ensure that paths derived from `entry.getName()` are validated to prevent directory traversal attacks. This involves:
1. Normalizing the path using `java.nio.file.Path.normalize()` or `java.io.File.getCanonicalFile()` to resolve any `../` or similar elements.
2. Verifying that the normalized path starts with the intended base directory using `java.nio.file.Path.startsWith()`.

The fix should be applied in the `JarFileUrlJar` class where `entry.getName()` is used, and in the `JspCServletContext` class where `entryName` is processed. Specifically:
- In `JarFileUrlJar`, sanitize the output of `entry.getName()` before returning it in `getEntryName()`.
- In `JspCServletContext`, validate `entryName` before adding it to `thePaths`.

### DIFF
--- a/java/org/apache/jasper/servlet/JspCServletContext.java
+++ b/java/org/apache/jasper/servlet/JspCServletContext.java
@@ -404,7 +404,12 @@ public class JspCServletContext implements ServletContext {
                             int sep = entryName.indexOf('/', jarPath.length());
                             if (sep < 0) {
                                 // This is a file - strip leading "META-INF/resources"
-                                thePaths.add(entryName.substring(18));
+                                String sanitizedEntryName = entryName.substring(18);
+                                File testFile = new File(basePath, sanitizedEntryName);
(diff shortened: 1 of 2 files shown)


markt-asf:
Potential security vulnerabilities **DO NOT** belong in public pull requests.

Guidance on how to report security vulnerabilities responsibly can be found at:
https://tomcat.apache.org/security.html

Fortunately, this report is invalid so there is no harm done by the irresponsible reporting.

PR #865, opened 2025-06-05
Enhance isEncodeable to determine putting session id in URL using path parameter

Chenjp:
It is not suitable to encode session id in URL if client support cookie.
per discuss in #862


markt-asf:
This change is unnecessary. If cookies are supported, the original session will have been created with a tracking cookie.

Irrespective of the above, if the client has, for whatever reason, opted not to use session cookies and prefers to use URL tracking Tomcat should respect that. If an application doesn't want to respect that, it can disable URL tracking.

PR #867, opened 2025-06-11
Optimize Request#getCharsetHolder to avoid repeated parsing when charset is null

StudentGu:
In org.apache.coyote.Request#getCharsetHolder, when charsetHolder.getName() returns null, parsing is performed repeatedly, resulting in unnecessary performance overhead. The following optimizations are proposed:

1. Initialize the org.apache.coyote.Request#charsetHolder member variable to null instead of CharsetHolder.EMPTY, so that null clearly indicates the unparsed state.
2. In getCharsetHolder, check if charsetHolder is null to determine whether parsing is needed, instead of checking charsetHolder.getName() == null.
3. Once charsetHolder is assigned (regardless of whether name is null), consider it parsed and avoid any further parsing.

This avoids repeated parsing when charsetHolder.getName() is null, improving performance. Please ensure thread safety and that assignment semantics remain consistent during implementation.

### DIFF
--- a/java/org/apache/coyote/Request.java
+++ b/java/org/apache/coyote/Request.java
@@ -139,7 +139,7 @@ public final class Request {
      */
     private long contentLength = -1;
     private MessageBytes contentTypeMB = null;
-    private CharsetHolder charsetHolder = CharsetHolder.EMPTY;
+    private CharsetHolder charsetHolder = null;
 
     /**
      * Is there an expectation ?
@@ -389,7 +389,7 @@ public final class Request {
     // -------------------- encoding/type --------------------
 
     public CharsetHolder getCharsetHolder() {
-        if (charsetHolder.getName() == null) {
+        if (charsetHolder == null) {
             charsetHolder = CharsetHolder.getInstance(getCharsetFromContentType(getContentType()));
         }
         return charsetHolder;
@@ -397,6 +397,10 @@ public final class Request {
 
 
     public void setCharsetHolder(CharsetHolder charsetHolder) {
+        if (charsetHolder == null || charsetHolder.getName() == null) {
+            this.charsetHolder = null;
+            return;
+        }
         this.charsetHolder = charsetHolder;
     }
 
@@ -724,7 +728,7 @@ public final class Request {
 
         contentLength = -1;
         contentTypeMB = null;
-        charsetHolder = CharsetHolder.EMPTY;
+        charsetHolder = null;
         expectation = false;
         headers.recycle();
         trailerFields.recycle();


morning-gu:
@markt-asf @michael-o Please take a look.

rmaucher:
Why would the charset name be null ?

morning-gu:
> Why would the charset name be null ?


Because it is not included in the request headers, as shown below:
‌Content-Type=application/x-www-form-urlencoded

The `Request#getCharsetHolder` method is too performance-intensive, as shown in the flame graph below.
image

morning-gu:
@rmaucher Please review and approve the workflows.
image

ChristopherSchultz:
@morning-gu Please don't ping specific members of this team to ask for a review. Everyone is aware that this PR is here, and it will be reviewed when these volunteers have time to review it.

java/org/apache/coyote/Request.java
```diff
@@ -389,14 +389,18 @@ public void setLocalPort(int port) {
...
             charsetHolder = CharsetHolder.getInstance(getCharsetFromContentType(getContentType()));
         }
         return charsetHolder;
     }
 
 
     public void setCharsetHolder(CharsetHolder charsetHolder) {
+        if (charsetHolder == null || charsetHolder.getName() == null) {
+            this.charsetHolder = null;
+            return;
```
ChristopherSchultz, line 402: Prefer `else` here to `return` inside the predicate.

morning-gu: Okay, I have resolved it, thank you.

ChristopherSchultz:
This analysis looks correct to me, and the solution is staightforward.

ChristopherSchultz:
What does your flame graph look like after this change?

morning-gu:
> What does your flame graph look like after this change?

Sorry, I don't have a flame graph right now. I can capture one tomorrow.

rmaucher:
I think it is likely ok. I did not like it since it is backwards from the original intent of the refactoring.

markt-asf:
Looks reasonable to me. Needs a change log entry.

morning-gu:
> Looks reasonable to me. Needs a change log entry.

Thank you for the feedback. I have added a changelog entry to webapps/docs/changelog.xml for this change.

morning-gu:
> What does your flame graph look like after this change?

After optimization, the consumption of `Request#getCharsetHolder` is only present during the first parsing. Subsequently, due to the caching of null values, this method's consumption is not visible on the flame graph.
image

PR #868, opened 2025-06-12
 recycle RequestInfo when Request is reused in pool

qingdaoheze:
The request statistics will be wrong when Request object is reused in the object pool.

### DIFF
--- a/java/org/apache/coyote/RequestInfo.java
+++ b/java/org/apache/coyote/RequestInfo.java
@@ -261,4 +261,23 @@ public class RequestInfo {
     public void setLastRequestProcessingTime(long lastRequestProcessingTime) {
         this.lastRequestProcessingTime = lastRequestProcessingTime;
     }
+
+    public void recycle() {
+        this.bytesSent = 0;
+        this.bytesReceived = 0;
+
+        // Total time = divide by requestCount to get average.
+        this.processingTime = 0;
+        // The longest response time for a request
+        this.maxTime = 0;
+        // URI of the request that took maxTime
+        this.maxRequestUri = null;
+
+        this.requestCount = 0;
+        // number of response codes >= 400
+        this.errorCount = 0;
+
+        // the time of the last request
+        this.lastRequestProcessingTime = 0;
+    }
 }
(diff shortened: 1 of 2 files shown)


rmaucher:
Are you trying to fix a problem you noticed, or is this PR based on code analysis ?

qingdaoheze:
Sorry. I don't understand what you mean. My issue is that the metrics in org.apache.coyote.RequestGroupInfo#removeRequestProcessor is not counted correctly when org.apache.coyote.http2.Http2Protocol#discardRequestsAndResponses is false(default). This PR can resolve this issue partially, because there is still some special scenarios that miss some count. I hope you can give an better solution.
@rmaucher

rmaucher:
Definitely this PR is wrong. RequestInfo is not the stats for an individual request, so it should not be reset.

markt-asf:
HTTP/2 is handled differently because of the multiplexing. We probably do need a `recycle()` method on `RequestInfo`. It looks like it should be called around line 440 (current 9.0.x code) of `StreamProcessor`.

qingdaoheze:
image

@rmaucher When the org.apache.coyote.Request pool is used, its org.apache.coyote.Request#reqProcessorMX is not recycled after the request is recycled. So when org.apache.coyote.RequestGroupInfo#removeRequestProcessor is run, its stat count is counted many times. So the result count in org.apache.coyote.RequestGroupInfo will be wrong.

rmaucher:
> HTTP/2 is handled differently because of the multiplexing. We probably do need a `recycle()` method on `RequestInfo`. It looks like it should be called around line 440 (current 9.0.x code) of `StreamProcessor`.

That looks correct, you know that code better than I do.

qingdaoheze:
> HTTP/2 is handled differently because of the multiplexing. We probably do need a `recycle()` method on `RequestInfo`. It looks like it should be called around line 440 (current 9.0.x code) of `StreamProcessor`.

@markt-asf  For this solution, there is also a concurrency problem. Because the request has already returned to the pool on the line 150 of StreamProcessor. Then the recycle method is executed on line 151 of StreamProcessor. In the progress, the request has been returned to the pool may be borrowed by another request and its reqProcessorMX may has been changed.
image

markt-asf:
Fixed in 4dc1a0f9 and back-ported

PR #869, opened 2025-06-19
Discard unsupported part header field while handling multipart fileupload request

Chenjp:
comply rfc7578 section [4.8](https://datatracker.ietf.org/doc/html/rfc7578#section-4.8)
```
4.8.  Other "Content-" Header Fields
   The multipart/form-data media type does not support any MIME header
   fields in parts other than Content-Type, Content-Disposition, and (in
   limited circumstances) Content-Transfer-Encoding.  Other header
   fields MUST NOT be included and MUST be ignored.
```
Only listed part header fields are supported. ignore others.


ChristopherSchultz:
I've got bad news: this code is a packaged-renamed copy of Apache commons-fileupload, so this PR actually should be re-worked and filed against that project instead. We are downstream of commons-fileupload.

The good news is that the changes should be almost identical, but the file paths will be different.

PR #870, opened 2025-06-19
Avoid to change the status of the response in StandardHostValve.java …

rmannibucau:
…if response is committed

main issue: you get a 500 in the access log whereas it was a 200 (for example)

### DIFF
--- a/java/org/apache/catalina/core/StandardHostValve.java
+++ b/java/org/apache/catalina/core/StandardHostValve.java
@@ -276,7 +276,7 @@ final class StandardHostValve extends ValveBase {
              * the status to 500 if an error status has not already been set and check for custom error-page for the
              * status.
              */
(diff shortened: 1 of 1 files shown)


markt-asf:
This is unnecessary. See https://github.com/apache/tomcat/blob/main/java/org/apache/catalina/connector/Response.java#L1220

rmannibucau:
@markt-asf hmm, can you detail how it is unecessary since it happens in async exchanges? :thinking:

markt-asf:
The check you've added is a duplicate of the first check that happens when the method in the `if` block is called.

rmannibucau:
@markt-asf hmm, you're right, wonder if it can be a timing issue so it would be helpful to set the "original status" in notes (or internal attributes) and check it in access log to not report a wrong status? any other idea?

PR #871, opened 2025-06-26
Enable rejectSuspiciousURIs by default to comply servlet 6 spec

Chenjp:
Make site safer, for those behind reverse proxy apps especially.

### DIFF
--- a/java/org/apache/catalina/connector/Connector.java
+++ b/java/org/apache/catalina/connector/Connector.java
@@ -291,7 +291,7 @@ public class Connector extends LifecycleMBeanBase {
     protected boolean useBodyEncodingForURI = false;
 
 
-    private boolean rejectSuspiciousURIs;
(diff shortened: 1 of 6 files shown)


markt-asf:
It was a deliberate decision not to reject the servlet defined suspicious URIs by default as it would introduce breakage for some users (there are users that use %2f in URIs for example).

Chenjp:
Rarely "suspicious" edge cases (e.g. having %2f in uri) deserve a manual explicit acceptance of suspicious URI, buy for potential security risks.

A great parts others should have opportunity to enjoy benefit from rejecting suspicious uri settings implicitly.

Regards,
Chenjp
________________________________

Chenjp:
@markt-asf servlet project contains a testcase (CanonicalUriPathTest.java), is it a part of servlet TCK?

PR #873, opened 2025-06-27
Filter to limit target file size of ranged put request

Chenjp:
see BZ 69446 and PR #823.
Target file maxSize is configurable now.


Chenjp:
Enable the opportunity of configurable max-range limitation.

Chenjp:
To reduce potential Storage abuse / DoS.

PR #876, opened 2025-07-28
Adding Tomcat favicon to the Tomcat Manager webapp

xsimo:
1. Tomcat Manager is sometimes run on server on which there is no ROOT context, therefore no default favicon is not found. This commits adds the tomcat favicon to the manager webapp directory and references it from the HTML_HEADER_SECTION in Constants.java


markt-asf:
Also forward-ported to 10.1.x, 11.0.x and 12.0.x

PR #877, opened 2025-07-30
Update RUNNING.txt

Petsku01:
Description:
This PR adds a new subsection (3.5) to `RUNNING.txt` under "(3.4) Using the setenv script" to provide platform-specific examples for configuring JVM memory options using `CATALINA_OPTS`. The example shows how to set initial and maximum heap sizes (-Xms512m -Xmx2048m) for Windows and *nix systems, with a note about system resource considerations. This improves setup clarity for users, especially in production environments, aligning with Tomcat’s goal of supporting mission-critical applications.

Changes:
- Added subsection (3.5) to `webapps/docs/RUNNING.txt` with `CATALINA_OPTS` memory configuration example.

Testing:
- Verified formatting in `RUNNING.txt` for consistency with existing style (plain text, no tabs).
- Confirmed the new section renders correctly in the docs webapp at http://localhost:8080/docs/RUNNING.txt.
- Tested the example `setenv` scripts to ensure Tomcat starts with the specified JVM options.

Related Issues:
- No specific Bugzilla issue, but addresses common user questions about memory configuration (seen in tomcat-users mailing list).


ChristopherSchultz:
Is there evidence that downstream users don't know how to set their JVM settings, or that heap settings are particularly worthy of documentation?

isapir:
Modern JVMs do a pretty good job at setting the memory values in the examples according to the specific system.  I personally find the `-Xms` flag to not be useful in modern JVMs, and the upper limit of `-Xmx` to be too constricting for many systems at 2GB

While I do think that a bit more information about `CATALINA_OPTS` can be useful, it can be added to the existing section 3.3 rather than adding a new section for it

I am also concerned that the example snippets would discard other settings that already exist in CATALINA_OPTS as they set the values without taking into account the fact that a value might already exist

ChristopherSchultz:
I see no reason to change `RUNNING.txt` at all. Anyone wanting to set CATALINA_OPTS should know what they are doing and don't need any further examples. There are dozens of JVM options and it does not make any sense to document any specific subset of them anywhere in Tomcat.

PR #878, opened 2025-07-31
9.0.x Host-Manager favicon.ico and Manager JSPs

xsimo:
as per Konstantin Kolinko suggestions [1], this pull request aims to correct and improve the previous #876.

1. Constants is re-used by the Host Manager web application, but the ico file had not been added to that application.
2. moving ico files to respective webapps/images directories.
3. Also adding the favicon link in the header of several JSPs in respective webapps/manager/WEB-INF/jsp directories.

[1] https://lists.apache.org/thread/ofxfdjf2jhz9owst0mf1pmk63rmcoz7m


markt-asf:
Thanks for this PR. I'll merge into 9.0.x, double check a few things and then forward port.

PR #880, opened 2025-08-02
BZ 69762 - integer overflow detection fixe enhance

Chenjp:
consider the possibility of integer overflow before result add.


markt-asf:
Thanks for spotting that. I'll back-port it to the other branches.

PR #885, opened 2025-08-27
Update logging.properties

trashz403:
Log levels set to INFO or WARNING to balance detail and noise in production.

Logs retained for 30 days to balance disk space usage and audit needs.

All logs use UTF-8 encoding, supporting internationalization.

Console logs limited to SEVERE to reduce noise for console viewers.

Comments included to guide when to increase log verbosity for troubleshooting.


markt-asf:
The default logging configuration is only ever going to be a starting point. We can't possibly expect to provide a default logging configuration that is suitable to the huge range of uses Tomcat is put to.

I see no good reason to make the suggested changes and several not to.

In addition this PR includes a number of unwanted changes such as removing examples and changing the license header.

PR #886, opened 2025-08-27
fix(session): Remove unreachable file existence check in FileStore#load

trashz403:
fix(session): Remove unreachable code in FileStore#load

The `if (!file.exists())` check within the try-with-resources block in the `load()` method is unreachable.

The `new FileInputStream()` constructor, which is executed before the try block's body, will throw a `FileNotFoundException` if the file does not exist. This prevents the conditional check from ever being reached, making it redundant code.

This commit removes the unnecessary check to simplify the method.


markt-asf:
The issue identified is valid but:
- this PR does not implement the proposed solution
- the proposed solution is not the best solution
- this PR introduces a number of unacceptable linear white space changes

PR #887, opened 2025-08-29
Add webapp scoped log capture helper and STRICT_SERVLET_COMPLIANCE web.xml parsing test

dsoumis:
1) Adds TomcatBaseTest.WebappLogCapture to  capture logs emitted under the webapp classloader during context start.
2) Add TestStrictComplianceDeployment for STRICT_SERVLET_COMPLIANCE web.xml parsing. Test whether a SAXException is present in the log when strict servlet compliance turned on in catalina.properties.


rmaucher:
Testing the log contents is not as stable as checking some other things, so it should be avoided if possible. However, sometimes it is useful so adding the capability to the testsuite seems like a good idea to me.

markt-asf:
Big +1 to expanding the coverage of the unit tests.
I was concerned about setting the system property impacting other tests but that is not an issue for tests run via Ant as each test class is tested in a separate JVM (by default and we haven't changed that). Tests run in the IDE will run in a single JVM but normally it is only single tests (or a package of tests) that are run so that won't normally be an issue.
My other concern is that there are two things being tested here. 1) that setting `STRICT_SERVLET_COMPLIANCE` enables XML validation and 2) that XML validation works as expected. I think it might be better to split that into two tests. One testing that the XML validation works (configured using the Context attributes) and one testing that setting `STRICT_SERVLET_COMPLIANCE` sets the attributes it is documented to set.

rmaucher:
Yes, this kind of use of a system property has been shown to cause testsuite unreliability issues in the past. This was indeed mostly a "fail if XML validation is enabled" test.

I think the utility that allows processing logs in the testsuite should be committed first.

dsoumis:
I have split the tests in two, one testing xml validation and the other STRICT_SERVLET_COMPLIANCE sets all the attributes as expected. Test will be skipped now, if the constants have already been initialized under the same JVM. I think this will avert any issues if the test suite is being run through IDE.

markt-asf:
LGTM. One minor nit is that continuation lines should be indented 8 spaces rather than 4.

PR #888, opened 2025-09-04
Add multiple cert/key to the sslcontext

jfclere:
That allows PQC and classical crypto to work with tomcat on the same connector.
Apache httpd does something similar.

### DIFF
--- a/java/org/apache/tomcat/util/net/AbstractEndpoint.java
+++ b/java/org/apache/tomcat/util/net/AbstractEndpoint.java
@@ -406,7 +406,7 @@ public abstract class AbstractEndpoint<S, U> {
                 sslHostConfig.setEnabledCiphers(sslUtil.getEnabledCiphers());
             } else {
                 /* with openssl we need to add the certificatte to the context */
-                sslUtil.addcertSSLContext(oldctx, certificate);
+                sslUtil.addSecondCertificate(oldctx, certificate);
             }
 
             SSLContext sslContext = certificate.getSslContext();
--- a/java/org/apache/tomcat/util/net/AbstractEndpoint.java
+++ b/java/org/apache/tomcat/util/net/AbstractEndpoint.java
@@ -397,12 +397,16 @@ public abstract class AbstractEndpoint<S, U> {
     protected void createSSLContext(SSLHostConfig sslHostConfig) throws IllegalArgumentException {
 
         boolean firstCertificate = true;
+        SSLContext oldctx = null;
         for (SSLHostConfigCertificate certificate : sslHostConfig.getCertificates(true)) {
             SSLUtil sslUtil = sslImplementation.getSSLUtil(certificate);
             if (firstCertificate) {
                 firstCertificate = false;
                 sslHostConfig.setEnabledProtocols(sslUtil.getEnabledProtocols());
                 sslHostConfig.setEnabledCiphers(sslUtil.getEnabledCiphers());
+            } else {
+                /* with openssl we need to add the certificatte to the context */
+                sslUtil.addcertSSLContext(oldctx, certificate);
             }
 
             SSLContext sslContext = certificate.getSslContext();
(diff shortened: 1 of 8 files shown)


rmaucher:
Ok so this needs some work (if this is done like that at least for OpenSSL, then definitely it should not be creating a second SSLContext), cleanup and testing, but ok. Thanks for testing that OpenSSL can work that way, it's easy. No idea about JSSE.

Ultimately, I still think Tomcat should be processing the client hello, find client supported groups, see what cert types are configured, and send to the appropriate SSLContext. This gives more control, but it's a lot more complex (we need group configuration, but here we clearly don't).

But I would say we can integrate this "hack" (although maybe it's not a hack, time will tell) for now.

rmaucher:
So I would be adding SSLUtil.addSecondCertificate for now (the name comes from the OpenSSL command line arguments descriptions), as a default method. Shouldn't be causing too many problems and we can stop using it later. It would only be used if SSLHostConfig is configured with *only TLS 1.3*.
I don't think we should add the MLDSA cert types and so on for now, just use some place holder for now until we figure it out (DSA or whatever).

jfclere:
> Ultimately, I still think Tomcat should be processing the client hello, find client supported groups, see what cert types are configured, and send to the appropriate SSLContext. This gives more control, but it's a lot more complex (we need group configuration, but here we clearly don't).

While looking in openssl there is no easy way to select the cert/key, ssl_callback_ClientHello() allows to check everything but I don't think we can choose the cert/key there... The cert/key are in the sslcontext.

rmaucher:
I cleaned it up a little, and verified it works (well, since mod_ssl does it ...). The change is not very intrusive overall. This still feels like a temporary solution at the moment though.

About the client hello processing: Tomcat does it in TLSClientHelloExtractor.

I skipped adding the specific certificate type at the moment since it's also a bit useless.

The config looks like (using DSA as a placeholder for "non RSA", and with TLS 1.3 being enabled by default - it could be reasonable to restrict this to a TLS 1.3 only SSLHostConfig):
```
    <Connector port="8454" protocol="HTTP/1.1"
               SSLEnabled="true" scheme="https" secure="true">
        <UpgradeProtocol className="org.apache.coyote.http2.Http2Protocol" />
        <SSLHostConfig>
            <Certificate certificateFile="${catalina.home}/conf/localhost-mldsa.crt"
               certificateKeyFile="${catalina.home}/conf/localhost-mldsa.key"
                         type="DSA" />
            <Certificate certificateFile="${catalina.home}/conf/localhost-rsa.crt"
               certificateKeyFile="${catalina.home}/conf/localhost-rsa.key"
                         type="RSA" />
        </SSLHostConfig>
    </Connector>
```

jfclere:
> I cleaned it up a little, and verified it works (well, since mod_ssl does it ...). The change is not very intrusive overall. This still feels like a temporary solution at the moment though.
> 
> About the client hello processing: Tomcat does it in TLSClientHelloExtractor.
> 
> I skipped adding the specific certificate type at the moment since it's also a bit useless.
> 
> The config looks like (using DSA as a placeholder for "non RSA", and with TLS 1.3 being enabled by default - it could be reasonable to restrict this to a TLS 1.3 only SSLHostConfig):
> 
> ```
>     <Connector port="8454" protocol="HTTP/1.1"
>                SSLEnabled="true" scheme="https" secure="true">
>         <UpgradeProtocol className="org.apache.coyote.http2.Http2Protocol" />
>         <SSLHostConfig>
>             <Certificate certificateFile="${catalina.home}/conf/localhost-mldsa.crt"
>                certificateKeyFile="${catalina.home}/conf/localhost-mldsa.key"
>                          type="DSA" />
>             <Certificate certificateFile="${catalina.home}/conf/localhost-rsa.crt"
>                certificateKeyFile="${catalina.home}/conf/localhost-rsa.key"
>                          type="RSA" />
>         </SSLHostConfig>
>     </Connector>
> ```

Hm I have used type="RSA" for the 2 cert/key, that works too, is that OK?

rmaucher:
After refactoring the certificate loading algorithm, this seemed nice enough to merge. I'm pretty sure this will eventually be removed, but we'll see.
Thanks for finding it in the mod_ssl code and testing, I would have done it using a much more complex technique (although possibly more JSSE friendly).
I added back the MLDSA certificate type.

PR #632, opened 2023-07-04
Add support for CIDR notation in `RemoteIpFilter`

fabien-chebel:
# Context

Tomcat's `RemoteIpFilter` currently allows configuring trusted/internal proxies using regexp.

When integrating with reverse proxies with a large number of IP addresses, regexp configuration gets cumbersome.

# Suggestion

I suggest adding support for IP ranges in CIDR notation to make it easier to setup the filter in these cases.

For backward compatibility, matching with masks is only performed when the trusted/internal proxies patterns are null.

Depending on the feedback I receive on this PR, I may add the same changes to Tomcat's `RemoteIpValve`.

PR #894, opened 2025-09-12
Add support for CIDR notation in RemoteIpFilter(#632)

ABin-Huang:
### CIDR Rule Storage Logic:
A dedicated netMaskSet storage mechanism is introduced to store the parsed CIDR network segment rules for 
internal proxies and trusted proxies, providing data support for subsequent IP verification.
### Proxy IP Verification Logic:
  When verifying an internal proxy IP, it first checks whether the IP conforms to the original regular expression format. If not, it 
  further verifies whether the IP falls within the stored CIDR network segments of internal proxies.
  When verifying a trusted proxy IP, it also first checks whether the IP matches the original regular expression format. If not, it 
  further confirms whether the IP is within the stored CIDR network segments of trusted proxies.


markt-asf:
Looks good. I'm just making a few changes locally:
 - fixing a few formatting nits
 -  mirroring the changes to the Valve
 -  adding the new attributes to the docs
 
 This is mostly done. I should be in a position to commit by Monday.

PR #897, opened 2025-09-18
Remove redundant conversion of applicationEventListenerList

OneEpitome:
Hello,

While reviewing the source code, I found a redundant conversion of `applicationEventListenerList` in `StandardContext.java`.
This pull request removes the unnecessary conversion.

Thank you for reviewing my contribution.


markt-asf:
It is done like that to support sub-classes that might have overridden the `getApplicationEventListeners()` method.

PR #900, opened 2025-09-23
typo: war-tracking -> war-tracker

mauk81:
https://github.com/apache/tomcat/blob/main/java/org/apache/catalina/startup/Constants.java#L37


markt-asf:
Better to update the docs than the code.

mauk81:
> Better to update the docs than the code.

I think I updated the doc (webapps/docs/config/host.xml). I just linked the code for reference.
I need a hint to the right place (another repository?) if that wasn't the doc.

markt-asf:
So you did.

PR #901, opened 2025-09-24
Access log nanoseconds

markt-asf:
Adding nanosecond resolution to the access log for 9.0.x is non-trivial because all the interface methods for access logging expect request duration to be passed in milliseconds.

The current state of 9.0.x is that some preparatory commits have been applied that should not change the current behaviour although they do add a little overhead.

I think we have three options:
1. Don't provide nanosecond resolution to the access logs for 9.0.x. Revert the preparatory commits.
2. Apply just the first of the commits in this PR. This is less invasive BUT changes nearly all of the various interface methods for access logging from milliseconds to nanoseconds. This might have backwards compatibility issues for some users although the most likely point of extension - custom access logging - is implemented so that they should be compatible.
3. Apply both commits in this PR. It is more invasive but should be backwards compatible for everyone.

Regardless of what we do with this PR, we have changed the meaning of request duration in the interface methods for access logging between 9.0.x and 10.1.x (actually 10.0.x). That decision has already been made so we just have to live with it.

If we adopt option 2 or 3, there will likely be a clean-up commit afterwards to mark various unused methods as deprecated in 9.0.x to 11.0.x and remove them in 12.0.x

### DIFF
--- a/java/org/apache/catalina/AccessLog.java
+++ b/java/org/apache/catalina/AccessLog.java
@@ -16,6 +16,8 @@
  */
 package org.apache.catalina;
 
+import java.util.concurrent.TimeUnit;
+
 import org.apache.catalina.connector.Request;
 import org.apache.catalina.connector.Response;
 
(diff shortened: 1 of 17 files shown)


ChristopherSchultz:
Mark,

Replying on-list because it's not really a part of the conversation.

Should we post this to users@ as well? This is probably far more 
relevant to users of Tomcat than just to those who follow the dev@ 
mailing list.

-chris

funkman:
(related to %F) Is it worth tweaking FirstByteTimeElement (or similar) to allow/expose nano resolution since that also seems disruptive?

markt-asf:
> Should we post this to users@ as well?

Good idea - I'll post something.

> Is it worth tweaking FirstByteTimeElement (or similar)...

`%F` is milliseconds in httpd. We can always add options to expose microseconds or nanoseconds if there is demand (for all versions). Similarly, if there was demand to align %T and %D in 9.0.x as well, we could make that configurable.

johnthad:
I have no need for this change.

HoffmannTom:
Support in version 10 onwards would be sufficient, imho.
The use cases are quite limited.

rmaucher:
Implementing this in Tomcat 9 is a lot more complicated than what I expected !

markt-asf:
> Implementing this in Tomcat 9 is a lot more complicated than what I expected !

It would be worth double checking my approach in case I missed a simpler solution.

Poojashree123456:
@markt-asf 
Implementing this in Tomcat 10 and above would be sufficient.

@johnthad 
Having the option of precise timestamps in logs would be helpful for detailed debugging and performance analysis.

jfclere:
System.nanoTime() won't slow down the whole server?

markt-asf:
> System.nanoTime() won't slow down the whole server?

No. Testing in main showed a small performance improvement.

markt-asf:
It is looking like option 1 is the preferred option so far. I'l leave it until after the weekend before taking any action to give folks more time to comment if they wish.

I will note that option 1 is the "safe" option in that a) there is no risk of a regression and b) we can also revisit this decision and implement option 2 or 3 at some point in the future if strong demand for nanosecond resolution in 9.0.x's access logs emerges.

markt-asf:
Proceeding with option 1.

PR #903, opened 2025-09-25
refactor: formatter refactoring

minjund:
Description
---
This PR made the following two modifications

- Unify constant delimiters
- Refactoring for better readability

Explanation
---
**Unify constant delimiters**
Constants managed in the Constants class in the coyote package are separated by _ for better readability.

globalCacheSize -> GLOBAL_CACHE_SIZE
localCacheSize -> LOCAL_CACHE_SIZE

**Refactoring for better readability**
The OneLineFormatter class format() method contains thread logic. I think separating it into its own method would make the code easier to read. I also refactored the thread output into the JsonFormatter class, which extend from OneLineFormatter, by declaring it as protected.

### DIFF
--- a/java/org/apache/juli/JsonFormatter.java
+++ b/java/org/apache/juli/JsonFormatter.java
@@ -65,14 +65,7 @@ public class JsonFormatter extends OneLineFormatter {
 
         // Thread
         sb.append("\"thread\": \"");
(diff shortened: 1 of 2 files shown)


minjund:
This is my first PR, so please let me know if there's anything wrong!

PR #904, opened 2025-10-02
add .webmanifest mime mapping

NicolaIsotta:
For reference:

- https://www.w3.org/TR/appmanifest/
- https://w3c.github.io/manifest/
- [https://www.iana.org/assignments/media-types/application/manifest+json](https://www.iana.org/assignments/media-types/application/manifest+json)

This PR targets branch 9.0.x, let me know if I need to provide PRs for the other branches too.

### DIFF
--- a/conf/web.xml
+++ b/conf/web.xml
@@ -4336,6 +4336,10 @@
         <mime-type>video/webm</mime-type>
     </mime-mapping>
     <mime-mapping>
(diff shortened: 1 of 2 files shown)


markt-asf:
Tomcat generally keeps its file extension to mime-type mapping aligned with httpd. This looks like something to return to when the specification is more stable. For now, applications that need this can add it in their web.xml.

PR #914, opened 2025-11-01
Add configurable compression

Long9725:
Make compression pluggable and configurable (gzip level/buffer), retain API compatibility
  
## Overview
  
  - Introduces configurable gzip compression (level, buffer size) and a pluggable output filter factory for HTTP/1.1. Preserves the original API to maintain backward compatibility.
  
## Changes
  
  - CompressionConfig: add gzipLevel and gzipBufferSize with validation; add configurable noCompressionEncodings; extend useCompression to accept an encoding; restore the original useCompression(Request, Response) overload
  delegating to gzip.
  - AbstractHttp11Protocol: expose getters/setters for gzip settings; add OutputFilterFactory support (by instance and by class name with error handling); delegate compression checks using the factory’s encoding; expose get/
  setNoCompressionEncodings.
  - http11.filters: add OutputFilterFactory interface and default GzipOutputFilterFactory; update GzipOutputFilter to honor level and buffer size.
  - Http11Processor/HTTP2 StreamProcessor: construct gzip filter using protocol-provided level/buffer settings.
  - Tests: update existing tests and add coverage for gzip level/buffer, output filter factory, and no-compression encodings.
  - Docs/I18N: document gzipLevel, gzipBufferSize and noCompressionEncodings in webapps/docs/config/http.xml; add related LocalStrings messages.
  
## Influence
  
  - Compatibility: no breaking changes; existing behavior remains default (gzip enabled per prior rules, default level -1, buffer 512).
  - Extensibility: allows custom compression filters and fine-tuned gzip settings; avoids double-compression via configurable encodings.
  - Risk: low; validated by added/updated unit tests.


markt-asf (changes requested):
Thanks for providing the PR. It has been very thought provoking.

I think we need to look at more than just GZip. Other compression algorithms are available and I think we need a more generic approach. I don't think we should be setting algorithm specific settings in the Protocol.  I also don't think they should be set on CompressionCongfig. Speaking of CompressionConfig, are the new getters there used? I looked but couldn't see any calls to them.

I think we might need a new server.xml element for output compression (and possibly one for input compression too). With attributes for className, encoding and algorithm specific attributes. It should be possible to override the default gzip support.

The handling of custom InputFilters might provide some inspiration for OutputFilters.

There are some elements of the PR that can be used as is. I'll start cherry-picking those shortly.

Long9725:
Apologies for the delayed response — I missed this review earlier. Thank you for the thoughtful feedback.

I agree that we should look beyond gzip and move toward a more generic compression approach. Supporting additional algorithms such as Zstandard was also one of the motivations behind this work. My intention was to approach this incrementally in two stages:

1. Improving gzip configuration support as a small and predictable step  
2. Moving toward a more generic design to support multiple compression algorithms as a larger follow-up change  

I split it this way because the first step felt relatively contained, while the second seemed likely to require broader architectural discussion and potentially significant changes depending on the preferred direction. Given your feedback, would you prefer this PR to remain focused on the incremental improvements, with the broader compression design handled separately, or should this PR evolve toward a more generic compression architecture?

Separately, regarding CompressionConfig — one reason I made changes there is that it is also used by the Spring Boot Tomcat integration. While investigating how to make these options configurable from Spring Boot, I found that Tomcat first needs to expose the relevant configuration properties before they can be surfaced on the Spring side.

Long9725:
I considered splitting this work into two steps, but that would force two released versions and require downstream modules to bump versions twice. To avoid that churn, I applied your feedback and implemented the general approach in one pass.

Long9725:
What I changed

## Pluggable factories
Introduce a list of OutputFilterFactory with default auto-registration of gzip; add `GzipOutputFilterFactory` with level and bufferSize tunables.
  - Implementation note: Http11OutputBuffer allocates activeFilters = new OutputFilter[filterLibrary.length + 1] to allow one dynamic filter (e.g., negotiated compression) in addition to all library filters.

## Compression negotiation
Add quality-aware selection for TE and Accept-Encoding (highest q wins; wildcard supported; q=0 excluded; ties resolved by server priority/index; case-insensitive).
  - Processors: Integrate negotiation in HTTP/1.1 and HTTP/2; install negotiated filter dynamically; set headers correctly (Transfer-Encoding for TE, Content-Encoding for Accept-Encoding); mark content length unknown when compressing.
  - EncodingNegotiator (new): Extract negotiation into a public utility with clear rules and Javadoc; CompressionConfig now delegates directly to it.

## ETC
  - Tests: Add TestEncodingNegotiator for rule-focused unit tests; keep end-to-end negotiation/headers in
  TestCompressionConfig; remove duplicated negotiation tests; fix Accept-Encoding typo and prefer getHeader() for
  TE assertions.
  - Docs: Update webapps/docs/config/http.xml to document factory-based gzip config and noCompressionEncodings; fix small typos and wrap long lines.

rmaucher:
Personally I think this is way overengineered. The only real benefit is to be able to avoid picking up any feature requests in that area, since "this is pluggable".

Background: I came up with these filters a long time ago. I thought at the time that maybe more would appear or something, but this did not materialize (that was the most likely outcome to be honest). Nobody said back then this was too complicated and I should simply fully hardcode the two actually useful ones, but there was always the excuse of "what if".

Long9725:
> Personally I think this is way overengineered. The only real benefit is to be able to avoid picking up any feature requests in that area, since "this is pluggable".
> 
> Background: I came up with these filters a long time ago. I thought at the time that maybe more would appear or something, but this did not materialize (that was the most likely outcome to be honest). Nobody said back then this was too complicated and I should simply fully hardcode the two actually useful ones, but there was always the excuse of "what if".

Thanks for the feedback.

I agree that this approach introduces abstraction before there is a concrete second implementation. My intention was to make it easier to support additional compression algorithms in the future, but I understand that this may be premature at this stage.

My main goal is to support multiple compression algorithms (such as Zstandard or Brotli), rather than introducing a pluggable structure upfront.

For additional codecs like Zstandard, I would prefer to explore them in a separate PR, starting with a straightforward codec-specific implementation. If multiple implementations emerge, we can then revisit whether a more generic structure is warranted.

If supporting additional codecs like Zstandard is not something Tomcat is aiming for, that would also be helpful to understand.

Please let me know if this direction aligns better with the preferred approach.

PR #936, opened 2026-01-06
Add virtual thread attribute to WsWebSocketContainer

adj5672:
In my app, I use WsWebSocketContainer to create websocket client.
As the number of sessions increases, memory usage also increased. So "OOM Killed" occured frequently.
I know that It is a normal situation in thread-per-connection model which is current applied.
Current model have advantage that It is very intuitive and sometimes it behaves like kind of rate limiter.

But I think that threads for R/W may not working busy in some case.
Also platform thread will be created twice of session count and it allocate a lot of default stack memory.
furthermore context switching cost expensive too when using platform thread.

So I think that instead of create two platform thread per session, using virtual thread will be helpful in some cases.
Since I may not understand overall about websocket codes and architecture, I just create draft pull request.
Adding choice to use virtual threads would be helpful for optimizing memory usage.

### DIFF
--- a/java/org/apache/tomcat/websocket/AsyncChannelWrapperSecure.java
+++ b/java/org/apache/tomcat/websocket/AsyncChannelWrapperSecure.java
@@ -42,6 +42,7 @@ import javax.net.ssl.SSLException;
 import org.apache.juli.logging.Log;
 import org.apache.juli.logging.LogFactory;
 import org.apache.tomcat.util.res.StringManager;
+import org.apache.tomcat.util.threads.VirtualThreadExecutor;
 
 /**
  * Wraps the {@link AsynchronousSocketChannel} with SSL/TLS. This needs a lot more testing before it can be considered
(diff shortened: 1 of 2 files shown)


VolareYang:
Hi! I have a question: how do I enable virtual threads in Tomcat 11? After enabling virtual threads, which log in Tomcat's startup logs can I check to confirm that virtual threads are actually being used?

My environment:
Tomcat: 11.0.23
JDK: 21
server.xml configuration:
<Executor name="virtualThreadExecutor" className="org.apache.catalina.core.StandardVirtualThreadExecutor" namePrefix="tomcat-virt-"/>
<Connector executor="virtualThreadExecutor" port="8080" protocol="HTTP/1.1"
connectionTimeout="20000"
redirectPort="8443" />

Startup logs show:
23-Jun-2026 14:51:39.306 INFO [main] org.apache.coyote.AbstractProtocol.start Starting ProtocolHandler ["http-nio-8080"]
23-Jun-2026 14:51:39.313 INFO [main] org.apache.catalina.startup.Catalina.start Server startup in [799] milliseconds

When I check the thread stack with the jstack command, I don't see any threads related to tomcat-virt- so I can't be sure if virtual threads are actually enabled. Could you guide me? Thanks!

PR #988, opened 2026-04-13
fix: Escape directory labels in DefaultServlet HTML listings

marcusschiesser:
## Summary

Escape the current directory label and parent label in `DefaultServlet` HTML directory listings.

This fixes a stored XSS issue in shared-hosting or otherwise untrusted-content deployments where directory listings are enabled and an attacker can influence listed directory names.

## Attack scenario

1. An untrusted content publisher can create directories under a served path.
2. Directory listings are enabled for that shared content root.
3. The attacker creates a directory named ``.
4. An operator browses the listing page to inspect uploaded content.
5. Tomcat renders the directory name unescaped in the generated listing page.
6. The browser executes the injected JavaScript in the origin of that site.

## Changes

- escape the current directory label in the generated listing title and heading
- escape the parent directory label in the generated listing breadcrumb
- add a regression test for malicious directory labels in `TestDefaultServlet`

## Verification

- `ant test-compile test-only -Dtest.entry=org.apache.catalina.servlets.TestDefaultServlet`

### DIFF
--- a/java/org/apache/catalina/servlets/DefaultServlet.java
+++ b/java/org/apache/catalina/servlets/DefaultServlet.java
@@ -1769,6 +1769,7 @@ public class DefaultServlet extends HttpServlet {
         StringManager sm = StringManager.getManager(DefaultServlet.class.getPackageName(), request.getLocales());
 
         String directoryWebappPath = resource.getWebappPath();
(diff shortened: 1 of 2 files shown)


rmaucher:
If you can modify the webapp resources, you can already send compromised content to the user. As a result filtering was not added to these.
Note: security concerns should always be discussed on the security mailing list

PR #990, opened 2026-04-13
Add changelog entry guidance and agreed upon standards for new contributors

csutherl:
While working on some changelog docs, I noticed that we don't really have any guidance for new contributors to create changelog entries which usually results in committers needing to do that. This commit adds a section to the CONTRIBUTORS.md guide outlining the expectation for changelog entries. Pushing here to get some agreement on the verbiage from the community before committing so feedback welcome.


markt-asf:
Nice. I like it.

PR #991, opened 2026-04-14
pom: split compound licenses

earldouglas:
Combining the licenses creates invalid URLs:

```
<licenses>
  <license>
    <url>
      http://www.apache.org/licenses/LICENSE-2.0.txt and
      http://www.opensource.org/licenses/cddl1.txt
    </url>
  </license>
</licenses>
```

Compared to using separate entries for each license:

```
<licenses>
  <license>
    <url>
      http://www.apache.org/licenses/LICENSE-2.0.txt
    </url>
  </license>
  <license>
    <url>
      http://www.opensource.org/licenses/cddl1.txt
    </url>
  </license>
</licenses>
```

This breaks tools that validate the URLs, including BOM standards such as CycloneDX:

```
license.url: does not match the iri-reference pattern must be a valid RFC 3987 IRI-reference
```

### DIFF
--- a/res/maven/tomcat-jsp-api.pom
+++ b/res/maven/tomcat-jsp-api.pom
@@ -28,17 +28,27 @@
   <licenses>
     <license>
       <name>
-        Apache License, Version 2.0 and
-        Common Development And Distribution License (CDDL) Version 1.0
+        Apache License, Version 2.0
       </name>
       <url>
-        http://www.apache.org/licenses/LICENSE-2.0.txt and
-        http://www.opensource.org/licenses/cddl1.txt
+        http://www.apache.org/licenses/LICENSE-2.0.txt
       </url>
       <distribution>repo</distribution>
       <comments>
         The Apache License, version 2.0 applies to all files apart from
-        jsp_2_2.xsd to which the CDDL version 1.0 applies.
+        jsp_2_2.xsd.
+      </comments>
+    </license>
+    <license>
+      <name>
+        Common Development And Distribution License (CDDL) Version 1.0
+      </name>
+      <url>
+        http://www.opensource.org/licenses/cddl1.txt
+      </url>
+      <distribution>repo</distribution>
+      <comments>
+        The CDDL version 1.0 applies to jsp_2_2.xsd.
       </comments>
     </license>
   </licenses>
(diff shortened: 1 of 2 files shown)


earldouglas:
If this looks good, I'll create similar PRs for the 8.5.x, 9.0.x, 10.0.x, 10.1.x, and 11.0.x branches.

n828cl:
Not commenting on the content, but please don't create multiple PRs - especially for Tomcat versions that are not supported. If and when the PR is accepted, a committer will merge the changes into each appropriate branch.

markt-asf:
While I appreciate the intention of this PR, the definition of the `<licenses>` element in a POM is that it defines a choice from multiple licenses. That isn't what we have here. I am concerned that if multiple license elements are defined, consumers of the JARs (and tooling in particular) will not take account of the notes.

Also Tomcat 7.0.x is unsupported and has been since March 2021.

earldouglas:
> the definition of the `<licenses>` element in a POM is that it defines a choice from multiple licenses

I'm not sure that's the case.  My read of <https://maven.apache.org/pom.html#Licenses> is that there should be one entry for each license that applies to the project.  In this case, two licenses always apply, and so both should get its own entry.

Combining them into a single entry is misleading, since both the `name` and `url` are invalid.  This breaks principle of least surprise for the user, as they are forced to manually read the invalid entry to interpret it.  This also breaks standards, which can prevent the use of this library and its POM as a dependency.

> Tomcat 7.0.x is unsupported

Would this PR be merged into a supported branch?

markt-asf:
> > the definition of the `<licenses>` element in a POM is that it defines a choice from multiple licenses
> 
> I'm not sure that's the case. My read of https://maven.apache.org/pom.html#Licenses is that there should be one entry for each license that applies to the project. In this case, two licenses always apply, and so both should get its own entry.

This is not the full definition:
https://maven.apache.org/ref/3.9.14/maven-model/maven.html

> Would this PR be merged into a supported branch?

No.

earldouglas:
> If multiple licenses are listed, it is assumed that the user can select any of them, not that they must accept all.

Oh interesting, I didn't know that.  Thanks for the context!

It seems like in this case it's a no-win.  We don't want to list multiple licenses, because there is not an option for the user to choose between them, and we can't combine them because that's not supported by Maven.

PR #992, opened 2026-04-14
Fix BZ 70001: _jspx_dependants contains non-deterministic absolute JA…

jluehe:
…R paths for taglibs resolved from external JARs

### DIFF
--- a/java/org/apache/jasper/compiler/Compiler.java
+++ b/java/org/apache/jasper/compiler/Compiler.java
@@ -37,6 +37,7 @@ import org.apache.jasper.servlet.JspServletWrapper;
 import org.apache.juli.logging.Log;
 import org.apache.juli.logging.LogFactory;
 import org.apache.tomcat.Jar;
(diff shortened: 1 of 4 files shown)


rmaucher:
It looks good to me. The situation it covers is not supposed to be the main usual case obviously (never been a fan of the "external shared jar" use case, but I also got the comments "I have 1000 webapps" in the past, so ...) but it's a nice improvement. Let's see if there are any final objections to this.

jluehe:
> It looks good to me. The situation it covers is not supposed to be the main usual case obviously (never been a fan of the "external shared jar" use case, but I also got the comments "I have 1000 webapps" in the past, so ...) but it's a nice improvement. Let's see if there are any final objections to this.

Thank you, @rmaucher!

jluehe:
Hi @rmaucher, just circling back on this to see if it can be merged and included in the next release - or would you rather wait a bit longer? Thank you!

markt-asf:
LGTM. I'll  merge, run the TCKs tests and back-port assuming there are no issues.

PR #994, opened 2026-04-21
Fix race condition in StandardContext.getServletContext()

dsoumis:
The context field was not volatile and getServletContext()was not synchronized, allowing concurrent threads to create competing ApplicationContext instances during reload. This could cause the TEMPDIR servlet context attribute to be permanently lost.


rmaucher:
Ok, I think it makes sense. No need to add weird tests after fixing though, you might as well testing the pattern itself in a generic way at some point.

dsoumis:
I think the test adds value to the commit as it acts both as a reproducer and as a resource for anyone inspecting the commit to seamlessly understand what was wrong and fixed. I would keep it if no objections.

PR #1001, opened 2026-04-29
Add SECURITY.md per dev-list discussion

csutherl:
This change follows up on the dev-list discussion at https://lists.apache.org/thread/4mzg1hfp79tvjrk2mgw7oqx9ydwtd3l1 with a proposal for discussion.


SECURITY.md
```diff
@@ -0,0 +1,127 @@
...
+2. **Validation** - We run your test case and assess impact against the security model
+3. **Fix & Disclosure** - If valid, we develop a fix and coordinate public disclosure timing
+4. **Credit** - Valid reports receive acknowledgment in security advisories
+
+Allow reasonable time (typically 90+ days) for fix development. We'll work with you on disclosure timing.
+
+## Security Model Quick Reference
+
+**Trusted (not security bugs):**
+- Administrative users, configuration files, binaries
```
n828cl: Might want to qualify "binaries" with something like "unmodified Tomcat binaries".

SECURITY.md
```diff
@@ -0,0 +1,127 @@
+# Security Policy
+
+## Before You Report - Required Self-Check
+
+**Complete this checklist. If you answer "No" to any question, do not submit a report:**
+
+- [ ] I have read the [Tomcat Security Model](https://tomcat.apache.org/security-model.html) and my finding doesn't require access to config files, binaries, or admin interfaces
+- [ ] I have written a working Tomcat JUnit test case that compiles, runs, and demonstrates the vulnerability
```
dsoumis: There may be corner cases that wouldn't be able to be tested through unit tests. I recall some recent examples with byte manipulation..

dsoumis:
Not crucial, but just a note from my point of view that it would be better to remove ❌ and 🔒 emojis in order to align with the tone of the rest of the docs.

SECURITY.md
markt-asf: Maybe add "... or I have provided a justification as to why a JUnit test case cannot be written for this report and I have provided a minimal proof of concept."

SECURITY.md
csutherl: I added some clarification there.

SECURITY.md
csutherl: True, I incorporated Mark's suggested text in both the checklist and added an exception section to the POC requirements section.

csutherl:
I incorporated the changes from the brief SECURITY.md that was recently added into the proposal. I only made one small tweak to what I had already done since the current SECURITY.md topics are covered in this proposed change.

markt-asf:
How do you propose to address the conflicts? There looks to be some useful wording in the current file we should retain.

I'm not sure about listing non-issues here and in the security model. I'd prefer not to duplicate.

csutherl:
I accidentally pushed a bad update, so I revisited this PR and think it's worth another look now.

SECURITY.md
```diff
@@ -1,14 +1,136 @@
...
-  <https://tomcat.apache.org/security-model.html>
-- **How to report a vulnerability**: see the Security section
-  of <https://tomcat.apache.org/>.
+**Complete this checklist. If you answer "No" to any question, do not submit a report:**
 
-The project website is the authoritative source; this file
-exists so agents and tooling that look for `SECURITY.md` in
-the repository can mechanically follow the link to the
-canonical documents.
+- [ ] I have read the [Tomcat Security Model](https://tomcat.apache.org/security-model.html) and my finding doesn't require access to config files, binaries, or admin interfaces
```
ChristopherSchultz: We might want to add "data stores" or something like that, here. I want to avoid reports of "oh, I put a serialized gadget into the session database and it exploded when Tomcat loaded it".

SECURITY.md
```diff
@@ -1,14 +1,136 @@
...
+- [ ] I have written a working Tomcat JUnit test case that compiles, runs, and demonstrates the vulnerability, or I have provided a justification as to why a JUnit test case cannot be written for this report and I have provided a minimal proof of concept
+- [ ] I have tested against a real Tomcat instance - this is not theoretical analysis or scanner output
```
ChristopherSchultz: Should we also add "the most recent release of a supported Tomcat version" to this?

I'm not saying we wouldn't accept a CVE against an older version (e.g. 8.5.last), though we obviously wouldn't fix it. But filing a CVE against 9.0.25 isn't helpful to anyone.

SECURITY.md
```diff
@@ -1,14 +1,136 @@
...
+- [ ] I am submitting in plain text (no PDFs, archives, videos, or formatted documents)
+
+**If you cannot check all boxes, your report will be rejected.**
```
ChristopherSchultz: Should we require this checklist to be included in any submitted report? My corporate gitlab instance has a template for all MRs including checklists similar to this.

SECURITY.md
csutherl: Sounds good. I added "data stores" to both the self-check checklist and Security Model reference trusted list to cover the scenario mentioned.

SECURITY.md
csutherl: I reworded this to explicitly require the most recent, supported version of tomcat.

SECURITY.md
csutherl: I think that's a good idea so I updated that to require inclusion of the completed list. We can't really automatically enforce it like that though since they reports come via email rather than issues/PRs.

csutherl:
I've resolved the feedback here and no updates in a week, so merging and will backport to 9.0.x also.

PR #1003, opened 2026-05-02
11.0.x of PR#1002: Add access-log pattern: %%

fabian-hahn:
- Support for `%%` in access-log format string, as in Apache httpd.
- Support for `{xxx}` modifiers in `%D` pattern (undocumented).
- Test-case for `%%` pattern in log format.

Back port of PR#1002 for <= 11.0.x


markt-asf:
Support added for '%%'.
   
Proposed changes to %D and %T should be discussed on the dev list.

PR #1005, opened 2026-05-08
Update StoreRegistry to dynamically load clustering classes

csutherl:
I noticed that if you use the StoreConfigListener but tomcat doesn't have the clustering JARs available that it prevents tomcat from starting successfully. The StoreConfigListener had hard dependencies on the clustering libraries, so I made those dynamic. Also added a few tests to make sure the reflection is working as expected.

### DIFF
--- a/test/org/apache/catalina/storeconfig/TestStoreRegistry.java
+++ b/test/org/apache/catalina/storeconfig/TestStoreRegistry.java
@@ -1,0 +1,104 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.catalina.storeconfig;
+
+import java.lang.reflect.Method;
+
+import org.junit.Assert;
+import org.junit.Test;
+
+import org.apache.catalina.Manager;
+import org.apache.catalina.Realm;
+import org.apache.catalina.Valve;
+
+/**
+ * Test StoreRegistry behavior, particularly dynamic loading of optional classes like clustering.
+ *
+ * Verifies StoreRegistry uses the same dynamic loading pattern.
+ */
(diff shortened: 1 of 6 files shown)


test/org/apache/catalina/storeconfig/TestStoreRegistry.java
```diff
@@ -0,0 +1,115 @@
...
+ */
+public class TestStoreRegistry {
+
+    /**
+     * Test that StoreRegistry initializes successfully without errors.
+     *
+     * Clustering classes are loaded dynamically.
+     */
+    @Test
+    public void testRegistryInitialization() {
```
dsoumis: kind of redundant

test/org/apache/catalina/storeconfig/TestStoreConfig.java
```diff
@@ -83,17 +83,23 @@ public void testListener() throws Exception {
...
-        org.apache.catalina.tribes.group.GroupChannel tc_GroupChannel_52 = new org.apache.catalina.tribes.group.GroupChannel();
-        ((org.apache.catalina.tribes.transport.ReceiverBase) tc_GroupChannel_52.getChannelReceiver()).setHost("localhost");
-
-        org.apache.catalina.tribes.membership.cloud.CloudMembershipService tc_CloudMembershipService_53 =
-                new org.apache.catalina.tribes.membership.cloud.CloudMembershipService();
-        tc_CloudMembershipService_53.setMembershipProviderClassName("org.apache.catalina.tribes.membership.cloud.KubernetesMembershipProvider");
-        tc_GroupChannel_52.setMembershipService(tc_CloudMembershipService_53);
-        tc_SimpleTcpCluster_51.setChannel(tc_GroupChannel_52);
-        tomcat.getEngine().setCluster(tc_SimpleTcpCluster_51);
+        // Clustering setup - optional, may not be available in all deployments
```
dsoumis: hmm.. the test classpath always includes the clustering JARs, so the catch block is dead code.

rmaucher: Yes, we always forget that the tests don't work off the built JARs, which can cause problems sometimes (most tests are not testing the fully packaged Tomcat).

rmaucher:
Too bad we have to add that complexity, but I guess the clustering is optional elsewhere, so ok.

test/org/apache/catalina/storeconfig/TestStoreRegistry.java
csutherl: Yeah, it's covered in the other tests. I'll remove it.

test/org/apache/catalina/storeconfig/TestStoreConfig.java
csutherl: Yeah...that's my fault. Thanks for catching that! After playing around with it some more I see that clustering is a hard dependency all over the place and is always required for test-compile. I'll revert that hunk, leave TestStoreConfig alone, and cover the reflection based class inclusions in the new TestStoreRegistry class.

I'd say moving to make clustering truly optional all the way through would be beneficial, but I don't think it's valuable to the community as we're always including the clustering libraries in the release deliverables :)

PR #1006, opened 2026-05-12
Add explicit read-only permissions to CI workflows

arpitjain099:
## Summary
- Add explicit workflow `permissions` blocks to:
  - `.github/workflows/ci.yml`
  - `.github/workflows/ci-macos.yml`
  - `.github/workflows/coverity.yml`
  - `.github/workflows/validate.yml`
- Set each workflow to least privilege with `contents: read`.

## Why
These workflows only need repository read access for checkout and build/test tasks. Explicit permissions improve GitHub Actions hardening and avoid relying on broader default token scopes.

### DIFF
--- a/.github/workflows/ci-macos.yml
+++ b/.github/workflows/ci-macos.yml
@@ -20,6 +20,9 @@ on:
     branches:
       - main
 
+permissions:
+  contents: read
+
 env:
   LC_ALL: en_US.UTF-8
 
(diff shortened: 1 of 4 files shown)


markt-asf:
Not required since the permissions are set at the repository level.
Might as well merge, if only to stop the steady stream of spam PRs and scan results claiming this is an issue.

arpitjain099:
> Not required since the permissions are set at the repository level. Might as well merge, if only to stop the steady stream of spam PRs and scan results claiming this is an issue.

Thank you @markt-asf

PR #1008, opened 2026-05-15
[INFRA] Set up default rulesets for default and release branches

asf-gitbox-commits:
This Pull Request enables the repository to conform with the "sane default security settings" of the Apache Software Foundation by configuring a default branch ruleset that protects the default branch and any release branches.

Note that `~DEFAULT_BRANCH` is a GitHub symbolic link to the current default branch (HEAD) of the repository and does not need changing.
If the managing project does not wish to set up these defaults, please close this Pull Request. Alternatively, the project may merge this Pull Request to apply the changes immediately.

If no action is taken, this Pull Request will be automatically merged by the Apache Infrastructure team on **2026-06-14** (30 days from now).

For any further information, please reach us on Slack or at: users@infra.apache.org

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -1,0 +1,14 @@
+github:
+  rulesets:
+    - name: "Default Branch Protection"
+      type: branch
+      branches:
+        includes:
+          - "~DEFAULT_BRANCH"
+          - "release/*"
+          - "rel/*"
+        excludes: []
+      bypass_teams:
+        - root
+      restrict_deletion: true
+      restrict_force_push: true
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -7,6 +7,12 @@ github:
           - "~DEFAULT_BRANCH"
           - "release/*"
           - "rel/*"
+          - "11.0.x"
+          - "10.1.x"
+          - "10.0.x"
+          - "9.0.x"
+          - "8.5.x"
+          - "7.0.x"
         excludes: []
       bypass_teams:
         - root


.asf.yaml
```diff
@@ -0,0 +1,14 @@
+github:
+  rulesets:
+    - name: "Default Branch Protection"
+      type: branch
+      branches:
+        includes:
+          - "~DEFAULT_BRANCH"
+          - "release/*"
+          - "rel/*"
```
ChristopherSchultz, line 9: It looks like this won't match any of our release branch names.

We can either do it with wildcards (which might be a tiny bit dangerous), explicitly mention each release branch, or re-name our existing branches to match these globs.

ChristopherSchultz (changes requested):

.asf.yaml
markt-asf: Lets just add our release branch names.

markt-asf: There have been no objections so I'll likely do this tomorrow.

.asf.yaml
```diff
@@ -0,0 +1,20 @@
...
+          - "11.0.x"
```
ChristopherSchultz, line 10: It looks like we might be able to do this with a quasi-regex like this: `[0-9]*` which just means "any branch name starting with a number". It might be a little dangerous if we forget and create a branch like "1st-try-fix-that-bug" and then we are stuck with it forever.

PR #1010, opened 2026-05-21
Fix implicit type‑conversion precision issues in equalsIgnoreCase

Chenjp:
See testcase, 
```java
        // bc ByteChunk bytes: Hello
        Assert.assertTrue(bc.equalsIgnoreCase("heLLo"));
        Assert.assertFalse(bc.equalsIgnoreCase("\u8a48\u8a65\u8a6c\u8a6c\u8a6f"));
```
Ascii.toLower(c) implicitly assume c<=0xff.
It is hard to exploit with this weakness. Now add an explicit check to improve robustness.


markt-asf:
Thanks for the PR. I've fixed this and a handful of similar issues.

PR #1013, opened 2026-05-28
catch invalid base64 in MimeUtility encoded-word decoder

sahvx655-wq:
Base64.getDecoder().decode throws IllegalArgumentException, not IOException, so a Content-Disposition param like `filename="=?utf-8?B?@@@@?="` lets the unchecked exception escape `decodeWord` and `ParameterParser.parse` straight out of multipart parsing, where the only catch handlers are for IOException/IllegalStateException. Adding it to the existing catch keeps the contract that decodeWord throws UnsupportedEncodingException for malformed encodings, which decodeText/ParameterParser already handle by keeping the raw value.


markt-asf:
This is a package renamed fork of Commons File Upload. Any changes need to be made in that project and then Tomcat will pick up those changes when it is next released.

PR #1014, opened 2026-05-29
Fix race condition in MemoryUserDatabase.save() by extending lock scope

sahvx655-wq:
This PR fixes a concurrency issue in MemoryUserDatabase.save() where file rename operations were performed after releasing the writeLock.

Under concurrent save requests, multiple threads could interfere with the same .new and .old files, leading to IOExceptions or inconsistent tomcat-users.xml state.

This change extends the writeLock scope to cover the complete file replacement workflow, ensuring the save operation remains atomic.

Changes
Hold writeLock until all rename operations complete
Move backup/rename logic inside the protected section
Add TestMemoryUserDatabaseConcurrency.java regression test

### DIFF
--- a/java/org/apache/catalina/users/MemoryUserDatabase.java
+++ b/java/org/apache/catalina/users/MemoryUserDatabase.java
@@ -585,37 +585,37 @@ public class MemoryUserDatabase implements UserDatabase {
                 throw ioe;
             }
             this.lastModified = fileNew.lastModified();
-        } finally {
-            writeLock.unlock();
-        }
 
-        // Perform the required renames to permanently save this file
-        File fileOld = new File(pathnameOld);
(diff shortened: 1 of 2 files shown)


markt-asf (changes requested):
This doesn't even compile. With the compilation issues fixed, the test always fails - even with the fix.

Given the poor quality, I was close to simply closing this PR. The next iteration needs to be a lot better to avoid being closed.

markt-asf:
No response for two weeks. Simple fix applied.

sahvx655-wq:
Apologies for the delay and for the issues in the original submission. Thank you for taking the time to review it. If you are open to it, I would like to work on a corrected version and submit a new PR

markt-asf:
No objections.

PR #1016, opened 2026-05-29
Ensure atomic session persistence in FileStore using write-then-rename pattern

sahvx655-wq:
Updated FileStore.save(Session) to use an atomic write-then-rename pattern. Previously, session data was written directly to the final file, which could leave partially written or inconsistent session state if the process was interrupted during persistence. The new logic writes to a temporary .tmp file first and renames it only after a successful write, improving persistence reliability and reducing the risk of corruption.

### DIFF
--- a/java/org/apache/catalina/session/FileStore.java
+++ b/java/org/apache/catalina/session/FileStore.java
@@ -24,6 +24,9 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.ObjectInputStream;
 import java.io.ObjectOutputStream;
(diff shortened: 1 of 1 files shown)


markt-asf (changes requested):
The proposed changes no longer handle the (very likely) case that the destination file exists reliably.

sahvx655-wq:
I have updated the implementation to handle existing destination files more reliably and pushed the changes for review.

markt-asf:
Thanks for the updated PR. I have applied it manually with a slightly modified try/finally structure as there were code paths (e.g. `IOException` during session write) where the temporary file would not have been deleted.

PR #1017, opened 2026-05-30
Remoteip host header parsing fix

sahvx655-wq:
Add validation to check whether the host header value becomes empty after comma parsing and trimming. If the resulting value is empty, skip host header processing and log a debug message, allowing the request to continue using the original server name instead of attempting to parse an invalid host value.


markt-asf:
The header does not support multiple values.

PR #1020, opened 2026-06-09
Add showReport attribute to valves

dsoumis:


markt-asf:
+1. You'll want a change log entry as well. +1 for back-porting too.

PR #1021, opened 2026-06-09
Make Cluster digester rules fully conditional on ha and tribes JARs

dsoumis:
The ObjectCreate rules for the <Cluster> element were registered
  unconditionally in EngineRuleSet and HostRuleSet, while the child
  element rules in ClusterRuleSet were already guarded behind a
  try/catch in Catalina.addClusterRuleSet(). This caused a
  NoClassDefFoundError crash when catalina-tribes.jar was removed
  but catalina-ha.jar was present and <Cluster> was enabled in
  server.xml.

  Move the Cluster element rules into addClusterRuleSet() alongside
  the child element rules, and add a probe for tribes availability.
  Either both JARs are present and clustering works, or the <Cluster>
  element is ignored with an INFO log.

### DIFF
--- a/java/org/apache/catalina/startup/Catalina.java
+++ b/java/org/apache/catalina/startup/Catalina.java
@@ -597,8 +597,15 @@ public class Catalina {
         Constructor<?> constructor;
         try {
             clazz = Class.forName("org.apache.catalina.ha.ClusterRuleSet");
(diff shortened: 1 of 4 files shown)


ChristopherSchultz:
+1

csutherl:
LGTM

PR #1025, opened 2026-06-22
Add test for embedded Tomcat stop() idempotency

fatima-qaisar:
Verify that calling Tomcat.stop() on an already stopped instance has
no effect, per the Lifecycle interface contract. Confirms that
destroy() is invoked only once on the contained servlet even when
stop() is called twice.

Also fix typo in testGetBrokenContextPerAddWebapp() name


fatima-qaisar:
Closing this because, after taking a closer look, I don't think it adds enough value to justify a dedicated test

PR #1026, opened 2026-06-23
Add RFC 10008 (The HTTP QUERY Method) Support

desiderantes:
This change adds support for RFC 10008 (The HTTP QUERY Method). QUERY is a safe an idempotent method that carries a body. This implementation adds support for redirections, caching, mandatory content type, and ALLOW.

### DIFF
--- a/java/jakarta/servlet/http/HttpServlet.java
+++ b/java/jakarta/servlet/http/HttpServlet.java
@@ -80,6 +80,7 @@ public abstract class HttpServlet extends GenericServlet {
     private static final String METHOD_POST = "POST";
     private static final String METHOD_PUT = "PUT";
     private static final String METHOD_TRACE = "TRACE";
+    private static final String METHOD_QUERY = "QUERY";
 
     private static final String HEADER_IFMODSINCE = "If-Modified-Since";
     private static final String HEADER_LASTMOD = "Last-Modified";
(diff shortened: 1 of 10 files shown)


rmaucher:
EE specification link for these changes ?
No test that parameters processing works (probably it does).

markt-asf:
It isn't part of the Servlet spec yet so we can't modify specification classes.

This looks like a reasonable proposal. I'll look at getting the Servlet specification updated.

desiderantes:
> EE specification link for these changes ?

Sorry, no EE spec change at this time (Honestly I don't' know where or how to start doing that)

> No test that parameters processing works (probably it does).

Added

desiderantes:
> It isn't part of the Servlet spec yet so we can't modify specification classes.

Should I remove that addition then?

> This looks like a reasonable proposal. I'll look at getting the Servlet specification updated.

Thank you! Yes, RFC 10008 is a very simple addition, hopefully we can get it when 12 gets released

ffray:
Coded nearly the same this morning. It'd be awesome to have a QUERY method.
To avoid P2 data in URLs, QUERY would be just great to stop abusing POST.

I'd like to contribute the AJP support, messages and JSP support if this feature is accepted.

markt-asf:
See https://github.com/jakartaee/servlet/pull/1069

markt-asf:
Implemented with a few minor variations. Obviously, only available in Tomcat 12 due to the API changes.

PR #1028, opened 2026-06-29
Const time equal params order

Chenjp:
Ensure the first one is from server-side, and the 2nd is came from user input.


markt-asf:
Please provide a justification for this change.

ChristopherSchultz:
I actually had this backward in my mind, thinking that user-input-first was better because it would mean that runtime is based upon the user's input, which they already know and wouldn't learn anything about the secret. But I think that constant-time across calls is actually better, where regardless of the user-input, the comparison always takes the same amount of time.

It also means that the attacker can't send unreasonably-large values and cause `equals()` to run longer than necessary.

markt-asf:
You could argue it either way. I think, on balance, time based on user input is marginally lower risk (assuming sensible per request limits).

I'm not seeing a strong argument in favour of either approach which for me suggests retaining the status quo.

markt-asf:
Closing as no basis provided for change.

PR #1031, opened 2026-07-28
Improve defensive programming and modernize code in Import.processFile()

Sovtouch:
**Motivation:**
While reviewing static analysis findings, a potential NullPointerException was reported for a call to Properties.getProperty(). In the current implementation, the code iterates over props.keySet(), so every key is guaranteed to exist and getProperty() will not return null in practice. However, this safety relies on an implicit assumption about how the loop works today. If the code is refactored in the future to use keys from another source, the warning could become a real issue.

**Changes:**
Added an explicit value == null check before accessing the string value. This makes the code more defensive and aligns it with the Properties.getProperty() contract, which allows returning null.
Replaced trim().length() == 0 with the more idiomatic and readable trim().isEmpty().

**Impact:**
Low risk. This affects a build utility (org.apache.tomcat.buildutil.translate.Import) rather than the Tomcat runtime itself. It improves readability and makes the code more resilient to future changes without affecting the current behavior.

### DIFF
--- a/java/org/apache/tomcat/buildutil/translate/Import.java
+++ b/java/org/apache/tomcat/buildutil/translate/Import.java
@@ -72,7 +72,7 @@ public class Import {
             String key = (String) objKey;
             String value = props.getProperty(key);
             // Skip untranslated values
-            if (value.trim().isEmpty()) {
+            if (value == null || value.trim().isEmpty()) {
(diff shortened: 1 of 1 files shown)


markt-asf:
The code is unnecessary.

PR #1036, opened 2026-08-05
Make HPACK encoding test deterministic

sainadh777:
## Summary

- allow package-local injection of the HPACK header encoding policy
- make `TestHpack.testEncode()` explicitly force Huffman encoding
- remove the checked-in TODO and the test's dependency on the default length heuristic

## Rationale

`TestHpack.testEncode()` asserts a 27-byte encoded result, which assumes Huffman encoding. The default encoder currently chooses Huffman encoding for values longer than five characters, but that heuristic may change. This implements the guidance in `TODO.md` by making the test select Huffman encoding directly.

The default constructor still uses `DEFAULT_HEADER_FUNCTION`, so production behavior is unchanged. The new constructor and the exposed nested interface remain package-private.

## Verification

- `ant test -Dtest.entry=org.apache.coyote.http2.TestHpack -Dtest.silent=true`
- `ant -Dexecute.validate=true validate`
- complete unfiltered suite on Linux/JDK 21: `ant test -Dtest.silent=true -Dtest.threads=4 -Dtest.openssl.path=/usr/bin/openssl`
- clean source build: `ant clean`, then `ant`
- runtime smoke: started the generated distribution, received HTTP 200 from `http://127.0.0.1:8080/`, and stopped Tomcat cleanly

### DIFF
--- a/java/org/apache/coyote/http2/HpackEncoder.java
+++ b/java/org/apache/coyote/http2/HpackEncoder.java
@@ -109,7 +109,11 @@ class HpackEncoder {
     private final HpackHeaderFunction hpackHeaderFunction;
 
     HpackEncoder() {
-        this.hpackHeaderFunction = DEFAULT_HEADER_FUNCTION;
+        this(DEFAULT_HEADER_FUNCTION);
+    }
+
+    HpackEncoder(HpackHeaderFunction hpackHeaderFunction) {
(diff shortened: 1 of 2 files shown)


sainadh777:
Can I get reviews on this please ?

sainadh777:
@markt-asf could you help to review this ?

PR #1038, opened 2026-08-08
Add tests to verify tomcat functions properly when clustering JARs are not present at runtime

csutherl:
This change follows up on 91ac7518 and a few other recent commits which make cluster classloading fully dynamic by using reflection to handle clustering classes that are missing at runtime. These tests verify that tomcat work correctly when `catalina-ha.jar` and `catalina-tribes.jar` are not present using filtering classloaders to simulate  that environment.

The tests cover:

- Catalina startup: Tomcat starts and serves requests when clustering classes are unavailable (`addClusterRuleSet()` logs instead of failing)
- `sessionsList.jsp`: The Manager app sessions page compiles and renders without clustering, exercising the reflection-based fallback from 91ac7518
- StandardEngineSF/StandardHostSF:  `StoreConfig` classes load without error; `clusterValveClass` fields are null when clustering is absent
- `StoreRegistry.getInterfaces()`: Returns only the 10 core interfaces (no clustering interfaces) when clustering classes are not on the classpath

Each test uses a custom classloader (`NoClusterClassLoader` or `StoreConfigIsolatingClassLoader`) that throws a `ClassNotFoundException` for `org.apache.catalina.ha.*` and `org.apache.catalina.tribes.*` packages.

They're just a tests but I'm submitting this PR for some feedback since it's a bit of a hacky solution before committing.

### DIFF
--- a/test/org/apache/catalina/startup/TestWithoutClustering.java
+++ b/test/org/apache/catalina/startup/TestWithoutClustering.java
@@ -1,0 +1,447 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.catalina.startup;
+
+import java.io.File;
+import java.io.IOException;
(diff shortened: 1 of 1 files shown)


markt-asf:
The solution looks reasonable to me.

Not all of the tests confirm that clustering classes have not been loaded. If it isn't too convoluted, I think it would be worth confirming that in every test.

csutherl:
I pushed an update to make the class name and methods consistent so that it's all "NoClustering" instead of a mix. Also updated the javadoc and method name from `testCatalinaStartsWithoutClustering` to `testTomcatStartsWithoutClustering` since I wasn't actually testing `addClusterRuleSet()` as I originally intended. And added a check to each test to confirm clustering is in fact not present.

markt-asf:
LGTM.

I might be tempted to modify the "clustering not available check" to use a servlet just to be certain that those classes aren't visible to either Tomcat or a web application but that is only a nice to have.

csutherl:
Thanks for the review! I think it make sense to confirm the classes aren't visible. Modifying the current check was a bit annoying, so I added a new test and servlet to confirm.

markt-asf:
LGTM

csutherl:
For future reference, the failing smoke test on Mac seemed to be caused by timeouts on the test running and definitely not introduced by this test addition, so I waived it and proceeded with the merge.

PR #1039, opened 2026-08-12
Fix subsection heading levels for nested subsections (h5/h6)

fatima-qaisar:
Fix the XSLT template for nested `<subsection>` elements so that the
generated heading level reflects the nesting depth.

Previously, every `<subsection>` was rendered as `<h4>`, regardless of
whether it was nested inside another subsection.

The template now generates:
- `<h4>` for subsections directly under a `<section>`
- `<h5>` for subsections nested one level deep
- `<h6>` for subsections nested two or more levels deep

Testing:
- `ant build-docs` - confirmed the generated documentation builds
  successfully and the headings are rendered correctly in the output
- Verified visually with a local test XML document containing nested
  subsections; the generated headings now use the expected h4 → h5 → h6
  hierarchy.

Screenshots attached showing the correct rendering in both:
- an actual Tomcat documentation page
  
test

- the local nested-subsection test XML output

nesting test

### DIFF
--- a/webapps/docs/tomcat-docs.xsl
+++ b/webapps/docs/tomcat-docs.xsl
@@ -231,11 +231,23 @@
 
     <div class="subsection">
       <!-- Subsection heading -->
-      <!-- TODO: When a <subsection> is nested in another <subsection>,
-           the output should be <h5>, not <h4>. Same with <h6>. -->
-      <h4 id="{$name}">
-        <xsl:value-of select="@name"/>
-      </h4>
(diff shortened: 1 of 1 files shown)


ChristopherSchultz:
Even though we don't actually have any sub-subsections anywhere in our documentation, this does resolve a 13-year-old TODO. :)

fatima-qaisar:
There was ONE instance of it in the User Guide. And at least this was interesting to resolve :)

PR #1040, opened 2026-08-13
Wrap Instant conversion failures in ELException

lihongyi87:
## Problem

Coercion to `java.time.Instant` (added for Tomcat 12) can leak unchecked exceptions instead of the `ELException` required by the EL specification.

In `ELSupport.coerceToInstant()`:

- `Instant.parse("not-a-date")` throws `java.time.format.DateTimeParseException`
- `Instant.from(LocalDate.of(...))` throws `java.time.DateTimeException` / `UnsupportedTemporalTypeException`

Both are `RuntimeException` subclasses that propagate directly to the caller, while the method contract (and the rest of `ELSupport`, e.g. `coerceToNumber`) requires a `jakarta.el.ELException`.

## Fix

Wrap the failures from `Instant.from()` and `Instant.parse()` in `ELException`, matching the pattern already used by `coerceToNumber` for `NumberFormatException`.

## Testing

Added three tests to `TestELSupport`:

- `testCoerceToInstant01` — valid string coerces to `Instant`
- `testCoerceToInstant02` — invalid string throws `ELException`
- `testCoerceToInstant03` — `LocalDate` (no timezone) throws `ELException`

`TestELSupport` passes (54 tests, 0 failures) on JDK 23.

### DIFF
--- a/java/org/apache/el/lang/ELSupport.java
+++ b/java/org/apache/el/lang/ELSupport.java
@@ -25,6 +25,7 @@ import java.lang.reflect.Proxy;
 import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.time.Clock;
+import java.time.DateTimeException;
 import java.time.Instant;
 import java.time.temporal.TemporalAccessor;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


ChristopherSchultz:
LGTM

PR #1041, opened 2026-08-13
Fix automatic pong close race

moritzfl:
## Fix automatic Pong/session-close race

A WebSocket session may begin closing after `WsFrameBase` checks `WsSession.isOpen()` but before it calls `sendPong()`. In that interval, `sendPong()` throws `IllegalStateException` on the inbound frame-processing thread. 

=> we have a TOCTOU situation

Suppress this exception only when closing has begun; propagate it otherwise. Add regression coverage for both outcomes.

### Backport applicability

The same non-atomic `isOpen()` / `sendPong()` sequence exists in all supported release branches: `11.0.x`, `10.1.x`, and `9.0.x`. These should be considered for backport after merge to `main`. We would need a fix for `11.0.x` at the company I work for.

### Testing

`ant test` with JDK 21.0.9 ran for 57m57s on my Windows machine but failed due to broad NIO2 test discovery failures (`No tests found`) affecting unrelated tests. The added `TestWsFrame` coverage passed under NIO: 4 tests, 0 failures.

### DIFF
--- a/java/org/apache/tomcat/websocket/WsFrameBase.java
+++ b/java/org/apache/tomcat/websocket/WsFrameBase.java
@@ -377,7 +377,14 @@ public abstract class WsFrameBase {
             wsSession.onClose(new CloseReason(Util.getCloseCode(code), reason));
         } else if (opCode == Constants.OPCODE_PING) {
             if (wsSession.isOpen()) {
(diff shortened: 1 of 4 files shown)


markt-asf:
Thanks for the PR. Looks good. I'll also take care of back ports.

PR #1042, opened 2026-08-14
Avoid infinite loop in DeltaManager with invalid batch size

lihongyi87:
## Problem

When `sendAllSessions` is `false` and `sendAllSessionsSize` is configured to zero or a negative value, the batch loop in `handleGET_ALL_SESSIONS` never terminates:

```java
for (int i = 0; i < currentSessions.length; i += getSendAllSessionsSize()) {
```

- With `sendAllSessionsSize == 0`: `i` never advances, so the loop runs indefinitely, sending empty session arrays and sleeping on each iteration. This blocks the cluster receiver thread on the master node and prevents `EVT_ALL_SESSION_TRANSFERCOMPLETE` from ever being sent.
- With `sendAllSessionsSize < 0`: `i` decreases, and `new Session[len]` throws `NegativeArraySizeException`.

The `setSendAllSessionsSize` setter performs no validation, and the documentation (`cluster-manager.xml`) does not state a minimum value.

## Fix

When `sendAllSessionsSize` is not positive, fall back to sending all sessions in a single batch (the same behaviour as `sendAllSessions == true`):

```java
if (isSendAllSessions() || getSendAllSessionsSize() <= 0) {
```

This is a one-line change at the point of use, keeping the setter unmodified for consistency with the rest of `DeltaManager`.

### DIFF
--- a/webapps/docs/changelog.xml
+++ b/webapps/docs/changelog.xml
@@ -320,6 +320,13 @@
   <subsection name="Cluster">
     <changelog>
       <!-- Entries for backport and removal before 12.0.0-M1 below this line -->
+      <fix>
+        Validate that the <code>DeltaManager</code> attribute
+        <code>sendAllSessionsSize</code> is a positive integer. Zero or
+        negative values previously caused an infinite loop or a
+        <code>NegativeArraySizeException</code> during session state transfer.
+        (lihongyi87)
+      </fix>
     </changelog>
   </subsection>
   <subsection name="WebSocket">
--- a/java/org/apache/catalina/ha/session/DeltaManager.java
+++ b/java/org/apache/catalina/ha/session/DeltaManager.java
@@ -438,9 +438,15 @@ public class DeltaManager extends ClusterManagerBase {
     /**
      * Set the batch size for sending all sessions.
      *
-     * @param sendAllSessionsSize The batch size value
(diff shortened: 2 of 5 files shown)


markt-asf:
I'd rather document that `sendAllSessionsSize` must be a positive integer and throw `IllegalArgumentException` if it isn't.

lihongyi87:
Thanks for the review. Reworked as suggested: the setter now rejects a non-positive `sendAllSessionsSize` with an `IllegalArgumentException`, the documentation states the value must be a positive integer, and the fallback in `handleGET_ALL_SESSIONS` has been reverted. Tests updated accordingly.

webapps/docs/changelog.xml
```diff
@@ -331,6 +319,13 @@
   </subsection>
   <subsection name="Cluster">
     <changelog>
+      <fix>
+        Validate that the <code>DeltaManager</code> attribute
+        <code>sendAllSessionsSize</code> is a positive integer. Zero or
+        negative values previously caused an infinite loop or a
+        <code>NegativeArraySizeException</code> during session state transfer.
+        (lihongyi87)
+      </fix>
```
markt-asf, line 329: This needs to be below the line since it will be back-ported.

lihongyi87: Moved below the line. Thanks for the pointer.

PR #1043, opened 2026-08-14
Fix typos

k-tamura:
There were several obvious typos regarding RFC number in the source code.

- RFC 7450 - Automatic Multicast Tunneling 🙅 
- RFC 7540 - Hypertext Transfer Protocol Version 2 (HTTP/2) 🙆


k-tamura:
Since these are just simple typos, I didn’t create an issue for this in Bugzilla.

ebourg:
Good catch, thank you

PR #1044, opened 2026-08-15
Wrap RewriteRule syntax reference

sainadh777:
## Summary

- replace the RewriteRule regular-expression syntax reference with a semantic two-column table
- use `<code>...</code>` only for syntax in the first column and standard prose in the second
- remove the checked-in TODO about the non-wrapping block
- add the corresponding changelog entry

## Rationale

The original reference was a raw `pre` element and did not adapt to narrow displays. The initial PR revision used Tomcat's wrapped source template, but that treated the entire block as source code. Following review feedback, the table now models the content correctly: syntax is code, descriptions are prose, and the browser can wrap the descriptive column naturally.

## Impact

This is documentation-only. The five existing sections and their reference content are preserved while the generated markup becomes semantic and responsive.

## Validation

- `ant -q build-docs` — passed; generated `rewrite.html` contains the two-column table
- generated-page responsive inspection at 1280 px and 375 px viewports — passed; all five section headings were present, the description column contained no `code` elements, and all 22 data cells had zero internal overflow at 375 px
- `ant -Dexecute.validate=true validate` — passed (Checkstyle 13.9.0 across 7,677 checked files, 54 seconds)
- `ant clean` — passed (1 second)
- `ant` — passed clean source build (27 seconds; existing deprecation, BND, and Java 8 source/target warnings only)
- `git diff --check` — passed
- `ant test -Dtest.silent=true -Dtest.threads=4 -Dtest.openssl.path=/opt/homebrew/opt/openssl@3/bin/openssl` — passed complete unfiltered suite in 16 minutes 37 seconds: 649 suites, 41,257 tests, 0 failures, 0 errors, 263 project-declared skips
- generated-distribution smoke with `JAVA_HOME=/opt/homebrew/opt/openjdk/libexec/openjdk.jdk/Contents/Home` — passed: `http://127.0.0.1:8080/docs/rewrite.html` returned HTTP 200, contained the new table, and the server PID exited cleanly after shutdown

### DIFF
--- a/webapps/docs/rewrite.xml
+++ b/webapps/docs/rewrite.xml
@@ -527,33 +527,30 @@ RewriteRule ^/(.*)$ ${uc:$1}
       <p>Some hints on the syntax of regular
       expressions:</p>
 
-<!-- TODO: Why is the following pre-formatted non-wrappable text? -->
-<pre>
-<strong>Text:</strong>
-  <strong><code>.</code></strong>           Any single character
-  <strong><code>[</code></strong>chars<strong><code>]</code></strong>     Character class: Any character of the class 'chars'
-  <strong><code>[^</code></strong>chars<strong><code>]</code></strong>    Character class: Not a character of the class 'chars'
-  text1<strong><code>|</code></strong>text2 Alternative: text1 or text2
+<source wrapped="true"><strong>Text:</strong>
+  <strong>.</strong>           Any single character
+  <strong>[</strong>chars<strong>]</strong>     Character class: Any character of the class 'chars'
+  <strong>[^</strong>chars<strong>]</strong>    Character class: Not a character of the class 'chars'
+  text1<strong>|</strong>text2 Alternative: text1 or text2
 
 <strong>Quantifiers:</strong>
-  <strong><code>?</code></strong>           0 or 1 occurrences of the preceding text
-  <strong><code>*</code></strong>           0 or N occurrences of the preceding text (N &gt; 0)
-  <strong><code>+</code></strong>           1 or N occurrences of the preceding text (N &gt; 1)
+  <strong>?</strong>           0 or 1 occurrences of the preceding text
+  <strong>*</strong>           0 or N occurrences of the preceding text (N &gt; 0)
+  <strong>+</strong>           1 or N occurrences of the preceding text (N &gt; 1)
 
 <strong>Grouping:</strong>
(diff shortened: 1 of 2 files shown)


sainadh777:
@FSchumacher, would you be willing to review this small RewriteValve documentation formatting change? I selected you because of your prior contributions to `webapps/docs/rewrite.xml`. Thank you.

sainadh777:
@markt-asf, would you also be willing to review this small RewriteValve documentation formatting change? I selected you because you are the leading contributor to `webapps/docs/rewrite.xml`. Thank you.

rmaucher:
The current one seems readable and fine to me.

webapps/docs/rewrite.xml
```diff
@@ -527,33 +527,30 @@ RewriteRule ^/(.*)$ ${uc:$1}
       <p>Some hints on the syntax of regular
       expressions:</p>
 
-<!-- TODO: Why is the following pre-formatted non-wrappable text? -->
-<pre>
```
markt-asf, line 531: Removing this is good. This is what the TODO was for,

webapps/docs/rewrite.xml
```diff
@@ -527,33 +527,30 @@ RewriteRule ^/(.*)$ ${uc:$1}
...
-<strong>Text:</strong>
-  <strong><code>.</code></strong>           Any single character
-  <strong><code>[</code></strong>chars<strong><code>]</code></strong>     Character class: Any character of the class 'chars'
-  <strong><code>[^</code></strong>chars<strong><code>]</code></strong>    Character class: Not a character of the class 'chars'
-  text1<strong><code>|</code></strong>text2 Alternative: text1 or text2
+<source wrapped="true"><strong>Text:</strong>
+  <strong>.</strong>           Any single character
+  <strong>[</strong>chars<strong>]</strong>     Character class: Any character of the class 'chars'
+  <strong>[^</strong>chars<strong>]</strong>    Character class: Not a character of the class 'chars'
+  text1<strong>|</strong>text2 Alternative: text1 or text2
```
markt-asf: This is the wrong approach. The main issue is that the whole block is not source code, only the first column is. What is required here is a table where the first column uses <code>...</code> and the second column uses standard text. The formatting for different screen widths should then sort itself out.

markt-asf (changes requested):
The comment on the `Text` block applies to all of the blocks.

PR #1051, opened 2026-08-24
Skip OCSP tests when responder port is unavailable instead of fail

csutherl:
`TesterOcspResponder` binds to a fixed port, 8888 (which is baked into the test certificates' AIA extension), so it cannot start when another process is already using that port. The existing `Assume.assumeNotNull(ocspResponder)` guard was not enough because the field was assigned before `start()` was  called and the bind failure was only printed, leaving the field non-null and causing all tests to fail during the TLS handshake. This change only assigns `ocspResponder` on a successful start. On a `BindException`, we leave the field `null` so the tests are skipped as an environmental issue rather than failing. Any other startup failure is rethrown so genuine responder bugs are not masked.

Submitting a PR to get consensus on whether or not this is the correct behavior, or if we feel a failure is appropriate.

### DIFF
--- a/test/org/apache/tomcat/util/net/ocsp/TestOcspEnabled.java
+++ b/test/org/apache/tomcat/util/net/ocsp/TestOcspEnabled.java
@@ -16,6 +16,7 @@
  */
 package org.apache.tomcat.util.net.ocsp;
 
+import java.net.BindException;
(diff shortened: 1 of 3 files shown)


markt-asf:
They get logged as skipped so I think that is OK.

PR #1052, opened 2026-08-24
Skip tribes membership tests when IP multicast is unavailable rather than fail

csutherl:
Adding another skip rather than fail scenario for tests that cannot utilize multicast in the test environment.


dsoumis:
lgtm

PR #1053, opened 2026-08-24
Upgrade actions/setup-java to v6

brunoborges:
## Summary

Upgrade `actions/setup-java` from v5 to v6 in every active workflow that uses it.

## Testing

- Ran `git diff --check`.
- Verified no pre-v6 `actions/setup-java` references remain in active workflows.


brunoborges:
my apologies for this PR. It was only supposed to hit a set of repos with a specific dependabot configuration that would prevent getting the alert of this new release

PR #1054, opened 2026-08-27
Allow Request#getDateHeader to parse dates in RFC 9651 Structured Field Values format

sabberworm:
[RFC 9651 (Structured Field Values for HTTP)](https://datatracker.ietf.org/doc/rfc9651/) defines formats for header values of specific types. The format it uses for dates is `@«timestamp»`, which is different from other date headers.

This PR aims to support the date header format from RFC 9651 in `org.apache.catalina.connector.Request#getDateHeader(String)`.

### DIFF
--- a/java/org/apache/tomcat/util/http/FastHttpDateFormat.java
+++ b/java/org/apache/tomcat/util/http/FastHttpDateFormat.java
@@ -63,13 +63,13 @@ public final class FastHttpDateFormat {
         FORMAT_OBSOLETE_ASCTIME = new ConcurrentDateFormat(DATE_OBSOLETE_ASCTIME, Locale.US, tz);
         FORMAT_RFC9651 = dateString -> {
             if(dateString == null || !dateString.startsWith("@")) {
-                throw new ParseException("Date " + dateString + " not in RFC 9651 format", 0);
+                return -1;
             }
             try {
                 // An RFC 9651 timestamp is in seconds, not milliseconds.
                 return Long.parseLong(dateString.substring(1)) * 1_000;
             } catch (NumberFormatException e) {
-                throw new ParseException("Unable to parse number in date string " + dateString, 1);
+                return -1;
             }
         };
 
@@ -158,13 +158,9 @@ public final class FastHttpDateFormat {
 
         long date = -1;
         for (int i = 0; (date == -1) && (i < httpParseFormats.length); i++) {
-            try {
-                date = httpParseFormats[i].tryParseDate(value);
-                updateParseCache(value, Long.valueOf(date));
-            } catch (ParseException e) {
-                // Ignore
-            }
+            date = httpParseFormats[i].tryParseDate(value);
         }
+        updateParseCache(value, Long.valueOf(date));
 
         return date;
     }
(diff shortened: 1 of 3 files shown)


java/org/apache/tomcat/util/http/FastHttpDateFormat.java
```diff
@@ -60,9 +61,19 @@ private FastHttpDateFormat() {
         FORMAT_RFC5322 = new ConcurrentDateFormat(DATE_RFC5322, Locale.US, tz);
         FORMAT_OBSOLETE_RFC850 = new ConcurrentDateFormat(DATE_OBSOLETE_RFC850, Locale.US, tz);
         FORMAT_OBSOLETE_ASCTIME = new ConcurrentDateFormat(DATE_OBSOLETE_ASCTIME, Locale.US, tz);
+        FORMAT_RFC9651 = dateString -> {
+            if(dateString == null || !dateString.startsWith("@")) {
+                throw new ParseException("Date " + dateString + " not in RFC 9651 format", 0);
+            }
```
sabberworm, line 67: I don’t like the overhead of using exception handling for control flow here. Maybe use `-1` here too?

java/org/apache/tomcat/util/http/FastHttpDateFormat.java
```diff
@@ -147,7 +158,7 @@ public static long parseDate(String value) {
         long date = -1;
         for (int i = 0; (date == -1) && (i < httpParseFormats.length); i++) {
             try {
-                date = httpParseFormats[i].parse(value).getTime();
+                date = httpParseFormats[i].tryParseDate(value);
                 updateParseCache(value, Long.valueOf(date));
```
sabberworm: Is this missing a 
```suggestion
                updateParseCache(value, Long.valueOf(date));
                break;
```
?

To me it doesn’t look like any string can ever match more than one format so we’re needlessly trying all of them.

sabberworm: Sorry, this is already covered by the `(date == -1)` condition in the `for` loop.

java/org/apache/tomcat/util/http/FastHttpDateFormat.java
sabberworm: Done

java/org/apache/tomcat/util/http/FastHttpDateFormat.java
```diff
@@ -146,13 +158,9 @@ public static long parseDate(String value) {
...
         for (int i = 0; (date == -1) && (i < httpParseFormats.length); i++) {
-            try {
-                date = httpParseFormats[i].parse(value).getTime();
-                updateParseCache(value, Long.valueOf(date));
-            } catch (ParseException e) {
-                // Ignore
-            }
+            date = httpParseFormats[i].tryParseDate(value);
         }
+        updateParseCache(value, Long.valueOf(date));
```
sabberworm, line 163: Now this is also called for values that failed to parse. Not sure if that’s better or worse. IMHO it makes sense to also cache these, since they’re the most expensive to calculate as they have to be tried (and rejected) by all the parsers.

java/org/apache/tomcat/util/http/FastHttpDateFormat.java
```diff
@@ -185,5 +193,14 @@ private static void updateParseCache(String key, Long value) {
...
     }
 
-
+    @FunctionalInterface
+    interface TryParseDateToTimestamp {
+        /**
+         * Tries to parse the given string as a date and returns it as a timestamp.
+         *
+         * @param dateString the string representation of the date trying to be parsed
+         * @return the number of *milli*seconds since January 1, 1970, 00:00:00 GMT or -1 if parsing failed
```
sabberworm, line 202:
```suggestion
         * @return the number of <em>milli</em>seconds since January 1, 1970, 00:00:00 GMT or -1 if parsing failed
```

PR #1059, opened 2026-09-03
Add opt-in ServerInfo.properties override support

csutherl:
This change fixes what I think is a bug in how we override `ServerInfo.properties` at runtime. The issue is that when using an override `ServerInfo.properties` in `lib/` and not setting some values (i.e. `server.built`), the defaults that are hardcoded in the class are used rather than the bundled values.

The change adds an opt-in system property, `org.apache.catalina.util.LOAD_SERVER_INFO_OVERRIDE` (exposed as `Globals.LOAD_SERVER_INFO_OVERRIDE`). When set, every `ServerInfo.properties` on the class path is merged in reverse order (`lib/` is loaded first), so the bundled `catalina.jar` copy supplies the defaults and an override file (e.g. in `$CATALINA_BASE/lib`) replaces only the individual properties it sets instead of falling back to the hard coded defaults. When unset (the default), behaviour is unchanged: only the first `ServerInfo.properties` found is loaded and if you don't provide a value in your override file, the hardcoded value is used.

My reading of the documentation in the `security-howto.xml` is that users should change the values they want (i.e. the version string) by creating `CATALINA_BASE/lib/org/apache/catalina/util/ServerInfo.properties`. In practice the overridden values happen, but the other values in the `ServerInfo.properites` are set to the hardcoded defaults rather than the bundled values, which isn't explicitly stated and unexpected behavior. The addition of this flag makes the documented/expected behaviour actually work, via partial override.

The change is adding an opt-in configuration because merging by default would change class-path loading semantics for every existing deployment which may result in unexpected values at runtime. The flag keeps the default identical and lets operators opt in if desired. I think we should maybe add it and backport it, deprecate it in Tomcat 11, then switch the default and drop the configurable option in Tomcat 12. Thoughts?

### DIFF
--- a/java/org/apache/catalina/Globals.java
+++ b/java/org/apache/catalina/Globals.java
@@ -233,6 +233,15 @@ public final class Globals {
     public static final boolean STRICT_SERVLET_COMPLIANCE =
             Boolean.getBoolean("org.apache.catalina.STRICT_SERVLET_COMPLIANCE");
 
+    /**
+     * If {@code true}, every ServerInfo.properties on the class path is merged
+     * (bundled catalina.jar copy first as defaults, override files on top) so an
+     * override file can replace individual properties. Otherwise only the first
+     * ServerInfo.properties found is used.
+     */
+    public static final boolean LOAD_SERVER_INFO_OVERRIDE =
+            Boolean.getBoolean("org.apache.catalina.util.LOAD_SERVER_INFO_OVERRIDE");
+
 
     /**
      * Default domain for MBeans if none can be determined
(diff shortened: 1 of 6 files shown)


markt-asf:
I'd be tempted to skip the system property and go with the proposed fix for all current versions.

csutherl:
I'd be fine with that. I didn't want to push a fix that changes some behavior that could potentially result in different server info being available. If there's no objections, I'm happy to drop the config option and change the default behavior to match what I'd expect from reading the doc.

csutherl:
We're discussing a -1 vote for this change on the dev-list.

PR #1072, opened 2026-09-19
Enforce write interest registration state

sainadh777:
## Summary
- Complete TODO 17.1 by rejecting duplicate write-interest registration with a localized IllegalStateException until the pending callback begins.
- Clear the registration when OPEN_WRITE processing starts, or when blocking and vectored write continuations consume poller readiness.
- Add focused state-transition tests, update the Coyote changelog, and remove the completed TODO.

## Validation
- ant validate
- ant test -Dtest.entry=org.apache.tomcat.util.net.TestSocketWrapperBase (2 tests)
- ant test -Dtest.entry=org.apache.catalina.nonblocking.TestNonBlockingAPI -Dtest.entry.methods=testNonBlockingWrite,testNonBlockingWriteWithKeepAlive (2 tests)
- ant clean && ant
- Full unfiltered ant test: 656 suites, 41,398 tests, 0 failures, 0 errors, 328 skips
- Built distribution smoke: HTTP 200 from /, clean shutdown and no remaining listeners

Draft for maintainer review.


sainadh777:
@markt-asf @rmaucher Could you review the write-interest state transition and the poller callback clearing points when convenient? This implements TODO 17.1; the full suite and runtime smoke are green. Thank you.

rmaucher:
This change is very risky and the TODO might not be accurate anymore. But it is good that the testsuite passes (even though this does not prove reliability). The test is a mock so there's no point adding it.

markt-asf:
Agreed on the pointlessness of the test.

Why not deregister in the Poller rather than all the places the Poller calls?

The back-ports for this will be non-trivial since NIO2 has been removed in trunk. After a quick look at the NIO2 code it might already have suitable protection.

